PR #33, opened 2018-07-17
Add support for self-signed certificates w/ HTTPS

pepie:
Self-certificate errors can be bypassed by implementing:

```
Dio dio = new Dio();
dio.options.trustSelfSignedCerts = true;
```

This is accomplished by using HttpClient.badCertificateCallback()
See document here: https://api.dartlang.org/stable/1.24.3/dart-io/HttpClient/badCertificateCallback.html
```
HttpClient client = new HttpClient();
httpClient.badCertificateCallback = (X509Certificate cert, String host, int port)=> this.options.trustSelfSignedCerts??false;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -526,6 +526,23 @@ dio.cookieJar=new PersistCookieJar("./cookies");
 
 More details about [cookie_jar](https://github.com/flutterchina/cookie_jar)  see : https://github.com/flutterchina/cookie_jar .
 
+## HTTPS w/ Self-Signed Certificates
+```
+Options options = new Options(
+trustSelfSignedCerts: true,
+...
+);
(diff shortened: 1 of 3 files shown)


pepie:
Closing this request in favor of a better alternative.
HttpClient (the underlying class used by Dio) can be configured globally as illustrated here:
```class MyHttpOverrides extends HttpOverrides{

  @override
  HttpClient createHttpClient(SecurityContext context){
    return super.createHttpClient(context)
      ..badCertificateCallback = (X509Certificate cert, String host, int port)=> true;
  }
}

void main() async{
  HttpOverrides.global = new MyHttpOverrides();
  runApp(new App());
}
```

PR #96, opened 2018-11-09
fix: urlEncode lose index of List when content-type is 'application/x…

Simbachen:
…-www-form-urlencoded'


wendux:
Could you provide some sample data ?

Simbachen:
> Could you provide some sample data ?

data like this: 

```dart
Map data = {
    'a': 1,
    'b': 2,
    'c': [{
      'd': 1
    },{
      'e': 1
    }]
  }
```

Simbachen:
i fixed the code to get the same result as jQuery.post.

PR #98, opened 2018-11-16
fix use 'baseUrl'. can not use full url request.

zcweng:
like this:

  _dio = Dio(Options(
    baseUrl: 'http://xxxxxx',
  ));

  ///another full url path
  _dio.get('http://*****', data:{});

### DIFF
--- a/lib/src/dio.dart
+++ b/lib/src/dio.dart
@@ -358,7 +358,7 @@ class Dio {
     try {
       // Normalize the url.
       String url = options.path;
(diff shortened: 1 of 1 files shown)


JoniDS:
Why not make this a parameter in **Options** (e.g. {'isFullUrl' : true})?
This way it would accept any url and not be restricted to http/s

wendux:
Sorry, I can't understand this pr purpose, could you clarify it more ?

pedromorgan:
I dont get this PR either ?

PR #141, opened 2019-01-16
Adding an onUploadProgress Callback for Post Requests

jeffosmith:
Will provide feedback for upload progress indicators.  Modelled along the lines of the OnDownloadProgress in file Download.

Can probably look to add a similar approach to downloads.

Currently have 1 test failing, but don't believe this is due to these changes.  Appears an endpoint used for the test has moved (no returning a 302 rather than expected error).  Happy to resolve this test, but didn't want to jump around on it.

### DIFF
--- a/lib/src/dio.dart
+++ b/lib/src/dio.dart
@@ -1,14 +1,16 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
+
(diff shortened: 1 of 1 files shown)


1AlexFix1:
There is news when the branch will be merge?

PR #146, opened 2019-01-21
Update transformer.dart

kaushiksamanta:
Added JSON parsing in background thread using DART isolates.


wendux:
https://github.com/flutterchina/dio#in-flutter

PR #185, opened 2019-02-21
Adjust case for "cookie"

robisaks:
According to RFC 6265, the cookie header should be "Cookie" with a capital C.

https://tools.ietf.org/html/rfc6265#section-4.2


wendux:
Oh, This is case insensitive。
image

PR #266, opened 2019-04-21
Fix example in README using BaseOptions

omuomugin:
Hey, thanks for such a great http client library.
I was so impressed by this work and I'm know fan of this.

As I follow the introduction in README, I found that `BaseOptions` usage example was not correctly assigning to the right type. (By the way IDE told me this was error)

Therefore I updated the README for the newbies.
If I was misunderstanding something just let me know.

Thanks


wendux:
Yes! you're right, thanks for your pr.

PR #311, opened 2019-05-23
Throw exceptions

### LINKED ISSUE #309: Version 2.1.4 Regression - Catching DioError doesn't work
Hi, since version 2.1.4 catching DioError doesn't work anymore. I've written a unit test that fails, but should pass, using the syntax in the documentation:

```
import 'package:flutter_test/flutter_test.dart';

void main() {
  test("dio error test", () async {
    final dio = Dio();

    bool hasError = false;

    try {
      await dio.get("https://wendux.github.io/xsddddd");

      hasError = false;
    } on DioError catch (e) {
      hasError = true;
    }

    expect(hasError, true);
  });
}
```

passsy:
Fixes #309 

Fixes the regression introduced in #304 not throwing exceptions correctly.

I know that it is already fixed by @wendux in `2.1.5` by changing `DioError` to `DioError extends Error implements Exception`. This PR allows the removal of `extends Error` which would be dart idiomatic as explained in #304.

Sadly I can't PR the change on top of `2.1.5` because the changes haven't been pushed.

I added tests that DioErros are thrown and catchable.

### DIFF
--- a/package_src/lib/src/dio.dart
+++ b/package_src/lib/src/dio.dart
@@ -703,8 +703,8 @@ class Dio {
           response = _makeRequest<T>(data, cancelToken);
         } else {
           // Otherwise, use the Future value as the request result.
-          // If the return type is Error, we should throw it
-          if (data is Error) throw _assureDioError(data);
+          // If the return type is Exception, we should throw it
+          if (data is Exception) throw _assureDioError(data);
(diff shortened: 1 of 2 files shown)


wendux:
OK, I will merge this pr by hand, thanks.

PR #396, opened 2019-07-31
I'll delete the file if I need to.

krypticallusion:
### Pull Request Description

while Cancelling a file, the file should not be deleted, rather keep the file and let the user delete the file if he deems necessary.


wendux:
1. Add `deleteOnError` parameter to download API
2. I'm sorry to say that open file with "append" mode will not be supported, because it will cause more problems, such as recover the position of latest download.

PR #404, opened 2019-08-08
Added safe query support

programadorthi:

- [X] I have searched for a similar pull request in the [project](https://github.com/flutterchina/dio/pulls) and found none
- [X] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
- [X] I have added the required tests to prove the fix/feature I am adding

This merge request fixes / refers to the following issues: https://github.com/flutterchina/dio/issues/361

### Pull Request Description
When I have null values in query parameters, dio put in the URL null values. Retrofit remove query null values avoiding put them in the URL.

These changes avoid Dio to put null query values in the URL. When I have `queryParameters: {'something': null}` dio creates the request: `https://endpoint?something=null`. Using safe query we avoid to put null values in the URL and the final request is: `https://endpoint` only.

There is an example in `example/safe_query.dart`.

### DIFF
--- a/example/safe_query.dart
+++ b/example/safe_query.dart
@@ -1,0 +1,8 @@
+import 'package:dio/dio.dart';
+
+/// More examples see https://github.com/flutterchina/dio/tree/master/example
+main() async {
+  var dio = Dio();
+  Response response = await dio.get("https://google.com");
+  print(response.data);
+}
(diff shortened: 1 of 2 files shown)


wendux:
Fixed at v2.1.14.

PR #504, opened 2019-10-12
Update transformer.dart

DavidYang1121:

Error: type 'List<String>' is not a subtype of type 'String'.
### Pull Request Description
Because response.headers's type is Map<String, List<String>>, so  if the header's value is null, then it will use its default value "-1", and "-1" is a String rather than a List, so flutter throw an error.
...


DavidYang1121:
@wendux

amondnet:
```dart
length = int.parse(response.headers[Headers.contentLengthHeader]?.first ?? "-1");
```

wendux:
Fixed

Issue #314, opened 2019-05-28
Uncaught exception, try/catch not working


cosinus84:
Making dio.post throws uncaught exception
```
myAsync() async{
try {
      Response response = await dio
          .post(_appStore.getURL("test/1"), data: formData);
      return true;
    } catch (e) {
      print(e);
    }
}

```
The server requires GET and not POST so I should get an error 405. 
The problem is that the try/catch is not working somehow. When I run the app is pointing as an uncaught exception the exception not being caught.

Flutter (Channel dev, v1.6.3, on Mac OS X 10.14.4 18E226, locale en-RO)
    • Flutter version 1.6.3 at /Users/cosinus/flutter
    • Framework revision bc7bc94083 (5 days ago), 2019-05-23 10:29:07 -0700
    • Engine revision 8dc3a4cde2
    • Dart version 2.3.2 (build 2.3.2-dev.0.0 e3edfd36b2)

![Screenshot 2019-05-28 at 10 59 40](https://user-images.githubusercontent.com/1459132/58461188-ba485e00-8137-11e9-86ff-46d92e300249.png)

Thank you

cosinus84:
I think is something wrong with the debugger, it seems that it treats handled exceptions as uncaught.

cosinus84:
Yes running flutter run everything is working fine.

cosinus84:
There is some issue inside the lib, it may require some fix

https://github.com/flutter/flutter/issues/33427

akuma06:
Same problem here I was wondering why my stream wouldn't work :')

ppyupp:
Same problem...

MichaelPriebe:
Getting this.

jaasaria:
Any update on this?  Got the same problem.

wendux:
Please update to 3.0.x and try again.

fdonzello:
@wendux just added the dependencies and the bug is still there

jobehi:
same here

Chris1234567899:
Still getting this issue  (version 3.0.9)

RKasinCDRK:
Still getting this issue (version 3.0.8)

hugotai101:
```dart
  myAsync() async {
    Response response = await dio
        .post(_appStore.getURL("test/1"), data: formData)
        .catchError((error) {
      print(error);
    });
  }
```
I'm able to catch error this way, using the `.catchError`

bhaskar-nair2:
[...]

PR #593, opened 2019-12-06
Add onError callback to RequestConfig to allow circumventing issue #314

andrew-carroll:

This merge request fixes / refers to the following issues: #314 

### Pull Request Description

Sometimes when debugging in certain IDEs, `try`/`catch` blocks fail to actually catch thrown `DioError`s (see flutter/flutter#33427, in particular [feinstein's comment](https://github.com/flutter/flutter/issues/33427#issuecomment-523605100)). To circumvent this problem, with this PR you can pass an `onError` param to the `Options` of your request, like so:

```dart
Response _onError(DioError error) =>
  // Silently catch errors with responses
  // (e.g. http status errors) but preserve
  // other errors (like SocketExceptions) in
  // the returned Response's data parameter
  error?.response != null
  ? error.response
  : Response(data: error);

// ...

try {
  final response = await dio.get(
    "https://does.not.exist/",
    options: Options(onError: _onError)
  );
  if (response.data is DioError) {
    // Re-throw the preserved DioError in a new context
    // so our try/catch block can see it
    throw response.data;
  }
} on DioError catch (e) {
  // Handle errors
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -516,6 +516,35 @@ try {
 }
 ```
 
+Sometimes when debugging in certain IDEs, such `try`/`catch` blocks fail to actually catch thrown `DioError`s (see [flutter/flutter/issues/33427](https://github.com/flutter/flutter/issues/33427#issuecomment-523605100)). To circumvent this problem, you can pass an `onError` param to the `Options` of your request, like so:
+
+```dart
+Response _onError(DioError error) =>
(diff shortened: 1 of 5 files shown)


2shrestha22:
This problem still persists any workaround for that?

PR #639, opened 2020-01-09
fix(transformer): reduce memory consumption

### LINKED ISSUE #578: Too much memory allocation when using non stream responses
The default response transformer reads all the stream in a `List<int>` thus allocating memory for 8 times greater than needed.
https://github.com/flutterchina/dio/blob/fdf38275092c1f55f123380e4bfb4415893edf71/dio/lib/src/transformer.dart#L91

It should use a growing `Uint8List`,  maybe with the help of a `ByteConversionSink`

Endebert:

Fixes #578

### Pull Request Description

Instead of using a large `List<int>`, I keep references to the data chunks and at the very end create a `Uint8List` and fill it with the chunks. This seems to reduce memory allocation considerably.

### DIFF
--- a/dio/lib/src/transformer.dart
+++ b/dio/lib/src/transformer.dart
@@ -1,12 +1,13 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:typed_data';
+
 import 'package:http_parser/http_parser.dart';
 
 import 'adapter.dart';
-import 'options.dart';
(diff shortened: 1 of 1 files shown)


Endebert:
Could someone please take a look at this? I'd love to tick the remaining checkboxes, but I don't even know how to run the tests for this project.

We use this fix in production and it seems stable. Would be great if this could be merged...

PR #753, opened 2020-04-24
Make Options use a const constructor

lukepighetti:
Making `Options` use a const constructor allows us to put options in other constructors, which is particularly nice when building APIs with a code generation tool such as `retrofit`

```dart
  /// Test authentication token and get private user information
  @POST("/v1/auth/authenticate")
  Future<AuthenticateResponse> authenticate({
    @DioOptions() Options options = const Options(sendTimeout: 1),
  });
```

### Pull Request Description

...


lukepighetti:
This is stale because it hasn't been triaged.

wendux:
It's hard to make Options use a const constructor, because if we do this，the fields of Options must be final, but Options should be modified anytime and anywhere.

PR #760, opened 2020-04-29
Fixes the HTTP2 Adapter for Requests with Payload

hongyiweiwu:

The current HTTP2 adapter doesn't work when the request contains a payload, because it only starts sending the payload after the response header from the server has been received. However, the server won't start sending the response header until the entire payload is received from the client. Hence the connection is in a deadlock until either the client or the server times out.

This pull request fixes this issue by transmitting the entire payload body before starting to listen for responses.

...

### DIFF
--- a/plugins/http2_adapter/lib/src/http2_adapter.dart
+++ b/plugins/http2_adapter/lib/src/http2_adapter.dart
@@ -11,6 +11,7 @@ part 'client_setting.dart';
 
 part 'connection_manager_imp.dart';
 
+
 /// A Dio HttpAdapter which implements Http/2.0.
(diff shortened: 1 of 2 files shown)


posativ:
It will also fix #762 (endStream is no longer set to `false`).

krokyze:
@wendux could you please merge this? Thanks.

ylogx:
I've verified that the patch works for POST calls (which aren't working in the current release). This should be released on pub.dev soon.

byteslaves:
This is available, thank you

savy-91:
Hi! When is this going to be merged?

Issue #762, opened 2020-04-30
Http2Adapter will not receive the full response


posativ:
This is an appendum to #317 with a 100% reproducible test file and also a fix, that worked at least for me.

### Issue Info

| Info                           | Value                                                 |
| ------------------------------ | ----------------------------------------------------- 
| Platform Name                  | Dart VM
| Platform Version               | 2.7.1
| Dio Version                    | 3.0.9, master  
| Repro rate                     | 100%   
| Repro with our demo prj        | yes           
| Demo project link              | see test file below 

### Issue Description and Steps

```
~> dart test.dart
Unhandled exception:
DioError [DioErrorType.DEFAULT]: FormatException: Unterminated string (at character 149204)
..."Nuss Makronen","Eierschwämmchen","Pilaw","Seeberger Soft-Aprikosen","Klein
                                                                              ^

#0      DioMixin._dispatchRequest (package:dio/src/dio.dart:963:7)
<asynchronous suspension>
#1      DioMixin._request._interceptorWrapper.<anonymous closure>.<anonymous closure>.<anonymous closure> (package:dio/src/dio.dart:849:37)
#2      DioMixin.checkIfNeedEnqueue (package:dio/src/dio.dart:1117:22)
#3      DioMixin._request._interceptorWrapper.<anonymous closure>.<anonymous closure> (package:dio/src/dio.dart:846:22)
#4      new Future.<anonymous closure> (dart:async/future.dart:176:37)
#5      Timer._createTimer.<anonymous closure> (dart:async-patch/timer_patch.dart:23:15)
#6      _Timer._runTimers (dart:isolate-patch/timer_impl.dart:384:19)
#7      _Timer._handleMessage (dart:isolate-patch/timer_impl.dart:418:5)
#8      _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:174:12
```

The server is HTTP2 compatible and the backend correctly provides a Content-Length header. The received bytes is usually ~30-60 kilobytes less of what is expected. Total response size is around 200kb.

### Test file

```dart
import 'package:dio/dio.dart';
import 'package:dio_http2_adapter/dio_http2_adapter.dart';

void main() async {
  final dio = Dio()
    ..httpClientAdapter = Http2Adapter(ConnectionManager());

  await dio.get('https://au3.de/reduce-food-waste/ingredients/list');
}
```

### Possible fix

[...]

PR #763, opened 2020-04-30
use endStream: true behavior when making an HTTP/2 request

posativ:
Fix to #762


posativ:
I just noticed that #760 is much better as it fixes the whole HTTP2 in DIO. Right now it is not usable at all.

PR #812, opened 2020-05-31
do not treat 304 as good status code by default

nioncode:
HTTP servers MUST NOT include any content when sending a 304. When treating
304 as success, the response body is empty and transforming the response
leads to an empty string. This then fails in `assureResponse` when
`response.data` is assigned to the variable of type `T` (which is usually
not String, but the expected body type like `List<dynamic>`) with an
exception: type 'String' is not a subtype of type 'List<dynamic>'

By treating 304 as a failure, the request throws a DioError that can
easily be checked for the statusCode and then the application can handle
the 304 however it wants to (usually by reusing the data from the cache).

I did not add a test or updated the documentation, since I wanted to get feedback on this first. I know that I could supply a custom `validateStatus` callback to reject 304 manually, but I don't see why this shouldn't be the default, since Dio always fails to convert the response if the server responds with a 304.

### DIFF
--- a/dio/lib/src/dio.dart
+++ b/dio/lib/src/dio.dart
@@ -1091,7 +1091,7 @@ abstract class DioMixin implements Dio {
       validateStatus: opt.validateStatus ??
           options.validateStatus ??
           (int status) {
-            return status >= 200 && status < 300 || status == 304;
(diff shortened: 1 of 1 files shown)


nioncode:
Ping.

PR #917, opened 2020-09-01
clear connect futures in case of SocketException

krokyze:
Currently in case if `await SecureSocket.connect` throws `SocketException` it's cached in `_connectFutures` and all the following `getConnection` requests fail with the same exception.
<br>
Steps to reproduce:

- Turn off network
- Open the app and make an API call. It will return `SocketException: Failed host lookup: 'myapi.com' (OS Error: No address associated with hostname, errno = 7)`
- Turn on network
- Make another API call. It will return the same `SocketException: Failed host lookup: 'myapi.com' (OS Error: No address associated with hostname, errno = 7)`


anoreg:
I got the same error

krokyze:
Still a issue.

krokyze:
@wendux why was this closed? It's still an active issue.

PR #974, opened 2020-10-23
make headers case insensitive

TarekkMA:

TarekkMA/gql_dio_link#4

### Pull Request Description
The current implementation of the headers do allow multiple header with different letter case.

for example the following header are considered as unique.
Content-type: xxxxxx
content-type: xxxxxx
CoNtEnT-TyPe: xxxxx

The current solution I have used is the same used in http package
https://github.com/dart-lang/http/blob/master/lib/src/base_request.dart#L89
...


TarekkMA:
.

PR #1016, opened 2020-12-17
Fixing slow reduce

ericmartineau:
This part of the code is extremely inefficient.  It took several minutes of pegged CPU to buffer up a few megabytes worth of data.  The `reduce` method on the stream will perform thousands of array copies.  

I added the `buffer` library because it already provides utility classes for efficiently dealing with reading streams.  Even if you used a normal dart list, it would still be slow because the list would have to constantly resize and copy itself.


pawangjain:
Hello.

It appears that file download using dio.download takes some initial time delay to start and is slow at times.

If the above pull request fixes this issue, can it be reviewed and merged?

Thanks.

PR #1017, opened 2020-12-21
Avoid sending zero length data stream message to h2 server

yrom:

### Pull Request Description

See  plugins/http2_adapter/lib/src/http2_adapter.dart#L64-L69

https://github.com/flutterchina/dio/blob/21136168ab39a7536835c7a59ce0465bb05feed4/plugins/http2_adapter/lib/src/http2_adapter.dart#L64-L69

Closing outgoingMessages `stream.outgoingMessages.close();` will send a zero length DataStreamMessage to signal `endOfStream` however the `requestStream` was null.

Some h2 servers would reject zero length DataStreamMessage (response :status 501).

We need mark `endOfStream` to true in this case:
`transport.makeRequest(headers, endStream: true)`


AlexV525:
@yrom The `endStream` flag is no longer set in the current version. Does it regress the feature?

PR #1026, opened 2021-01-09
Add null-safety

### LINKED ISSUE #1002: Support null safety
This is a request to add support for [null safety](https://dart.dev/null-safety) to `package:dio`. This is a very popular package, so would be great to have null safety enabled early. See the [migration guide](https://dart.dev/null-safety/migration-guide) for details about enabling null safety.

### New Issue Checklist

- [x] I have searched for a similar issue in the [project](https://github.com/flutterchina/dio/issues) and found none

harsimranmaan:

- [ ] I have read the [Documentation](https://pub.dartlang.org/packages/dio)

### Pull Request Description

- Add conservative null-safety. Few tests are still failing with a common cause but looking for early feedback to see if this is the right direction to go. 
- Fix common  lint issues

Closes #1002

### DIFF
--- a/dio/lib/src/dio.dart
+++ b/dio/lib/src/dio.dart
@@ -43,18 +43,18 @@ import 'entry_stub.dart'
 ///  ```
 
 abstract class Dio {
-  factory Dio([BaseOptions options]) => createDio(options);
+  factory Dio([BaseOptions? options]) => createDio(options);
 
   /// Default Request config. More see [BaseOptions] .
-  BaseOptions options;
+  BaseOptions? options;
 
   Interceptors get interceptors;
 
-  HttpClientAdapter httpClientAdapter;
+  late HttpClientAdapter httpClientAdapter;
 
   /// [transformer] allows changes to the request/response data before it is sent/received to/from the server
   /// This is only applicable for request methods 'PUT', 'POST', and 'PATCH'.
-  Transformer transformer;
+  late Transformer transformer;
 
   /// Shuts down the dio client.
   ///
@@ -69,115 +69,115 @@ abstract class Dio {
   /// Handy method to make http GET request, which is a alias of  [BaseDio.request].
   Future<Response<T>> get<T>(
     String path, {
-    Map<String, dynamic> queryParameters,
-    Options options,
-    CancelToken cancelToken,
-    ProgressCallback onReceiveProgress,
+    Map<String, dynamic> queryParameters = const {},
+    Options? options,
+    CancelToken? cancelToken,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Handy method to make http GET request, which is a alias of [BaseDio.request].
   Future<Response<T>> getUri<T>(
     Uri uri, {
-    Options options,
-    CancelToken cancelToken,
-    ProgressCallback onReceiveProgress,
+    Options? options,
+    CancelToken? cancelToken,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Handy method to make http POST request, which is a alias of  [BaseDio.request].
   Future<Response<T>> post<T>(
     String path, {
     data,
-    Map<String, dynamic> queryParameters,
-    Options options,
-    CancelToken cancelToken,
-    ProgressCallback onSendProgress,
-    ProgressCallback onReceiveProgress,
+    Map<String, dynamic> queryParameters = const {},
+    Options? options,
+    CancelToken? cancelToken,
+    ProgressCallback? onSendProgress,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Handy method to make http POST request, which is a alias of  [BaseDio.request].
   Future<Response<T>> postUri<T>(
     Uri uri, {
     data,
-    Options options,
-    CancelToken cancelToken,
-    ProgressCallback onSendProgress,
-    ProgressCallback onReceiveProgress,
+    Options? options,
+    CancelToken? cancelToken,
+    ProgressCallback? onSendProgress,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Handy method to make http PUT request, which is a alias of  [BaseDio.request].
   Future<Response<T>> put<T>(
     String path, {
     data,
-    Map<String, dynamic> queryParameters,
-    Options options,
-    CancelToken cancelToken,
-    ProgressCallback onSendProgress,
-    ProgressCallback onReceiveProgress,
+    Map<String, dynamic> queryParameters = const {},
+    Options? options,
+    CancelToken? cancelToken,
+    ProgressCallback? onSendProgress,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Handy method to make http PUT request, which is a alias of  [BaseDio.request].
   Future<Response<T>> putUri<T>(
     Uri uri, {
     data,
-    Options options,
-    CancelToken cancelToken,
-    ProgressCallback onSendProgress,
-    ProgressCallback onReceiveProgress,
+    Options? options,
+    CancelToken? cancelToken,
+    ProgressCallback? onSendProgress,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Handy method to make http HEAD request, which is a alias of [BaseDio.request].
   Future<Response<T>> head<T>(
     String path, {
     data,
-    Map<String, dynamic> queryParameters,
-    Options options,
-    CancelToken cancelToken,
+    Map<String, dynamic> queryParameters = const {},
+    Options? options,
+    CancelToken? cancelToken,
   });
 
   /// Handy method to make http HEAD request, which is a alias of [BaseDio.request].
   Future<Response<T>> headUri<T>(
     Uri uri, {
     data,
-    Options options,
-    CancelToken cancelToken,
+    Options? options,
+    CancelToken? cancelToken,
   });
 
   /// Handy method to make http DELETE request, which is a alias of  [BaseDio.request].
   Future<Response<T>> delete<T>(
     String path, {
     data,
-    Map<String, dynamic> queryParameters,
-    Options options,
-    CancelToken cancelToken,
+    Map<String, dynamic> queryParameters = const {},
+    Options? options,
+    CancelToken? cancelToken,
   });
 
   /// Handy method to make http DELETE request, which is a alias of  [BaseDio.request].
   Future<Response<T>> deleteUri<T>(
     Uri uri, {
     data,
-    Options options,
-    CancelToken cancelToken,
+    Options? options,
+    CancelToken? cancelToken,
   });
 
   /// Handy method to make http PATCH request, which is a alias of  [BaseDio.request].
   Future<Response<T>> patch<T>(
     String path, {
     data,
-    Map<String, dynamic> queryParameters,
-    Options options,
-    CancelToken cancelToken,
-    ProgressCallback onSendProgress,
-    ProgressCallback onReceiveProgress,
+    Map<String, dynamic> queryParameters = const {},
+    Options? options,
+    CancelToken? cancelToken,
+    ProgressCallback? onSendProgress,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Handy method to make http PATCH request, which is a alias of  [BaseDio.request].
   Future<Response<T>> patchUri<T>(
     Uri uri, {
     data,
-    Options options,
-    CancelToken cancelToken,
-    ProgressCallback onSendProgress,
-    ProgressCallback onReceiveProgress,
+    Options? options,
+    CancelToken? cancelToken,
+    ProgressCallback? onSendProgress,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Assure the final future state is succeed!
@@ -243,13 +243,13 @@ abstract class Dio {
   Future<Response> download(
     String urlPath,
     savePath, {
-    ProgressCallback onReceiveProgress,
-    Map<String, dynamic> queryParameters,
-    CancelToken cancelToken,
+    ProgressCallback? onReceiveProgress,
+    Map<String, dynamic> queryParameters = const {},
+    CancelToken? cancelToken,
     bool deleteOnError = true,
     String lengthHeader = Headers.contentLengthHeader,
     data,
-    Options options,
+    Options? options,
   });
 
   ///  Download the file and save it in local. The default http method is "GET",
@@ -290,12 +290,12 @@ abstract class Dio {
   Future<Response> downloadUri(
     Uri uri,
     savePath, {
-    ProgressCallback onReceiveProgress,
-    CancelToken cancelToken,
+    ProgressCallback? onReceiveProgress,
+    CancelToken? cancelToken,
     bool deleteOnError = true,
     String lengthHeader = Headers.contentLengthHeader,
     data,
-    Options options,
+    Options? options,
   });
 
   /// Make http request with options.
@@ -307,11 +307,11 @@ abstract class Dio {
   Future<Response<T>> request<T>(
     String path, {
     data,
-    Map<String, dynamic> queryParameters,
-    CancelToken cancelToken,
-    Options options,
-    ProgressCallback onSendProgress,
-    ProgressCallback onReceiveProgress,
+    Map<String, dynamic> queryParameters = const {},
+    CancelToken? cancelToken,
+    Options? options,
+    ProgressCallback? onSendProgress,
+    ProgressCallback? onReceiveProgress,
   });
 
   /// Make http request with options.
@@ -322,17 +322,17 @@ abstract class Dio {
   Future<Response<T>> requestUri<T>(
     Uri uri, {
     data,
-    CancelToken cancelToken,
-    Options options,
-    ProgressCallback onSendProgress,
-    ProgressCallback onReceiveProgress,
+    CancelToken? cancelToken,
+    Options? options,
+    ProgressCallback? onSendProgress,
+    ProgressCallback? onReceiveProgress,
   });
 }
 
 abstract class DioMixin implements Dio {
   /// Default Request config. More see [BaseOptions].
   @override
-  BaseOptions options;
+  BaseOptions? options;
 
   /// Each Dio instance has a interceptor by which you can intercept requests or responses before they are
   /// handled by `then` or `catchError`. the [interceptor] field
@@ -343,7 +343,7 @@ abstract class DioMixin implements Dio {
   Interceptors get interceptors => _interceptors;
 
   @override
-  HttpClientAdapter httpClientAdapter;
+  late HttpClientAdapter httpClientAdapter;
 
   @override
   Transformer transformer = DefaultTransformer();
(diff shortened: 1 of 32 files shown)


harsimranmaan:
The tests fail with
```
00:05 +10 -5: test/request_test.dart: #test requests #test generic parameters [E]                             
  DioError [DioErrorType.DEFAULT]: type '() => Null' is not a subtype of type '(() => FutureOr<Socket>)?' of 'onTimeout'
  #0      Future.timeout (dart:async/future_impl.dart)
  #1      _ConnectionTarget.connect.<anonymous closure> (dart:_http/http_impl.dart:2222:37)
  #2      StackZoneSpecification._registerUnaryCallback.<anonymous closure>.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:126:26)
  #3      StackZoneSpecification._run (package:stack_trace/src/stack_zone_specification.dart:208:15)
  #4      StackZoneSpecification._registerUnaryCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:126:14)
  #5      _rootRunUnary (dart:async/zone.dart:1194:47)
  #6      _CustomZone.runUnary (dart:async/zone.dart:1097:19)
  #7      _FutureListener.handleValue (dart:async/future_impl.dart:150:18)
  #8      Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:703:45)
  #9      Future._propagateToListeners (dart:async/future_impl.dart:732:32)
  #10     Future._completeWithValue (dart:async/future_impl.dart:536:5)
  #11     Future._asyncCompleteWithValue.<anonymous closure> (dart:async/future_impl.dart:574:7)
  #12     StackZoneSpecification._run (package:stack_trace/src/stack_zone_specification.dart:208:15)
  #13     StackZoneSpecification._registerCallback.<anonymous closure> (package:stack_trace/src/stack_zone_specification.dart:116:48)
  #14     _rootRun (dart:async/zone.dart:1186:13)
  #15     _CustomZone.run (dart:async/zone.dart:1090:19)
  #16     _CustomZone.runGuarded (dart:async/zone.dart:994:7)
  #17     _CustomZone.bindCallbackGuarded.<anonymous closure> (dart:async/zone.dart:1034:23)
  #18     _microtaskLoop (dart:async/schedule_microtask.dart:41:21)
  #19     _startMicrotaskLoop (dart:async/schedule_microtask.dart:50:5)
  #20     _runPendingImmediateCallback (dart:isolate-patch/isolate_patch.dart:120:13)
  #21     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:185:5)
  
  package:dio/src/dio.dart 875:17   DioMixin._request._errorInterceptorWrapper.<fn>.<fn>.<fn>
  ===== asynchronous gap ===========================
  dart:async                        Future.then
  package:dio/src/dio.dart 873:18   DioMixin._request._errorInterceptorWrapper.<fn>.<fn>
  package:dio/src/dio.dart 1127:22  DioMixin.checkIfNeedEnqueue
  package:dio/src/dio.dart 870:16   DioMixin._request._errorInterceptorWrapper.<fn>
  dart:async                        _completeOnAsyncError
  package:dio/src/dio.dart          DioMixin._dispatchRequest
  ===== asynchronous gap ===========================
  dart:async                        Future.catchError
  package:dio/src/dio.dart 907:23   DioMixin._request.<fn>
  dart:collection                   ListMixin.forEach
  package:dio/src/dio.dart 906:18   DioMixin._request
  package:dio/src/dio.dart 764:12   DioMixin.request
  package:dio/src/dio.dart 368:12   DioMixin.get
  test/request_test.dart 136:28     main.<fn>.<fn>
  test/request_test.dart 132:38     main.<fn>.<fn>
  
  OS Error: Connection reset by peer, errno = 54
  dart:isolate  _RawReceivePortImpl._handleMessage
```
Haven't had a chance to dig deep for the underlying cause yet.

dio/lib/src/transformer.dart
```diff
@@ -138,27 +140,26 @@ class DefaultTransformer extends Transformer {
...
       responseBody = utf8.decode(responseBytes, allowMalformed: true);
     }
-    if (responseBody != null &&
-        responseBody.isNotEmpty &&
+    if (responseBody.isNotEmpty &&
         options.responseType == ResponseType.json &&
         _isJsonMime(response.headers[Headers.contentTypeHeader]?.first)) {
       if (jsonDecodeCallback != null) {
-        return jsonDecodeCallback(responseBody);
+        return jsonDecodeCallback!(responseBody);
```
kuhnroyal, line 154: Is this conversion needed?

venkatd: Don't think it should be necessary since we're doing a null check already, no?

harsimranmaan: Yes. I think VS tooling hasn't caught up yet for proper null safety checks. At the time of writing, I couldn't get the IDE to not complain here

venkatd: @harsimranmaan I think this a limitation of the language and nothing to do with VS tooling. You will have to declare a local variable like `final decodeCallback = jsonDecodeCallback` and use that.

See here for more information: https://stackoverflow.com/questions/65456958/dart-null-safety-doesnt-work-with-class-fields

miDeb:
@harsimranmaan the test failure seems to be caused by the dart sdk itself, I've opened https://github.com/dart-lang/sdk/issues/44895.

dio/lib/src/transformer.dart
mit-mit: Yes, we currently cannot promote on class members. Using a local variable is a good workaround.

miDeb:
@harsimranmaan the fix in `dart:io` has landed in the flutter master channel, and the tests pass again (well, they don't all pass on my machine but I believe that's because I'm on linux, so please check). I'd also have a few suggestions, but I can also open a follow up PR once this is merged. @wendux could you take a look?

harsimranmaan:
Tests pass now against the dart version with the fix for the timeout issue. There were line ending mismatch in `#test FormData`. Would be worth running the test suite on windows too.

dio/lib/src/dio.dart
```diff
@@ -935,12 +940,12 @@ abstract class DioMixin implements Dio {
         statusMessage: responseBody.statusMessage,
         extra: responseBody.extra,
       );
-      var statusOk = options.validateStatus(responseBody.statusCode);
-      if (statusOk || options.receiveDataWhenStatusError) {
+      var statusOk = options.validateStatus!(responseBody.statusCode);
+      if (statusOk || options.receiveDataWhenStatusError != null) {
```
miDeb, line 944: The check for the actual value of `options.receiveDataWhenStatusError` is missing
```suggestion
      if (statusOk || options.receiveDataWhenStatusError == true) {
```

miDeb:
Now the test is failing in windows :laughing: (checked in a windows vm). I believe `.replaceAll('\r\n', '\n')` should not be called on Windows.

harsimranmaan:
Fixed the tests. Should work on both windows and other OS now.

dio/lib/src/options.dart
```diff
@@ -80,27 +80,30 @@ class BaseOptions extends _RequestConfig {
           maxRedirects: maxRedirects,
           requestEncoder: requestEncoder,
           responseDecoder: responseDecoder,
-        );
+        ) {
+    this.queryParameters = queryParameters ?? {};
+    this.baseUrl = baseUrl ?? '';
```
miDeb, line 85: If you use an initializer list here
```dart
: queryParameters = queryParameters ?? {},
 super();
```

then queryParameters doesn't need to be `late`. The reason why we can't use a default parameter here is that default parameters need to be `const`, which would mean that you can't modify the map afterwards anymore.
(See [the section about constructors in the language tour](https://dart.dev/guides/language/language-tour#constructors) for more information about initializer lists.

dio/lib/src/options.dart
```diff
@@ -50,22 +50,22 @@ typedef RequestEncoder = List<int> Function(
...
-    String contentType,
-    ValidateStatus validateStatus,
-    bool receiveDataWhenStatusError = true,
-    bool followRedirects = true,
-    int maxRedirects = 5,
-    RequestEncoder requestEncoder,
-    ResponseDecoder responseDecoder,
+    int? receiveTimeout,
+    int? sendTimeout,
+    String? baseUrl,
```
miDeb, line 57: It is easier here to just use a default parameter instead of doing the check inside of the constructor.
```suggestion
    this.baseUrl = '',
```
You can then also remove `late` from `baseUrl`.

dio/lib/src/options.dart
```diff
@@ -109,48 +112,48 @@ class BaseOptions extends _RequestConfig {
       connectTimeout: connectTimeout ?? this.connectTimeout,
       receiveTimeout: receiveTimeout ?? this.receiveTimeout,
       sendTimeout: sendTimeout ?? this.sendTimeout,
-      extra: extra ?? Map.from(this.extra ?? {}),
-      headers: headers ?? Map.from(this.headers ?? {}),
-      responseType: responseType ?? this.responseType,
+      extra: extra ?? this.extra,
+      headers: headers ?? this.headers,
+      responseType: (responseType != null) ? responseType : this.responseType,
```
miDeb, line 117: You can revert the change on this line, using `??` is just syntactic sugar for what you wrote.

dio/lib/src/options.dart
```diff
@@ -109,48 +112,48 @@ class BaseOptions extends _RequestConfig {
       connectTimeout: connectTimeout ?? this.connectTimeout,
       receiveTimeout: receiveTimeout ?? this.receiveTimeout,
       sendTimeout: sendTimeout ?? this.sendTimeout,
-      extra: extra ?? Map.from(this.extra ?? {}),
-      headers: headers ?? Map.from(this.headers ?? {}),
-      responseType: responseType ?? this.responseType,
+      extra: extra ?? this.extra,
+      headers: headers ?? this.headers,
```
miDeb, line 116: Using `Map.from` probably had the intention to clone the map. I guess you should add it back

dio/lib/src/options.dart
```diff
@@ -169,62 +172,62 @@ class Options extends _RequestConfig {
...
+    RequestEncoder? requestEncoder,
+    ResponseDecoder? responseDecoder,
   }) {
     return Options(
       method: method ?? this.method,
       sendTimeout: sendTimeout ?? this.sendTimeout,
       receiveTimeout: receiveTimeout ?? this.receiveTimeout,
-      extra: extra ?? Map.from(this.extra ?? {}),
-      headers: headers ?? Map.from(this.headers ?? {}),
+      extra: extra ?? this.extra,
```
miDeb, line 193: Same as above - `Map.from` is probably required.

dio/lib/src/options.dart
```diff
@@ -169,62 +172,62 @@ class Options extends _RequestConfig {
...
-    int receiveTimeout,
+    String? method,
+    int? sendTimeout,
+    int? receiveTimeout,
     this.connectTimeout,
     this.data,
-    this.path,
-    this.queryParameters,
-    this.baseUrl,
+    String? path,
```
miDeb, line 215: Same as with the `BaseOptions` constructor. 
```suggestion
    this.path = '',
```

dio/lib/src/options.dart
```diff
@@ -169,62 +172,62 @@ class Options extends _RequestConfig {
...
+    Map<String, dynamic>? queryParameters,
+    String? baseUrl,
```
miDeb, line 217:
```suggestion
    this.baseUrl = '',
```

dio/lib/src/options.dart
```diff
@@ -239,32 +242,36 @@ class RequestOptions extends Options {
           maxRedirects: maxRedirects,
           requestEncoder: requestEncoder,
           responseDecoder: responseDecoder,
-        );
+        ) {
+    this.path = path ?? this.path;
+    this.queryParameters = queryParameters ?? {};
```
miDeb, line 247: Again, queryParameters can be in the initializer list and doesn't need to be late.

dio/lib/src/options.dart
```diff
@@ -278,16 +285,16 @@ class RequestOptions extends Options {
       onReceiveProgress: onReceiveProgress ?? this.onReceiveProgress,
       onSendProgress: onSendProgress ?? this.onSendProgress,
       cancelToken: cancelToken ?? this.cancelToken,
-      extra: extra ?? Map.from(this.extra ?? {}),
-      headers: headers ?? Map.from(this.headers ?? {}),
+      extra: extra ?? this.extra,
```
miDeb, line 288: `Map.from` is probably required.

dio/lib/src/options.dart
```diff
@@ -336,52 +343,55 @@ class _RequestConfig {
...
-    String contentType,
+    Map<String, dynamic>? extra,
+    Map<String, dynamic>? headers,
+    String? contentType,
     this.responseType,
     this.validateStatus,
-    this.receiveDataWhenStatusError = true,
-    this.followRedirects = true,
-    this.maxRedirects = 5,
+    bool? receiveDataWhenStatusError,
```
miDeb, line 351: Here we can again use default parameters to simplify the code.
```suggestion
    this.receiveDataWhenStatusError = true,
```

dio/lib/src/options.dart
```diff
@@ -336,52 +343,55 @@ class _RequestConfig {
...
+    bool? followRedirects,
```
miDeb, line 352:
```suggestion
    this.followRedirects = true,
```

dio/lib/src/options.dart
```diff
@@ -336,52 +343,55 @@ class _RequestConfig {
...
+    int? maxRedirects,
```
miDeb, line 353:
```suggestion
    this.maxRedirects = 5,
```

dio/lib/src/options.dart
```diff
@@ -336,52 +343,55 @@ class _RequestConfig {
     this.method,
     this.receiveTimeout,
     this.sendTimeout,
-    Map<String, dynamic> extra,
-    Map<String, dynamic> headers,
-    String contentType,
+    Map<String, dynamic>? extra,
+    Map<String, dynamic>? headers,
+    String? contentType,
```
miDeb, line 348:
```suggestion
    this.contentType,
```

dio/lib/src/options.dart
```diff
@@ -336,52 +343,55 @@ class _RequestConfig {
...
-    this.followRedirects = true,
-    this.maxRedirects = 5,
+    bool? receiveDataWhenStatusError,
+    bool? followRedirects,
+    int? maxRedirects,
     this.requestEncoder,
     this.responseDecoder,
   }) {
     this.headers = headers ?? {};
     this.extra = extra ?? {};
```
miDeb, line 358: If you move `headers` and `extra` to the initializer list they don't have to be `late`.

dio/lib/src/options.dart
```diff
@@ -396,34 +406,34 @@ class _RequestConfig {
...
 
   /// `validateStatus` defines whether the request is successful for a given
   /// HTTP response status code. If `validateStatus` returns `true` ,
   /// the request will be perceived as successful; otherwise, considered as failed.
-  ValidateStatus validateStatus;
+  ValidateStatus? validateStatus;
 
   /// Whether receiving response data when http status code is not successful.
-  bool receiveDataWhenStatusError;
+  bool? receiveDataWhenStatusError;
```
miDeb, line 417: `receiveDataWhenStatusError`, `followRedirects` and `maxRedirects` don't need to be nullable.

dio/lib/src/dio.dart
```diff
@@ -43,18 +43,18 @@ import 'entry_stub.dart'
 ///  ```
 
 abstract class Dio {
-  factory Dio([BaseOptions options]) => createDio(options);
+  factory Dio([BaseOptions? options]) => createDio(options);
 
   /// Default Request config. More see [BaseOptions] .
-  BaseOptions options;
+  BaseOptions? options;
```
miDeb, line 49: `options` can be non-nullable. I believe there are different ways to achieve that...
If you look around for its usages it is always assigned `BaseOptions()` as a default value, and every access uses `options!.foo`. I believe there is no reason to make it nullable.

dio/lib/src/dio.dart
```diff
@@ -43,18 +43,18 @@ import 'entry_stub.dart'
...
 
   Interceptors get interceptors;
 
-  HttpClientAdapter httpClientAdapter;
+  late HttpClientAdapter httpClientAdapter;
 
   /// [transformer] allows changes to the request/response data before it is sent/received to/from the server
   /// This is only applicable for request methods 'PUT', 'POST', and 'PATCH'.
-  Transformer transformer;
+  late Transformer transformer;
```
miDeb, line 57: Bonus points if you can make `options`, `httpClientAdapter` and `transformer` not be `late`; It's just because they are always initialized right away in constructors, which in most cases can be somehow rewritten to not require `late`. I don't think that change is absolutely necessary, though.

miDeb (changes requested):
Hope these suggestions make sense :)

dio/lib/src/options.dart
harsimranmaan: This is a bit tricky as it affect merge-ability. Would not be possible to detect if the source had the bool/int value set or default value of true is being used.

dio/lib/src/options.dart
harsimranmaan: contentType is not a class field bur set/get

harsimranmaan:
Thanks for the suggestions @miDeb. I have addressed the feedback. I have started to like dart now :)

dio/lib/src/options.dart
miDeb: I think you are referring to the `mergeOptions` or `merge` methods, right? The thing is, it was impossible to create a `_RequestConfig` with a null value for these fields. [Its constructor used to assign default values if you passed null](https://github.com/flutterchina/dio/pull/1026/commits/d3a208de4bf217663eaa3c83c33cf8d64daaaf56#diff-bed0ea39c9af0b71c65fb305e8c68236e0292cad32dd6693fe8092aefcdfb6a7L360). The way it is now, it would be possible to pass a null value, but my point is it was not possible before. It was only possible to re-set the default value and assign null after the object was created. I just felt that was unlikely to happen, but I obviously don't know how all the usage of `dio` looks like.
I do however think that the original author was confused about where a null value represents a default and where there are non-null default values. I noticed this in code I have written as well, null safety really helps here!

dio/lib/src/options.dart
miDeb: You're obviously right!

miDeb (approved):
I'm obviously not a mantainer here, but these changes now look good to me! Thanks for all your effort @harsimranmaan!

acoutts:
LGTM!

acoutts:
Don't forget to bump the version to next major release as per the null safety migration recommendations.

kuhnroyal:
Yea that needs to be done by the absent maintainer :)

miDeb:
One more thing to keep in mind is whether the fix to https://github.com/dart-lang/sdk/issues/44895 has landed in all null-safe release channels.

TitanKing:
Thanks for now I will use your package @harsimranmaan, appreciate your efforts. 

This might need to be created as a new package.

aarontwf:
Does the cookie_manager plugin also need updating?

miDeb:
> Does the cookie_manager plugin also need updating?
That requires this pr and [cookie_jar](https://github.com/flutterchina/cookie_jar/pull/24) first, but otherwise yes.

masseelch:
For everyone landing here: If you want to use this PR until it gets merged add this to your `pubspec.yaml`:

```yaml
dependency_overrides:
  dio: # https://github.com/flutterchina/dio/issues/1002
    git:
      url: git@github.com:harsimranmaan/dio.git
      ref: null-safety
      path: dio
```

Remove it once it gets merged.

j4qfrost:
@wendux Thanks for the merge on cookie_jar. Do you mind taking a look this too?

PR #1055, opened 2021-02-28
Prevents Bad State: No Element from being thrown

Molkars:
If the reduce method is called on an empty stream (like Stream.empty()), a Bad State: No Element exception is thrown. This change adds a check to determine if the stream is empty before sending the request. If the stream is empty, the request is sent but if the stream is not empty, the stream is reduced before being sent.

I ran into this issue when using doing a get request on flutter web.

### Pull Request Description

When using dio on web, like in dio's flutter example, a get request will fail and return a Bad State: No Element exception. This is caused by the reduce method being called on an empty stream. This pull requests adds a check to determine if the stream is empty before reducing it to prevent a Bad State: No element error.


wendux:
This feature is published  in 4.0.0-beta3 (preview version)

PR #1262, opened 2021-08-31
Fixed: update in example/lib/queued_interceptor_csrftoken.dart

seunghwanly:

This merge request fixes / refers to the following issues: example/interceptor_lock.dart

### Pull Request Description

In my case, the example code had keep occurring the infinite requests to the server
So, I changed the callback functions to `await` and it worked :)

Here is the code that I run
#### `options`
``` dart
dio.options.connectTimeout = 5000;
dio.options.maxRedirects = 3;
dio.options.responseType = ResponseType.bytes;
```
#### `onRequest`, if the token is null 
``` dart
           tokenDio
              .post('/obtain-token/', data: jsonEncode(requestBody))
              .then((res) {
            /// response will be different by server
            final parsedResult = jsonDecode(utf8.decode(res.data as Uint8List));
            accessToken = parsedResult['result']['access'];
            refreshToken = parsedResult['result']['refresh'];
            options.headers['Authorization'] = 'Bearer $accessToken';

            /// save to [SharedPrefernces], it can be null
            prefs.setString('accessToken', accessToken!);
            prefs.setString('refreshToken', refreshToken!);

            /// go on to next operation
            handler.next(options);
          }).catchError((error, stackTrace) {
            log(error.toString());
            handler.reject(error, true);
          }).whenComplete(
                  () => dio.unlock()); // unlock after all the task is done
```

#### `onError`, if the status code is 401
``` dart
     if ('Bearer $accessToken' != options.headers['Authorization']) {
            log('Bearer ${accessToken!}', name: 'string');
            log(options.headers['Authorization'], name: 'option');
            options.headers['Authorization'] = 'Bearer $accessToken';

            final response = await dio.fetch(options);
            return handler.resolve(response);
          }
```
### Summary

What I recommend is that avoid using too many callback functions in a single task 
I think nested callback functions can occur some conflicts and are hard to debug as well 👍

### DIFF
--- a/dio/lib/src/response.dart
+++ b/dio/lib/src/response.dart
@@ -32,16 +32,6 @@ class Response<T> {
   /// Http status code.
   int? statusCode;
 
-  /// Returns if the response has an error
-  ///
-  /// `hasError` is true when status code is not between 200 and 299
-  bool get hasError => statusCode != null && statusCode! ~/ 100 != 2;
-
-  /// Returns if the response has succeed
-  ///
-  /// `hasSucceed` is true when status code is between 200 and 299
-  bool get hasSucceed => !hasError;
-
   /// Returns the reason phrase associated with the status code.
   /// The reason phrase must be set before the body is written
   /// to. Setting the reason phrase after writing to the body.
--- a/dio/lib/src/response.dart
+++ b/dio/lib/src/response.dart
@@ -32,6 +32,16 @@ class Response<T> {
   /// Http status code.
   int? statusCode;
 
+  /// Returns if the response has an error
+  ///
+  /// `hasError` is true when status code is not between 200 and 299
+  bool get hasError => statusCode != null && statusCode! ~/ 100 != 2;
+
+  /// Returns if the response has succeed
+  ///
+  /// `hasSucceed` is true when status code is between 200 and 299
+  bool get hasSucceed => !hasError;
+
   /// Returns the reason phrase associated with the status code.
   /// The reason phrase must be set before the body is written
   /// to. Setting the reason phrase after writing to the body.
--- a/dio/README.md
+++ b/dio/README.md
@@ -444,28 +444,41 @@ Because of security reasons, we need all the requests to set up a csrfToken in t
   //  dio instance to request token
   var tokenDio = Dio();
   String? csrfToken;
-  dio.options.baseUrl = 'http://www.dtworkroom.com/doris/1/2.0.0/';
+  dio.options.baseUrl = 'https://seunghwanlytest.mocklab.io/';
   tokenDio.options = dio.options;
   dio.interceptors.add(QueuedInterceptorsWrapper(
-    onRequest: (options, handler) {
+    onRequest: (options, handler) async {
       print('send request：path:${options.path}，baseURL:${options.baseUrl}');
+
       if (csrfToken == null) {
         print('no token，request token firstly...');
-        tokenDio.get('/token').then((d) {
-          options.headers['csrfToken'] = csrfToken = d.data['data']['token'];
-          print('request token succeed, value: ' + d.data['data']['token']);
-          print(
-              'continue to perform request：path:${options.path}，baseURL:${options.path}');
-          handler.next(options);
-        }).catchError((error, stackTrace) {
-          handler.reject(error, true);
-        });
-      } else {
-        options.headers['csrfToken'] = csrfToken;
-        return handler.next(options);
+
+        final result = await tokenDio.get('/token');
+
+        if (result.hasSucceed) {
+          /// assume `token` is in response body
+          final body = jsonDecode(result.data) as Map<String, dynamic>?;
+
+          if (body != null && body.containsKey('data')) {
+            options.headers['csrfToken'] = csrfToken = body['data']['token'];
+            print('request token succeed, value: $csrfToken');
+            print(
+              'continue to perform request：path:${options.path}，baseURL:${options.path}',
+            );
+            return handler.next(options);
+          }
+        }
+
+        return handler.reject(
+          DioError(requestOptions: result.requestOptions),
+          true,
+        );
       }
+
+      options.headers['csrfToken'] = csrfToken;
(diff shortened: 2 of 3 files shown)


AlexV525:
Hi @seunghwanly, now `dio` has `QueuedInterceptor` that handles fetch by sequence. Can you verify if the change is still needed and make a rebase on the latest code? Thanks!

seunghwanly:
@AlexV525 
Thanks for the reply! It's been a while since I left this pull request. As you say so, I'll check if `dio` still needs some changes, or else may I just close this pull request with a comment?

AlexV525:
> or else may I just close this pull request with a comment?

We're planning to merge valid PRs to another community-maintained fork, so it would be great if you can follow up the upstream changes.

AlexV525:
Oh if you make force pushes on the `master` branch, the PR will be closed automatically.

seunghwanly:
Yeah, I thought the example I pushed is not suitable from now on... Let me add some examples for `QueuedInterceptorsWrapper` used in `example/lib/queued_interceptor_crsftoken.dart`

seunghwanly:
## New Pull Request Checklist
[x] I have read the [Documentation](https://pub.dartlang.org/packages/dio)
[x] I have searched for a similar pull request in the [project](https://github.com/flutterchina/dio/pulls) and found none
[x] I have updated this branch with the latest master to avoid conflicts (via merge from master or rebase)
[x] I have added the required tests to prove the fix/feature I am adding
[x] I have updated the documentation (if necessary)
[x] I have run the tests and they pass
This merge request fixes / refers to the following issues: example/lib/queued_interceptor_csrftoken.dart

## Pull Request Description

### Summary
- added getter methods in the `Response` class for better usage of `statusCode`
- removed callbacks 
- replaced callbacks(by using `.then( )` and `.catchError( )`s) with async/await 
- reflected the changes in README.md

### Issues
1. in the example(example/lib/queued_interceptor_csrftoken.dart) the `_onResult` method was not being called, since the example baseUrl ('http://www.dtworkroom.com/doris/1/2.0.0/') was not able to run
2. at line 34 in the example, the token is already expired because the status code was 401. And also in `onError` closure, after the token has been updated then the `dio` asks for `fetch` and `resolve` it.
So I thought the below code might be an extra task because the `dio` will update the token and resolve the request.
``` dart
// If the token has been updated, repeat directly.
        if (csrfToken != options.headers['csrfToken']) {
          options.headers['csrfToken'] = csrfToken;
          //repeat
          dio.fetch(options).then(
            (r) => handler.resolve(r),
            onError: (e) {
              handler.reject(e);
            },
          );
          return;
        }
```

### Opinion
- for better practice, using async/await can be more helpful than using in a synchronous way

dio/lib/src/response.dart
```diff
@@ -32,6 +32,16 @@ class Response<T> {
   /// Http status code.
   int? statusCode;
 
+  /// Returns if the response has an error
+  ///
+  /// `hasError` is true when status code is not between 200 and 299
+  bool get hasError => statusCode != null && statusCode! ~/ 100 != 2;
```
AlexV525, line 38: IIRC we have `ValidateStatus` as a prediction, it might be inappropriate to use such a getter.

AlexV525 (commented):
Thanks for the update!

AlexV525:
#1457 is trying to solve the issue by refactoring the interceptor in the example. Can you verify if both fixes are valid or which should we apply?

dio/lib/src/response.dart
seunghwanly: I thought it would be nice to have the getter direct to the `Response`object, but if it is an appropriate way I'll remove the getter. Thanks :)

seunghwanly:
> #1457 is trying to solve the issue by refactoring the interceptor in the example. Can you verify if both fixes are valid or which should we apply?

I've seen the #1457  pull request and I think we have both valid fixes. Since we are assuming the same way as below
``` json
{
  'data': {
    'token': 'valid token'
  }
}
```
and in my opinion, it needs some changes like
- too many callbacks occurred in a nested way
- too much implementation for updating the token, such as creating a private class, which makes it less reusable. Handling the token updates can be easily done by implementing `onRequest`, `onError`, and `onResponse`.
- handling `Future`s  in a nested callback will make it the users hard to debug and trace the issues.

AlexV525:
> I've seen the #1457 pull request and I think we have both valid fixes.

Great. Given your reply I would like to pick your PR first, then we can continue to discuss another one if it gets more feedback.

seunghwanly:
Glad to hear that :) I removed the getter methods as well.

AlexV525:
We need to fix these before merge.

```
Analyzing example...

  error - lib/queued_interceptor_crsftoken.dart:22:20 - The getter 'hasSucceed' isn't defined for the type 'Response<dynamic>'. Try importing the library that defines 'hasSucceed', correcting the name to the name of an existing getter, or defining a getter or field named 'hasSucceed'. - undefined_getter
  error - lib/queued_interceptor_crsftoken.dart:67:2[8](https://github.com/cfug/dio3/actions/runs/3291142303/jobs/5424914251#step:5:9) - The getter 'hasSucceed' isn't defined for the type 'Response<dynamic>'. Try importing the library that defines 'hasSucceed', correcting the name to the name of an existing getter, or defining a getter or field named 'hasSucceed'. - undefined_getter
```

seunghwanly:
Sorry, I forgot to remove the getter method in the example. Thanks for the notice!

AlexV525:
Hi @seunghwanly. We've made our hardfork repo public and published a new version of `dio`, named `diox`.

seunghwanly:
Thanks for letting me know :) Good to hear that more future jobs will be taken at diox library!
It seems that this PR is already in [diox's example](https://github.com/cfug/diox/blob/main/example/lib/queued_interceptor_crsftoken.dart). Is more action needed for the next step?

AlexV525:
> Is more action needed for the next step?

No. :) Thanks again for your contribution.

PR #1263, opened 2021-09-01
Enable multipart files to support additional headers

Virepri:

### Pull Request Description

Enables adding additional headers (outside Content-Type) to files in multipart requests [(akin to Golang's multipart support)](https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/mime/multipart/writer.go;l=97)

### DIFF
--- a/dio/lib/src/form_data.dart
+++ b/dio/lib/src/form_data.dart
@@ -22,7 +22,7 @@ class FormData {
   final _newlineRegExp = RegExp(r'\r\n|\r|\n');
 
   /// The form fields to send for this request.
-  final fields = <MapEntry<String, String>>[];
+  final fields = <MapEntry<String, String>>[]; // todo (which will probably require a breaking change): support additional headers for fields
 
   /// The [files].
   final files = <MapEntry<String, MultipartFile>>[];
@@ -88,6 +88,14 @@ class FormData {
     }
     header = '$header\r\n'
         'content-type: ${file.contentType}';
+    if (file.headers != null) { // append additional headers
+      file.headers?.forEach((key, values) {
+        values.forEach((value) {
+          header = '$header\r\n'
+              '$key: $value';
+        });
+      });
+    }
     return '$header\r\n\r\n';
   }
 
--- a/dio/lib/src/multipart_file.dart
+++ b/dio/lib/src/multipart_file.dart
@@ -21,6 +21,9 @@ class MultipartFile {
   /// The basename of the file. May be null.
   final String? filename;
 
+  /// The additional headers the file has. May be null.
+  final Map<String, List<String>>? headers;
+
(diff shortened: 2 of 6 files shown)


dio/lib/src/form_data.dart
```diff
@@ -22,7 +22,7 @@ class FormData {
   final _newlineRegExp = RegExp(r'\r\n|\r|\n');
 
   /// The form fields to send for this request.
-  final fields = <MapEntry<String, String>>[];
+  final fields = <MapEntry<String, String>>[]; // todo (which will probably require a breaking change): support additional headers for fields
```
jgoyvaerts, line 25: Remove this todo, I don't think it belongs in this PR

dio/lib/src/form_data.dart
```diff
@@ -88,6 +88,14 @@ class FormData {
     }
     header = '$header\r\n'
         'content-type: ${file.contentType}';
+    if (file.headers != null) { // append additional headers
+      file.headers?.forEach((key, values) {
```
jgoyvaerts, line 92: use ! instead of ?. You just checked it's not null so you can sefely use !

jgoyvaerts (changes requested):
Unfortunately this repo doesn't seem to be maintained anymore but I'd love to merge your PR into our fork, (dio_http)[https://github.com/dart-tools/dio_http] .

Just added 2 small change requests

seunghwanly:
> Unfortunately this repo doesn't seem to be maintained anymore but I'd love to merge your PR into our fork, (dio_http)[https://github.com/dart-tools/dio_http] .
> 
> Just added 2 small change requests
> 
> Edit: Also can you add the test commit to your main commit so you only have one commit? thanks

Hi, I faced the issue with the Interceptor while requesting APIs. 

I hope to see if your package `dio_http` does catch the Interceptor issue with `onRequest` and `onError` callback functions.

 I just solved it by using the `http` package to catch the issue that Interceptor does not catch 401 error with the `onError` function. And I think the way I did is not the right answer... it occurs too many callback functions..😭

jgoyvaerts:
@seunghwanly can you create an issue on https://github.com/dart-tools/dio_http with a code sample of what is not working with the interceptor?

seunghwanly:
> @seunghwanly can you create an issue on https://github.com/dart-tools/dio_http with a code sample of what is not working with the interceptor?

I have adjusted `dio_http` and found it worked well.. thank you very much!! 😭
Would you mind what was the issue with the interceptor in `dio` package?? What was the problem..?

Virepri:
>Edit: Also can you add the test commit to your main commit so you only have one commit? thanks

If you choose to merge, a squash n merge will do exactly that. That said, I have squashed under the [dart-tools-squash](https://github.com/Virepri/dio/tree/dart-tools-squash) branch on my PR, if you want to use that.

PR #1292, opened 2021-10-09
fix

maxzod:

if you declare the request data type  `final res = await Dio().get<T>(....)`;
it returns nullable object T? but i have specifically told it it is not nullable when i need to be nullable i will use T? instead

this PR fixes this issues

### DIFF
--- a/dio/lib/src/dio_error.dart
+++ b/dio/lib/src/dio_error.dart
@@ -42,7 +42,7 @@ class DioError implements Exception {
 
   /// The original error/exception object; It's usually not null when `type`
   /// is DioErrorType.DEFAULT
-  dynamic? error;
+  dynamic error;
 
   StackTrace? _stackTrace;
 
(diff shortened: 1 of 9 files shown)


NicolaVerbeeck:
Does this PR also fix the case where T _is_ nullable, this has been broken since 4.0.1 (wrong non-null cast in dio_mixin's assureResponse)

maxzod:
@NicolaVerbeeck  yes

kuhnroyal:
Not sure this is the right approach. In any case, this is a breaking change as it is now required to specify `T?` for everything that previously worked.

kuhnroyal:
While I like the idea of explicitly specifying if the response is nullable, this will be a breaking change.
We can consider this for a 5.0.0 release if there is a general consensus. I started a discussion for this: https://github.com/flutterchina/dio/discussions/1335

@NicolaVerbeeck The `T?` problem has been fixed in https://github.com/flutterchina/dio/pull/1322

PR #1317, opened 2021-11-13
Use catchError to mark thrown error as handled

markfili:

Debugger will halt execution while debugging an app that uses Dio to generate requests. This is really an issue with Dart that is known and is somewhat being worked on and this is a quickfix for everyone using Dio with debugger.

### Pull Request Description
Use catchError method to let the debugger know that the thrown DioError will be caught so it doesn't hold execution while debugging.
Fix for flutterchina#1306

### DIFF
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -588,7 +588,7 @@ abstract class DioMixin implements Dio {
           handler.reject(e as DioError, true);
         },
       ).catchError((e) {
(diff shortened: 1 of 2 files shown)


wendux:
Please try with 4.0.2.  changes as follows: 

```dart
    future = future.then(_requestInterceptorWrapper((
      RequestOptions reqOpt,
    ) {
      requestOptions = reqOpt;
      _dispatchRequest(reqOpt)
          .then((value) => handler.resolve(value, true))
          .catchError((e) {
        handler.reject(e as DioError, true);
      });
    }));
```

PR #1322, opened 2021-11-22
Null data

via-guy:

Nullable data responses were being cast to `T` instead of `T?`, which throws a runtime error when the data response is `null`.

### Pull Request Description

Instead of receiving a `null` response, fetching the data throws a runtime error because the code was attempting to cast a null value to a non-nullable type.


via-guy:
@wendux this is really important, can you please look at it?

kuhnroyal:
@jgoyvaerts This seems to make sense

PR #1333, opened 2021-11-28
Fix CI after merge

kuhnroyal:
* don't collect test result for example app, there are no test atm
* remove usage of `unawaited` - pedantic is deprecated and the new `unawait` from the Dart SDK required Dart 2.14 which would raise the min SDK for the library
* try to prevent test not completing with Dart 2.12 because of a a wrong expectation on a `Future` - use `expectLater` instead
* re-add example test which got lost in the merge

### Pull Request Description

...


kuhnroyal:
The basic_test ist still hanging in CI with Dart 2.12 - not sure why but it seems some Future in the native socket is already completed. I think this is related to the expectation on the future.

kuhnroyal:
The `upload_stream_test` and `basic_test` both resolve the `http://httpbin.org/status/` URL.

kuhnroyal:
The test problem was hot-fixed in `2.12.1` - we should require this as minimum version.
https://github.com/dart-lang/sdk/issues/45047
https://github.com/dart-lang/sdk/issues/45214

kuhnroyal:
Seems to have worked.

@wendux Can you setup a branch protection rule for develop requiring reviews, resolutions and status checks so we ensure it remains stable from now on?

Bildschirmfoto 2021-11-28 um 19 14 50

wendux:
I have add you as maintainer, you can do(merge、config...) it by yourself 😆

kuhnroyal:
@wendux  Thanks! Merging this as everything is green now.
But collaborators can not configure the branch protection rules :)

PR #1367, opened 2022-01-05
Make `ResponseBody.statusCode` non-nullable

ueman:

### Pull Request Description

This makes `ResponseBody.statusCode` non-nullable. It improves the API a little bit.
It's safe to change this as [`http`](https://github.com/dart-lang/http/blob/1e42ffa181b263f7790f276a5465832bff7ce615/lib/src/browser_client.dart#L57) also does this and has even wider adoption rates.


kuhnroyal (approved):
I agree, validated this in the `http` package.

kuhnroyal:
This change is actually breaking, `sentry_dio` doesn't compile with it, can you revert that in a new PR and pull it into develop-5.0?

ueman:
Dang it. Yeah, sure 👍

PR #1381, opened 2022-01-12
Export barrel dependencies for conditional imports

javaherisaber:
# Problem Statement
I have a `Flutter Web` application that needs to have *Cookie* in headers.
To enable sending `Cookie` in *request headers* we need to add the following code to our Dio object:
```dart
// conditional import to support BrowserHttpClientAdapter for Web.
import 'package:dio/dio.dart'
    if (dart.library.io) 'package:dio/native_imp.dart'
    if (dart.library.html) 'package:dio/browser_imp.dart';

_dio.options = BaseOptions(
      receiveTimeout: receiveTimeout,
      connectTimeout: connectTimeout,
      sendTimeout: sendTimeout,
      extra: kIsWeb ? {'withCredentials': true} : null, // allow sending Cookie in header
 );
```

When `running` or `building` for Web we have these compile errors:
```bat
lib/data/repository.dart:236:8: Error: Type 'Dio' not found.
    late Dio _dio;^^^^
lib/data/repository.dart:246:20: Error: The method 'BaseOptions' isn't defined for the class '_RepositoryImpl'.
 - '_RepositoryImpl' is from 'package:zireh/data/repository.dart' ('lib/data/repository.dart').
Try correcting the name to the name of an existing method, or defining a method named 'BaseOptions'.
    _dio.options = BaseOptions(^^^^

Failed to compile application.
```

# What this pull request solves
This pull request is exporting dependencies for both `native` and `web` implementations of *Dio*.

### DIFF
--- a/dio/lib/barrels.dart
+++ b/dio/lib/barrels.dart
@@ -1,0 +1,14 @@
+export 'src/adapter.dart';
+export 'src/cancel_token.dart';
+export 'src/dio.dart';
+export 'src/dio_error.dart';
+export 'src/dio_mixin.dart' hide InterceptorState, InterceptorResultType;
+export 'src/form_data.dart';
+export 'src/headers.dart';
+export 'src/interceptors/log.dart';
+export 'src/multipart_file.dart';
+export 'src/options.dart';
+export 'src/parameter.dart';
+export 'src/redirect_record.dart';
+export 'src/response.dart';
+export 'src/transformer.dart';
--- a/dio/lib/browser_imp.dart
+++ b/dio/lib/browser_imp.dart
@@ -1,1 +1,3 @@
 export 'src/entry/dio_for_browser.dart' show DioForBrowser;
+
+export 'barrels.dart';
--- a/dio/lib/dio.dart
+++ b/dio/lib/dio.dart
@@ -5,17 +5,4 @@
 /// A open source project authorized by [https://flutterchina.club](https://flutterchina.club).
 library dio;
 
-export 'src/adapter.dart';
-export 'src/cancel_token.dart';
-export 'src/dio.dart';
-export 'src/dio_error.dart';
-export 'src/dio_mixin.dart' hide InterceptorState, InterceptorResultType;
-export 'src/form_data.dart';
-export 'src/headers.dart';
-export 'src/interceptors/log.dart';
-export 'src/multipart_file.dart';
-export 'src/options.dart';
-export 'src/parameter.dart';
-export 'src/redirect_record.dart';
-export 'src/response.dart';
-export 'src/transformer.dart';
+export 'barrels.dart';
--- a/dio/lib/native_imp.dart
+++ b/dio/lib/native_imp.dart
@@ -1,1 +1,3 @@
 export 'src/entry/dio_for_native.dart' show DioForNative;
+
+export 'barrels.dart';


ueman:
Kinda related https://github.com/flutterchina/dio/issues/1362

kuhnroyal:
For your example it should be sufficient to just import `dio/dio.dart`.
The platform handling is done automatically, unless you want to configure something on the `DefaultHttpClientAdapter` or `BrowserClientAdapter` but then you need to create platform specific files anyways.

I know that we can improve the adapter exports but I don't see how this PR fixes anything.

javaherisaber:
@kuhnroyal 
I am using `withCredentials` feature which is only available at `BrowserClientAdapter` through `browser_imp.dart`
This pull request exports dependencies for both **native** and **web** implementations of *Dio*
Without this PR it is not possible to achieve this, except importing from `/src/**`.

kuhnroyal:
Yea but `withCredentials` is just a value in the extras map. You don't need to set the `BrowserClientAdapter`, it is configured automatically.

ueman:
Exporting `BrowserClientAdapter` for native probably also breaks native because you can't use web on `dart:io` platforms

kuhnroyal:
So to be clear, this does the same thing:

```dart

_dio.options = BaseOptions(
      receiveTimeout: receiveTimeout,
      connectTimeout: connectTimeout,
      sendTimeout: sendTimeout,
      extra: kIsWeb ? {'withCredentials': true} : null, // allow sending Cookie in header
 );
 ```

**Comment by javaherisaber** [comment]
@kuhnroyal @ueman 
The below code is not sufficient to use `withCredentials`, because we don't have access to `BaseOptions` and even `Dio` in our conditional import.

> import 'package:dio/dio.dart';

> _dio.options = BaseOptions(
>     receiveTimeout: receiveTimeout,
>      connectTimeout: connectTimeout,
>      sendTimeout: sendTimeout,
>      extra: kIsWeb ? {'withCredentials': true} : null, // allow sending Cookie in header
> );

Here we say to import implementations for `dio`:

> import 'package:dio/dio.dart'
>   if (dart.library.io) 'package:dio/native_imp.dart'
>    if (dart.library.html) 'package:dio/browser_imp.dart';

So, dependencies are not visible to the consumer package. because `native_imp.dart` and `browser_imp.dart` are not exporting anything!

**Comment by javaherisaber** [comment]
@kuhnroyal 
> Yea but withCredentials is just a value in the extras map. You don't need to set the BrowserClientAdapter, it is configured automatically.

That's not the case for `Dio`, We only have access to this option here:
https://github.com/flutterchina/dio/blob/0e12e0198f5cd0bdd7992a1f0c529f203ba3385c/dio/lib/src/adapters/browser_adapter.dart#L34

**Comment by ueman** [comment]
Dio uses the `BrowserClientAdapter` on web because it does 

```dart
import 'package:dio/dio.dart'
    if (dart.library.io) 'package:dio/native_imp.dart'
    if (dart.library.html) 'package:dio/browser_imp.dart';
```
under the hood itself to set the correct adapter based on the underlying platform.

kuhnroyal:
> Dio uses the `BrowserClientAdapter` on web because it does
> 
> ```dart
> import 'package:dio/dio.dart'
>     if (dart.library.io) 'package:dio/native_imp.dart'
>     if (dart.library.html) 'package:dio/browser_imp.dart';
> ```
> 
> under the hood itself to set the correct adapter based on the underlying platform.

Exactly, the `BrowserClientAdapter` is used by default on web. You don't need to manually configure it.

javaherisaber:
@kuhnroyal @ueman 
This code this the trick:
```dart
import 'package:dio/native_imp.dart' if (dart.library.html) 'package:dio/browser_imp.dart';
```
Thanks for your time.

PR #1405, opened 2022-01-26
Improve `DioError`s

ueman:
NOTE: This goes against the `develop-5.0` branch, since it's a breaking change.

### Pull Request Description

This significantly improves the inner stacktraces of the `DioError`.


jgoyvaerts:
Good idea! Can you include an example of a stacktrace message before and after your PR so people can see the difference?

ueman:
> Good idea! Can you include an example of a stacktrace message before and after your PR so people can see the difference?

I'm still thinking about how to produce those cases, in order to test it. Do you have any ideas?

kuhnroyal:
Just a reminder too myself: Consider raising to Dart 2.16 and using `Error.rethrowWithStackTrace`.

ueman:
@kuhnroyal Do you mind sharing your test cases? Then I can go through them.

kuhnroyal:
It's nothing special, I only tested:

1. 404 response from a random URL (this has no inner error as the response is received)
2. Something like `post<void>(url, data: ArgumentError())` - this throws a serialization error an has an inner error/stacktrace

dio/lib/src/adapters/browser_adapter.dart
```diff
@@ -151,11 +149,12 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
     xhr.onError.first.then((_) {
       // Unfortunately, the underlying XMLHttpRequest API doesn't expose any
       // specific information about the error itself.
+      // https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget/onerror
```
AlexV525, line 152:
```suggestion
      // See also: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget/onerror
```

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -56,13 +56,18 @@ class DefaultHttpClientAdapter implements HttpClientAdapter {
...
+        rethrow;
       }
-      rethrow;
-    } on TimeoutException {
-      _throwConnectingTimeout();
+      throw DioError.connectionTimeout(
+        requestOptions: options,
+        timeout: options.connectTimeout ??
+            _httpClient.connectionTimeout ??
+            Duration(),
```
AlexV525, line 67: Maybe `Duration.zero`?

dio/lib/src/dio_error.dart
```diff
@@ -2,35 +2,109 @@ import 'options.dart';
 import 'response.dart';
 
 enum DioErrorType {
-  /// It occurs when url is opened timeout.
-  connectTimeout,
+  /// Caused by a connection timeout.
+  connectionTimeout,
```
AlexV525, line 6: The changes with `DioErrorType` should be considered as a breaking change, which means you'll need to update the migration guide?

AlexV525 (commented):
Hi @ueman. The change looks valid overall, with only a few nit-pickings.

Also, there are a few things I'm considering:
- Are you willing to contribute by letting us pick the PR to another community-maintained `dio`? (We'll announce the new repo later once we merge PRs as much as possible.)
- It might be better to add some tests.
- `.DS_Store` can be removed from the PR.

ueman:
> Hi @ueman. The change looks valid overall, with only a few nit-pickings.
> 
> Also, there are a few things I'm considering:
> 
> * Are you willing to contribute by letting us pick the PR to another community-maintained `dio`? (We'll announce the new repo later once we merge PRs as much as possible.)

Sure.

> * It might be better to add some tests.

I agree, any specific suggestions for tests? Just so I don't miss anything, you might think is important.

> * `.DS_Store` can be removed from the PR.

Yes, good point

dio/lib/src/adapters/io_adapter.dart
ueman: Yep, makes sense

ueman:
```suggestion
            Duration.zero,
```

dio/lib/src/dio_error.dart
ueman: Yes, this is a breaking change

AlexV525:
> Sure.

Thanks!

> I agree, any specific suggestions for tests? Just so I don't miss anything, you might think is important.

IMO tests should be covered by checking if requests throw corresponding exceptions in condition.

dio/lib/src/dio_error.dart
ueman: I don't think I need to add anything to the migration guide. From a user's point of view, nothing changes. The user can still catch exceptions the same way as before. All breaking changes are done from within the package in this PR.

dio/lib/src/dio_mixin.dart
```diff
@@ -630,6 +636,11 @@ abstract class DioMixin implements Dio {
         }
       }
 
+      StackTrace? stackTrace;
+      if (maybeStackTrace is StackTrace?) {
```
AlexV525, line 640: `maybeStackTrace` will always be `StackTrace`. The condition can be `maybeStackTrace != StackTrace.empty`.

dio/CHANGELOG.md
```diff
@@ -1,17 +1,23 @@
-# Unreleased
-- require Dart `2.12.1` which fixes exception handling for secure socket connections (https://github.com/dart-lang/sdk/issues/45214)  
-- `ResponseBody.statusCode` is now non-nullable
-- Only delete file if it exists when downloading.
-
 # Unreleased 5.0
-- add option to instantiate a `HttpClientAdapter`, which is platform independent
 
-# 4.0.5-beta1
+- Add option to instantiate a `HttpClientAdapter`, which is platform independent
```
AlexV525, line 3: Seems like we need to change the base branch to `develop`.

dio/lib/src/adapters/browser_adapter.dart
```diff
@@ -74,17 +74,17 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
...
-            completer.completeError(
-              DioError(
-                requestOptions: options,
-                error: 'Connecting timed out [${options.connectTimeout}ms]',
-                type: DioErrorType.connectTimeout,
-              ),
-              StackTrace.current,
-            );
-            xhr.abort();
+          if (haveSent) {
```
AlexV525, line 77: This will be fix by another PR.

dio/lib/src/dio_error.dart
```diff
@@ -2,35 +2,109 @@ import 'options.dart';
...
 
-  /// When the server response, but with a incorrect status, such as 404, 503...
-  response,
+  /// The [DioError] was caused by an incorrect status code as configured by
+  /// [ValidateStatus].
+  badResponse,
 
   /// When the request is cancelled, dio will throw a error with this type.
-  cancel,
+  requestCancelled,
```
AlexV525, line 19: Users usually ignore canceled requests since they are meant to do it, which means they'll rely on the en. Maybe we can keep the naming at least for the least breaking.

AlexV525 (commented):
Given that we've already merged a couple PRs in the fork, I would suggest that we can submit this PR to the fork directly.

AlexV525 (commented):
I would now suggest submitting the PR to the fork directly. Also, the target branch should be `main`.

ueman:
> I would now suggest submitting the PR to the fork directly. Also, the target branch should be `main`.

I don't have the permissions to do that

AlexV525:
> I don't have the permissions to do that

Sorry. It should be granted now.

AlexV525:

kuhnroyal:
All the server response errors have no stackTrace at all. It is impossible to tell where the error originated.

PR #1414, opened 2022-02-07
Expect different keys for list of maps in multiCompatible FormData

nikokon:

### Pull Request Description

I was having an issue with posting lists of maps using formdata to my rails backend using the Listformat.multiCompatible.
My guess is that during key generation we do not wish to add [i] if the list entry value is a Map. See test-code added for my expectation.

I have currently not altered the expected outcome of nested lists as I am uncertain how they should be represented.

The frame of reference from my expectations come from rails documentation and working with forms in rails
https://guides.rubyonrails.org/v6.1/form_helpers.html#combining-them

If anyone has any more insight or pointers as to where I can find some specs I'm happy to learn more :)

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -2,6 +2,29 @@
 - require Dart `2.12.1` which fixes exception handling for secure socket connections (https://github.com/dart-lang/sdk/issues/45214)  
 - Only delete file if it exists when downloading.
 - Fix `BrowserHttpClientAdapter` canceled hangs
+- Fix encoding map using ListFormat.multiCompatible when formatting list of maps.
+  ```
+    FormData.fromMap({"nested": [
+        {"id": "idA", "name": "nameA"},
+        {"id": "idB"},
+      ]}, ListFormat.multiCompatible)
+  ```
+  previously:
+  ```
+    #=>
+    nested[0][id]=idA
+    nested[0][name]=nameA
+    nested[1][id]=idB
+  ```
+  current:
+  ```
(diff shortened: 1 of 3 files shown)


kuhnroyal:
Thanks!
Some formatting issue, test looks good.
Can you please add a line in the changelog?

kuhnroyal:
This seems to make sense but I can't find anything on it with a quick search.
We need to make sure this does not break existing on map serialization for existing users.

nikokon:
@kuhnroyal Thanks for the quick response. I agree with it being potentially breaking change. I would love to know if users are aware of this issue and what their thoughts are. I mean one could ADD a "new" listformat wich mimics my behavior but would love to know the original idea with multicompatible. The description reads very much like my implementation.

nikokon:
So what do you think about this PR?
Should I put this parsing under some other ListFormat flag?

Is there any documentation anywhere regarding the other ListFormats.
All I could find historically regarding the naming conventions was this old stack overflow issue. (which contains links to the rails approach)
https://stackoverflow.com/questions/2142591/names-of-html-form-naming-conventions

nikokon:
I've parted from the idea of going this route. couple of things to keep in mind.
if using this implementation it is of high importance that ALL objects have the same key structure as otherwise the object boundries are unclear. 
I shot myself in the foot here luckily, noticing a bug when I was passing a list of objects that would somethimes contain an id and sometimes contain a multipart file in data.

When reaching the backend, the data and id keys would sometimes belong to the wrong object :D

PR #1455, opened 2022-04-01
Make ResponseDecoder return nullable String type since the usage is r…

huhx:
This PR make the ResponseDecoder return nullable String
 
In options.dart:
```dart
typedef ResponseDecoder = String Function(
    List<int> responseBytes, RequestOptions options, ResponseBody responseBody);
```

The usage of this function: src/transformer.dart
```dart
String? responseBody;
if (options.responseDecoder != null) {
  responseBody = options.responseDecoder!(
    responseBytes,
    options,
    response..stream = Stream.empty(),
  );
} else {
  responseBody = utf8.decode(responseBytes, allowMalformed: true);
}
```
As we can see the responseBody is nullable, it means `ResponseDecoder` function return String? may be make sense.

### DIFF
--- a/dio/lib/src/transformer.dart
+++ b/dio/lib/src/transformer.dart
@@ -153,7 +153,7 @@ class DefaultTransformer extends Transformer {
     } else {
       responseBody = utf8.decode(responseBytes, allowMalformed: true);
     }
-    if (responseBody.isNotEmpty &&
+    if (responseBody !=null && responseBody.isNotEmpty &&
         options.responseType == ResponseType.json &&
         _isJsonMime(response.headers[Headers.contentTypeHeader]?.first)) {
       final callback = jsonDecodeCallback;
--- a/dio/lib/src/transformer.dart
+++ b/dio/lib/src/transformer.dart
@@ -150,10 +150,11 @@ class DefaultTransformer extends Transformer {
         options,
         response..stream = Stream.empty(),
       );
-    } else {
+    } else if (responseBytes.isNotEmpty) {
       responseBody = utf8.decode(responseBytes, allowMalformed: true);
     }
-    if (responseBody !=null && responseBody.isNotEmpty &&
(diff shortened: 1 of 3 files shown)


dio/lib/src/transformer.dart
```diff
@@ -153,7 +153,7 @@ class DefaultTransformer extends Transformer {
     } else {
       responseBody = utf8.decode(responseBytes, allowMalformed: true);
     }
-    if (responseBody.isNotEmpty &&
+    if (responseBody !=null && responseBody.isNotEmpty &&
```
AlexV525, line 156:
```suggestion
    if (responseBody != null &&
        responseBody.isNotEmpty &&
```

AlexV525 (commented):
Hi @huhx, since you've mentioned the transformer, can you add a check to return an encoded body when the response body bytes are not empty? Such as:

```dart
String? responseBody;
if (options.responseDecoder != null) {
  responseBody = options.responseDecoder!(
    responseBytes,
    options,
    response..stream = Stream.empty(),
  );
} else if (responseBytes.isNotEmpty) { // <-- Add this condition.
  responseBody = utf8.decode(responseBytes, allowMalformed: true);
}
```

Also, are you willing to contribute by letting us pick the PR to another community-maintained dio? (We'll announce the new repo later once we merge PRs as much as possible.) Also, it might be better to add some tests.

AlexV525:
@huhx Can you make a new PR based on the `develop` branch? Otherwise, it'll be impossible to merge your PR alone.

huhx:
> @huhx Can you make a new PR based on the `develop` branch? Otherwise, it'll be impossible to merge your PR alone.

👌

huhx:
> @huhx Can you make a new PR based on the `develop` branch? Otherwise, it'll be impossible to merge your PR alone.

This is the new PR: https://github.com/flutterchina/dio/pull/1581

PR #1490, opened 2022-05-30
 Fix the problem that the function of forcing the plain text response to json is invalid.

ipcjs:

### DIFF
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -602,7 +602,7 @@ abstract class DioMixin implements Dio {
       RequestInterceptorHandler handler,
     ) {
       requestOptions = reqOpt;
-      _dispatchRequest(reqOpt)
+      _dispatchRequest<T>(reqOpt)
           .then((value) => handler.resolve(value, true))
           .catchError((e) {
         handler.reject(e as DioError, true);
@@ -649,7 +649,7 @@ abstract class DioMixin implements Dio {
   }
 
   // Initiate Http requests
-  Future<Response<T>> _dispatchRequest<T>(RequestOptions reqOpt) async {
+  Future<Response<dynamic>> _dispatchRequest<T>(RequestOptions reqOpt) async {
     var cancelToken = reqOpt.cancelToken;
     ResponseBody responseBody;
     try {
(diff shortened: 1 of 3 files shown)


AlexV525:
Hi @ipcjs, can you explain more about the PR? I can't do a prediction based on your test. Also, I saw there is `forceConvert`, is it related?

dio/lib/src/dio_mixin.dart
```diff
@@ -602,7 +602,7 @@ abstract class DioMixin implements Dio {
       RequestInterceptorHandler handler,
     ) {
       requestOptions = reqOpt;
-      _dispatchRequest(reqOpt)
```
ipcjs, line 605: The original code didn't pass the generic T down.

AlexV525: Sure, but why would `_dispatchRequest` return a dynamic response?

dio/test/options_test.dart
```diff
@@ -270,4 +273,21 @@ void main() {
...
+
+  test('#test forceConvert responseType', () async {
+    final dio = Dio(BaseOptions(
+      baseUrl: MockAdapter.mockBase,
+    )) //
+      ..httpClientAdapter = MockAdapter();
+    final expectedResponseData = <String, dynamic>{"code": 0, "result": "ok"};
+
+    final response = await dio.get<Map<String, dynamic>>('/test-force-convert');
+    expect(response.data, expectedResponseData);
```
ipcjs, line 285: For this call, the generic T is `Map<String, dynamic>` and the responshould be forced to parse as json.

dio/lib/src/dio_mixin.dart
ipcjs: https://github.com/ipcjs/dio/blob/42f644a70e0d000067fa9a378fef906e8c4e1afa/dio/lib/src/dio_mixin.dart#L684

Because `transformer.transformResponse(reqOpt, responseBody)` return a dynamic

ipcjs: https://github.com/ipcjs/dio/blob/42f644a70e0d000067fa9a378fef906e8c4e1afa/dio/lib/src/dio_mixin.dart#L606

`handler.resolve(value, true)` also receives a `Response<dynamic>` parameter.

ipcjs: The generic type T here is more like a parameter that represents a type and does nothing else.🤔

AlexV525 (approved):
Thanks. Merging to the fork,

PR #1549, opened 2022-08-20
fix(pinning): add callback for leaf certificate for better pinning

timshadel:

- [X] I have searched for a similar pull request in the [project](https://github.com/flutterchina/dio/pulls) and found none
- [X] I have updated this branch with the latest `develop` to avoid conflicts (via merge from master or rebase)
- [X] I have added the required tests to prove the fix/feature I am adding
- [X] I have updated the documentation (if necessary)
- [X] I have run the tests and they pass

### Pull Request Description

There are debates about the usefulness of certificate or public key pinning. It is required at many major companies. If someone chooses to do it then they need access to a leaf certificate.

**The existing recommendations on certificate verification don't provide a leaf certificate for the request.**

All of the following are insufficient.

1. Add a certificate to the `SecureContext`
2. Use `badCertificateCallback`
3. Use the `http_cert_pinning` package

First. When you add a cert to the `SecureContext`, it will only check it against the root certificate. This means that you're going to declare trust in DigiCert or AWS or your self-signed cert. This works for a self-signed cert, but for all others this means that any middle man with a cert from your same provider can hijack your traffic.

Second. As my new test shows, `badCertificateCallback` does not return a leaf certificate. Instead it returns one higher up the chain. That often means that you're pinning to an AWS or other authority, and not the leaf certificate made specifically for your server.

Third. The `http_cert_pinning` package doesn't actually do anything with the certificates that Dio is handling. Instead, it maintains a cache of domains which it has separately tested against a list of certificate fingerprints it was given. When a request comes in, the plugin looks at the requested host, and simply asks the package if it ever checked that domain to see if _the package_ got the correct certificates. It doesn't actually look at the certificates from _this_ connection in Dio.

**The problem stems from Dart itself.**

As you can see, there is an open issue for Dart complaining that [the `badCertificateCallback` doesn't return a leaf certificate](https://github.com/dart-lang/sdk/issues/39425). But one of the commenters found a workaround, and [left some instructions on how to get the correct certificate](https://github.com/dart-lang/sdk/issues/39425#issuecomment-680830653).

That comment was _very_ easy to apply to the `DefaultHttpRequestAdapter`. It simply adds one more callback at the right place in the response processing.  If it isn't used, the request continues as expected. As my tests show, it properly delivers the leaf certificate to the callback where users can decide on the methods they want to use to approve or reject a certificate. The tests also show that `badCertificateCallback` does not return the leaf certificate.

Please consider adding this so that we may get one step closer to implementing full certificate pinning in pure Dart.

### DIFF
--- a/dio/lib/src/adapters/io_adapter.dart
+++ b/dio/lib/src/adapters/io_adapter.dart
@@ -7,6 +7,8 @@ import '../dio_error.dart';
 import '../redirect_record.dart';
 
 typedef OnHttpClientCreate = HttpClient? Function(HttpClient client);
+typedef ResponseCertApprover = bool Function(
+    X509Certificate? certificate, String host, int port);
 
 HttpClientAdapter createAdapter() => DefaultHttpClientAdapter();
 
@@ -16,6 +18,9 @@ class DefaultHttpClientAdapter implements HttpClientAdapter {
   /// If [onHttpClientCreate] is provided, [Dio] will call
   /// it when a HttpClient created.
   OnHttpClientCreate? onHttpClientCreate;
+
+  /// Allows the user to decide if the response certificate is good.
+  ResponseCertApprover? responseCertApprover;
 
   HttpClient? _defaultHttpClient;
 
@@ -102,6 +107,20 @@ class DefaultHttpClientAdapter implements HttpClientAdapter {
         error: 'Receiving data timeout[${options.receiveTimeout}ms]',
         type: DioErrorType.receiveTimeout,
       );
+    }
+
+    if (responseCertApprover != null) {
+      final host = options.uri.host;
+      final port = options.uri.port;
+      final isCertApproved =
+          responseCertApprover!(responseStream.certificate, host, port);
+      if (!isCertApproved) {
+        throw DioError(
+          requestOptions: options,
+          error: 'Response certificate NOT approved.',
+          type: DioErrorType.response,
+        );
+      }
     }
 
     var stream =
--- a/dio/lib/src/adapters/io_adapter.dart
+++ b/dio/lib/src/adapters/io_adapter.dart
@@ -20,6 +20,11 @@ class DefaultHttpClientAdapter implements HttpClientAdapter {
   OnHttpClientCreate? onHttpClientCreate;
 
   /// Allows the user to decide if the response certificate is good.
+  /// If this function is missing, then the certificate is allowed.
+  /// This method is called only if both the [SecurityContext] and
+  /// [badCertificateCallback] accept the certificate chain. Those
+  /// methods evaluate the root or intermediate certificate, while
+  /// [responseCertApprover] evaluates the leaf certificate.
   ResponseCertApprover? responseCertApprover;
 
   HttpClient? _defaultHttpClient;
--- a/dio/test/exception_test.dart
+++ b/dio/test/exception_test.dart
@@ -1,4 +1,5 @@
 import 'dart:io';
+import 'package:crypto/crypto.dart';
 import 'package:dio/dio.dart';
 import 'package:test/test.dart';
 import 'package:dio/adapter.dart';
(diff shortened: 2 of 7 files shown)


allanwolski-openco:
It would be great to have this

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -7,6 +7,8 @@ import '../dio_error.dart';
 import '../redirect_record.dart';
 
 typedef OnHttpClientCreate = HttpClient? Function(HttpClient client);
+typedef ResponseCertApprover = bool Function(
```
AlexV525, line 10: The function only represents a boolean value. I would prefer `CertificatePredicate`.

AlexV525 (commented):
Hi @timshadel, I left a comment about changing the type alias' name.

Are you willing to contribute by letting us pick the PR to another community-maintained `dio`? (We'll announce the new repo later once we merge PRs as much as possible.)

dio/lib/src/adapters/io_adapter.dart
timshadel: I took a deep dive in reaction to this question. Naming is important, and I think it deserves consideration. I looked at synonyms and definitions. In the end, I think we should change the typedef name, but I think we should follow existing Dio naming conventions. Consistency within a library is also incredibly important, in my opinion.

The [`RequestOptions` class](https://pub.dev/documentation/dio/latest/dio/RequestOptions-class.html) has a function that you can provide to check that the HTTP status code is acceptable. It's called [`ValidateStatus`](https://pub.dev/documentation/dio/latest/dio/ValidateStatus.html).

I think we should rename this to `ValidateCertificate` or maybe `ValidateResponseCertificate`.

The word "predicate" [has the precise definition](https://dcl-prog.stanford.edu/function-predicate.html) you mention, but functional programming is still niche, and sometimes intimidating. Rx-style work has widened that group, but the majority of developers will likely see it as a technical term that doesn't reflect the role this function has. (Even `CertificatePredicate` is too vague: does this predicate tell me if the last bit is a 1 or a 0? does it tell me if the host name is contained within the certificate?) The name should reflect the role it plays. I think `ValidateCertificate` conveys that role, implies a yes-or-no outcome, and provides parallelism with `ValidateStatus`.

What do you think? 😃

AlexV525: I agree with `ValidateCertificate` from your perspetive, and thanks for the rich context!

timshadel: Excellent, I'll contact you about the timing of this. Right now I'm thinking I can do it in about 2 weeks so we can migrate our dependencies internally.

timshadel: @jgoyvaerts Is there any chance that PRs like these will be merged? I see that #1590 was just recently merged. Lack of activity on this library is making a number of people nervous. Can we expect to see more PRs merging in the days ahead?

AlexV525 (approved):
The changes are LGTM with naming modifications, I'll go ahead and try to merge the changes in the community branch.

dio/test/exception_test.dart
```diff
@@ -45,6 +56,112 @@ void main() {
...
+        options: Options(validateStatus: (status) => true));
+    expect(response, isNotNull);
+  }, testOn: "!browser");
+
+  test('pinning: untrusted certificate tested and allowed', () async {
+    var dio = Dio();
+    // badCertificateCallback must allow the untrusted certificate through
+    (dio.httpClientAdapter as DefaultHttpClientAdapter).onHttpClientCreate =
+        (_) {
+      _.badCertificateCallback = (cert, host, port) => true;
```
AlexV525, line 109: The client is not returned here.

dio/test/exception_test.dart
```diff
@@ -1,9 +1,20 @@
...
 import 'package:dio/adapter.dart';
 
 void main() {
+  final trustedCertUrl = 'https://sha256.badssl.com/';
+  final untrustedCertUrl = 'https://wrong.host.badssl.com/';
+
+  // OpenSSL output like: SHA256 Fingerprint=EE:5C:E1:DF:A7:A4...
+  // All badssl.com hosts have the same cert, they just have TLS
+  // setting or other differences (like host name) that make them bad.
+  final lines = File('test/_pinning.txt').readAsLinesSync();
```
AlexV525, line 14: The filename for h2 is wrong, it should be `_pinning_http2.txt`.

AlexV525 (commented):
Tests didn't pass eventually, I've split them into separate tests to avoid tests on the Web.

The below comments are addressed by myself during the merge. It should not require further changes from you explicitly, and you'll see the PR once the repo is public.

AlexV525:
Hi @timshadel , I saw a package that does similar things: https://github.com/diefferson/http_certificate_pinning, can you identify the difference between your PR and the plugin? Thanks.

AlexV525:

PR #1578, opened 2022-10-16
Add resume download feature

PeterAkande:

- [ ] I have read the [Documentation](https://pub.dartlang.org/packages/dio)
- [ ] I have searched for a similar pull request in the [project](https://github.com/flutterchina/dio/pulls) and found none
- [ ] I have updated this branch with the latest `develop` to avoid conflicts (via merge from master or rebase)

This merge request fixes / refers to the following issues:  Add Resume download option

### Pull Request Description
Addition of resume download feature


dio/lib/src/entry/dio_for_native.dart
```diff
@@ -120,18 +121,40 @@ class DioForNative with DioMixin implements Dio {
...
+    // For the file resumption option.
+    // Its 0 for a new file.
+    int fileLength = 0;
+
+     if (resumeDownload) {
+      bool fileExists = await file.exists();
+
+      if (!fileExists) {
+        //For a resume uploads operation, the file must exist in the given directory.
+        //It it doesnt exist, throw an error.
```
AlexV525, line 133: I think it'll be better if we treat the condition as normal downloading, without throwing exceptions.

AlexV525 (commented):
Hi @PeterAkande, can you format files before it gets merged?

dio/lib/src/entry/dio_for_native.dart
PeterAkande: Ok.. would fix that.

PeterAkande: Hello @AlexV525 , just gave it a second thought. It might be better to throw an exception since the range might have been set in the header of the request. Having the range already set might lead to incomplete downloads.

It is written in the updated documentation that the range has to be specified in the header.

AlexV525: > It is written in the updated documentation that the range has to be specified in the header.

Yes that's the point. Users may not required to pass the range header once they enable the resume. Length detection should be done inside the download, as well as the write mode detection. WDYT?

PeterAkande: Yeah.. The range header shouldn't be passed by Users. it should be done in the download.
That would be a better approach.

I would work on that.

AlexV525: More thought on internally handling: If `download` automatically handles length, the flag might no longer be required.

PR #1586, opened 2022-10-30
Fixing multi value headers so they are serialized correctly

bradtalabat:

This merge request fixes / refers to the following issues: There was a change in version 4.0.1 that converted request header to their string values. This breaks multi value request headers as the get serialized like this `[value1,value2]` instead of like this `value1,value2`

### Pull Request Description
Change has been made to use the previous code, which worked in 4.0.0. This bug was found after upgrading from 4.0.0 to 4.0.6

### DIFF
--- a/dio/test/utils.dart
+++ b/dio/test/utils.dart
@@ -74,6 +74,17 @@ Future<void> startServer() async {
         return;
       }
 
+      if (path == '/multi-value-header') {
+        response.headers.contentType = ContentType('application', 'json');
+        response.headers.add("x-multi-value-request-header-echo", request.headers.value("x-multi-value-request-header").toString());
+        response
+          ..statusCode = 200
+          ..contentLength = -1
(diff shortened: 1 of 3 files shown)


dio/test/utils.dart
```diff
@@ -74,6 +74,17 @@ Future<void> startServer() async {
         return;
       }
 
+      if (path == '/multi-value-header') {
+        response.headers.contentType = ContentType('application', 'json');
+        response.headers.add("x-multi-value-request-header-echo", request.headers.value("x-multi-value-request-header").toString());
```
gerfalcon, line 79: Please, use single quotes here like this:

`        response.headers.add('x-multi-value-request-header-echo', request.headers.value('x-multi-value-request-header').toString());
`

gerfalcon (approved):

AlexV525:

PR #1587, opened 2022-11-01
Fix typo in `ProgressCallback` doc comment

tp:

Small typo in the doc comment for `ProgressCallback`.

### DIFF
--- a/dio/lib/src/options.dart
+++ b/dio/lib/src/options.dart
@@ -9,7 +9,7 @@ import 'utils.dart';
 /// [count] is the length of the bytes have been sent/received.
 ///
 /// [total] is the content length of the response/request body.
-/// 1.When receiving data:
+/// 1.When sending data:
 ///   [total] is the request body length.
 /// 2.When receiving data:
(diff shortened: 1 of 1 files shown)


AlexV525:
Hi @tp, can you explain the reason why it should be `sending`? IMO the callback will be only called when the corresponding data started to receive.

tp:
> Hi @tp, can you explain the reason why it should be `sending`? IMO the callback will be only called when the corresponding data started to receive.

Hey @AlexV525, the same callback definition is used for the sending (e.g. file upload) and response receiving progress:

CleanShot 2022-11-09 at 10 12 30@2x
source: https://pub.dev/documentation/dio/latest/dio/Dio/post.html

Thus the documentation lists the 2 distinct cases, the the value of `total` has a different meaning in each (outgoing vs. incoming bits).

AlexV525:
Hi @tp. We've made our repo public and published a new version of `dio`, named `diox`.
You can also see why we're working for a hardfork at https://github.com/cfug/diox/issues/29 and https://github.com/flutterchina/dio/pull/1550.

PR #1637, opened 2023-02-13
Merge changes from cfug/diox

AlexV525:
Merging all recent changes into the repo.


AlexV525:
Publishable tests are intended to be failed since we haven't published the new version for dio yet.

AlexV525:
@kuhnroyal Any ideas about why 2.15 tests failed with plugins?

kuhnroyal:
@AlexV525 Yea this happens because it is run with an older Dart version which does not support the use of `pubspec_overrides.yaml`.
Now that `5.0.0` is published, the build should work. It may fail again in the future if the plugins depend on a a new, unpublished version.

kuhnroyal:
I re-ran and everything is fine for now: https://github.com/cfug/dio/actions/runs/4160777691

PR #1656, opened 2023-02-15
fixes the assertion of the url host

### LINKED ISSUE #1655: Using Options BaseUrl assertion fails
### Package

dio

### Version

5.0.0

_No response_

### Dart Version

_No response_

### Steps to Reproduce

for example make a `Dio` client like this:

```
 static const String _baseURL = "https://example.com"

 final _http = Dio(
    BaseOptions(
      baseUrl: _baseURL,
      headers: {
        "accept": "application/json",
        "x-requested-With": "XMLHttpRequest",
        'content-type': 'application/json',
      },
    ),
  );
```

### Expected Result

the client should send the request and it works normally.

### Actual Result

```
Assertion failed:
options.dart:103
baseUrl.isEmpty || Uri.parse(baseUrl).host.isNotEmpty
is not true
```

AbanoubNassem:
the assertion of the url host , should check if it's empty .

fixes #1655

### DIFF
--- a/dio/lib/src/options.dart
+++ b/dio/lib/src/options.dart
@@ -100,7 +100,7 @@ class BaseOptions extends _RequestConfig with OptionsMixin {
     ResponseDecoder? responseDecoder,
     ListFormat? listFormat,
   })  : assert(connectTimeout == null || !connectTimeout.isNegative),
-        assert(baseUrl.isEmpty || Uri.parse(baseUrl).host.isNotEmpty),
+        assert(baseUrl.isEmpty || Uri.parse(baseUrl).host.isEmpty),
         super(
           method: method,
           receiveTimeout: receiveTimeout,


dio/lib/src/options.dart
```diff
@@ -100,7 +100,7 @@ class BaseOptions extends _RequestConfig with OptionsMixin {
     ResponseDecoder? responseDecoder,
     ListFormat? listFormat,
   })  : assert(connectTimeout == null || !connectTimeout.isNegative),
-        assert(baseUrl.isEmpty || Uri.parse(baseUrl).host.isNotEmpty),
+        assert(baseUrl.isEmpty || Uri.parse(baseUrl).host.isEmpty),
```
talisk, line 103: Maybe you should add a test case for it.

talisk (changes requested):

dio/lib/src/options.dart
kuhnroyal: Can you change the existing test here: https://github.com/cfug/dio/blob/main/dio/test/options_test.dart#L380 to test the actual assert when constructing the `BaseOptions` please. Currently we are just testing a Dart SDK function which doesn't help us.

AbanoubNassem:
sorry for the inconvenience, it was an issue from my side of configurations , since same configs used to work with older versions of `Dio`.

Issue #1661, opened 2023-02-16
There are many 401 Unauthorized  response in dio 5.0.0


alimaherofficial:
### Package

dio

### Version

5.0.0

### Output of `flutter doctor -v`

```shell
[!] Flutter (Channel stable, 3.7.3, on macOS 13.2 22D49 darwin-x64, locale en-EG)
    • Flutter version 3.7.3 on channel stable at /Users/ali/fvm/versions/stable
    ! Warning: `dart` on your path resolves to /usr/local/Cellar/dart/2.19.2/libexec/bin/dart, which is not inside your current Flutter SDK checkout at /Users/ali/fvm/versions/stable. Consider adding /Users/ali/fvm/versions/stable/bin to the front of your path.
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 9944297138 (7 days ago), 2023-02-08 15:46:04 -0800
    • Engine revision 248290d6d5
    • Dart version 2.19.2
    • DevTools version 2.20.1
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git" directly to perform update checks and upgrades.

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    • Android SDK at /Users/ali/Library/Android/sdk
    • Platform android-33-ext4, build-tools 33.0.2
    • Java binary at: /Applications/Android Studio.app/Contents/jre/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14C18
    • CocoaPods version 1.11.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2022.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)

[✓] VS Code (version 1.75.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.58.0

[...]

Issue #1653, opened 2023-02-15
[Notice] Set a proper `content-type` for requests when using Dio 5.0


AlexV525:
The issue intends to tell users that don't follow our migration guide well, then come to search for relevant issues. This issue will be opened and pinned until we release 5.1.x or 6.0.x.

## Context

We've discussed why we have removed the default `content-type` in https://github.com/cfug/diox/discussions/52. The implied `content-type` makes the maintainers hard to determine instantly if issues are related. So we decide to do less: remove all implied `content-type`.

## Summary

Since the 5.0 release, all options and transformers will no longer imply the default `content-type` according to your payload type. This means if you don't explicitly set the `content-type`, requests might resulting failure, and this widely affects most of your requests. In order to use 5.0.x, you'll need to configure the `content-type` properly.

Requests with the `FormData` payload requires no further migrations.

## How?

Reading the migration guide with the *"Set content-type manually"* section should be the first step. https://github.com/cfug/dio/blob/main/dio/migration_guide.md#set-content-type-manually

- Use `Headers.formUrlEncodedContentType` if your payloads are typically a `Map`.
- Use `Headers.jsonContentType` if your payloads are typically an encoded JSON string.
- Use `Headers.textPlainContentType` if your payloads are typically string without any structures.

### Quick handling

- Add the most used `content-type` to your `BaseOptions`. For example:
  ```dart
  dio.options.contentType = Headers.jsonContentType;
  ```
- Add `Options.contentType` for each requests with a payload. For example:
  ```dart
  dio.post(
    'your_url',
    data: jsonEncodedString,
    options: Options(contentType: Headers.jsonContentType),
  )
  ```

#### Using `Interceptor`

```dart
InterceptorsWrapper(
  onRequest: (
    RequestOptions options,
    RequestInterceptorHandler handler,
  ) {
    if (options.contentType == null) {
      final dynamic data = options.data;
      final String? contentType;
      if (data is FormData) {
        contentType = Headers.multipartFormDataContentType;
      } else if (data is Map) {
        contentType = Headers.formUrlEncodedContentType;
      } else if (data is String) {
        contentType = Headers.jsonContentType;
      } else if (data != null) {
        contentType = Headers.textPlainContentType; // Can be removed if unnecessary.
      } else {
        contentType = null;
      }
      options.contentType = contentType;
    }
    handler.next(options);
  },
)
```

HelenaSewell:
[...]

Issue #1650, opened 2023-02-14
"XX" in request body returns error


michaldev:
### Package

dio

### Version

5.0

### Steps to Reproduce

Just create query with parameters like below.

### Expected Result

This works very well in most cases, but when vehId is "23XX" i have error with parse json.
```
  Future<void> activateVehicle(String vehId) async {
    await dio.post("$host/api/diag/activate",
        data: {"vehicle_index": vehId, "seconds": 600});
  }
```

This works good with all cases:
```
var formData = {"vehicle_index": vehId, "seconds": 600};
await dio.post("$host/api/diag/activate", data: jsonEncode(formData));
```

I use backend created with Python and FastAPI.

### Actual Result

Error:
`"{"detail":[{"loc":["body",0],"msg":"Expecting value: line 1 column 1 (char 0)","type":"value_error.jsondecode","ctx":{"msg":"Expecting value","doc":"vehicle_index=23XX&seconds=600","pos":0,"lineno":1,"colno":1}}]}"`

michaldev:
I have json content type.

Issue #1648, opened 2023-02-14
Any http requests are not working with 5.0.0


bhanuka96:
### Package

dio

### Version

5.0.0

### Output of `flutter doctor -v`

```shell
Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
    • Android SDK at /Users/bhanukaisuru/Library/Android/sdk
    • Platform android-33, build-tools 32.1.0-rc1
    • Java binary at: /Library/Java/JavaVirtualMachines/jdk-18.0.1.1.jdk/Contents/Home/bin/java
    • Java version Java(TM) SE Runtime Environment (build 18.0.1.1+2-6)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14C18
    • CocoaPods version 1.11.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[!] Android Studio (version 2022.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    ✗ Unable to find bundled Java version.
    • Try updating or re-installing Android Studio.

[✓] VS Code (version 1.75.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.58.0

[✓] Connected device (3 available)
    • PRA LX2 (mobile) • 6MU7N17A12000618 • android-arm64  • Android 8.0.0 (API 26)
    • macOS (desktop)  • macos            • darwin-arm64   • macOS 13.2 22D49 darwin-arm64
    • Chrome (web)     • chrome           • web-javascript • Google Chrome 110.0.5481.77

[✓] HTTP Host Availability
    • All required HTTP hosts are available
```

### Steps to Reproduce

1. Upgrade to v5.0.0

any HTTP request not working with 5.0.0.

ex: 
**POST**

```
final response = await _dio.post('${env.cloudFunctionUrl}/createPayment', data: json, options: Options(headers: {'Authorization': 'Bearer $token'}));
```

**API**

```
  if (!req.body) {
    await cf.errorBooking(req.body, 'initBodyCheck')
[...]

PR #1671, opened 2023-02-18
✨ `ImplyContentTypeInterceptor`

### LINKED ISSUE #1667: Restore content-type auto-detection and only optionally disable it
### Request Statement

Hi,
in release 5.0.0 the content-type auto-detection logic. This seriously impact many applications relying on that logic (which, for many use-cases, was correct). SInce it was already possible to force manually content-type, would'nt be possible to specify a constructor flag in Dio to trigger the new logic and let the old one works flawlessly for compatibility with all existing code ?

Thank you,

### Solution Brainstorm

Pass parameter on constructor to explicitly disable the content-type auto-guess logic.

AlexV525:
Reverts the functionality of cfug/diox#69.

Resolves #1667 #1661 #1653 #1650 #1648 #1646

- The interceptor can be removed by calling `Interceptor.removeImplyContentTypeInterceptor`.
- The interceptor was intended not to be exposed to outside usages.

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -19,8 +19,6 @@
 
 ### Breaking Changes
 
-- Content type with `application/json` and `application/x-www-form-urlencoded`
-  will not be implied anymore in the transformer and the request option.
 - The default charset `utf-8` in `Headers` content type constants has been removed.
 - `BaseOptions.setRequestContentTypeWhenNoPayload` has been removed.
 - Improve `DioError`s. There are now more cases in which the inner original stacktrace is supplied.
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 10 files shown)


naamapps:
In my case if the body data is a Map then I want the content type to be implied as `application/json`, not `application/x-www-form-urlencoded`.
Can you please explain the decision to use `formUrlEncodedContentType` when the data is a Map?

AlexV525:
> Can you please explain the decision to use `formUrlEncodedContentType` when the data is a Map?

https://github.com/cfug/dio/blob/ac78e6151e1736f945cb9b215bbcfac230e19cf1/dio/lib/src/transformer.dart#L72-L85

AlexV525:
> In my case if the body data is a Map then I want the content type to be implied as `application/json`, not `application/x-www-form-urlencoded`.

You'll need to set as the JSON type in your case, which is the same before 5.0.0.

talisk (approved):
🚀

PR #1697, opened 2023-02-27
Add functionality tests for `ImplyContentTypeInterceptor`

skreborn:
The tests for `ImplyContentTypeInterceptor` were sorely lacking, so I've added a few basic expectations to test its behavior.

### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [x] I have added the required tests to prove the fix/feature I'm adding
- [x] I have updated the documentation (if necessary)
- [x] I have run the tests without failures
- [x] I have updated the `CHANGELOG.md` in the corresponding package (if necessary)

PR #1698, opened 2023-02-27
Fix "unsupported operation" error on web platform

skreborn:
For the web platform, an inline `bool.fromEnvironment` translates to `core.bool.fromEnvironment` instead of correctly being replaced by the constant value found in the environment at build time.

See https://github.com/flutter/flutter/issues/51186.

- [x] I have updated the `CHANGELOG.md` in the corresponding package (if necessary)

This fix is unfortunately not something we can test for as far as I can tell. I've tried numerous methods of automating the detection of the issue (and thus the fix), but it seems to only appear when running the Flutter example in a browser.

### DIFF
--- a/dio/lib/src/interceptors/imply_content_type.dart
+++ b/dio/lib/src/interceptors/imply_content_type.dart
@@ -5,6 +5,13 @@ import '../form_data.dart';
 import '../headers.dart';
 import '../options.dart';
 
+// For the web platform, an inline `bool.fromEnvironment` translates to
+// `core.bool.fromEnvironment` instead of correctly being replaced by the
+// constant value found in the environment at build time.
+//
+// See https://github.com/flutter/flutter/issues/51186.
+const _kReleaseMode = bool.fromEnvironment('dart.vm.product');
+
 /// {@template dio.interceptors.ImplyContentTypeInterceptor}
 /// The default `content-type` for requests will be implied by the
 /// [ImplyContentTypeInterceptor] according to the type of the request payload.
@@ -28,7 +35,7 @@ class ImplyContentTypeInterceptor extends Interceptor {
         contentType = Headers.jsonContentType;
       } else {
         // Do not log in the release mode.
-        if (!bool.fromEnvironment('dart.vm.product')) {
+        if (!_kReleaseMode) {
           dev.log(
             '${data.runtimeType} cannot be used'
             'to imply a default content-type, '
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix "unsupported operation" error on web platform.
+
 ## 5.0.1
 
 - Add `ImplyContentTypeInterceptor` as a default interceptor.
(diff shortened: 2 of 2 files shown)


ueman:
We could execute the test suite in a browser environment

skreborn:
> We could execute the test suite in a browser environment

I've tried running `dart test . --platform=chrome --debug` within `dio` but unfortunately it did not trigger the error - I'm unsure why as of now, but I'll investigate further.

skreborn:
Looking at the generated output; it seems `dart test` correctly transpiles the snippet in question by eliminating `bool.fromEnvironment` entirely.

```javascript
if (t1) {
  if (data instanceof A.FormData)
    contentType = "multipart/form-data";
  else if (type$.Map_dynamic_dynamic._is(data) || typeof data == "string")
    contentType = "application/json";
  else {
    J.get$runtimeType$(data).toString$0(0);
    A.StackTrace_current();
    contentType = null;
  }
  options.set$contentType(0, contentType);
}
```

I'm assuming this happens because `dart test` uses `dart compile js` in the background, which does not have a flag to compile in debug mode (and reducing the optimization level makes no difference, either).

```text
Compile Dart to JavaScript.

Usage: dart compile js [arguments] <dart entry point>
  -h, --help      Print this usage information (add -v for information about all options).
  -o, --output    Write the output to <file name>.
  -O<0,1,2,3,4>   Set the compiler optimization level (defaults to -O1).
```

Using [webdev](https://dart.dev/tools/webdev) (which uses the Dart Development Compiler), however, we can get our hands on the erroneous output.

```javascript
if (data != null && options.contentType == null) {
  let contentType = null;
  if (form_data.FormData.is(data)) {
    contentType = "multipart/form-data";
  } else if (core.Map.is(data) || typeof data == 'string') {
    contentType = "application/json";
  } else {
    if (!core.bool.fromEnvironment("dart.vm.product")) {
      developer.log(dart.str(dart.runtimeType(data)) + " cannot be used" + "to imply a default content-type, " + "please set a proper content-type in the request.", {level: 900, name: "🔔 Dio", stackTrace: core.StackTrace.current});
    }
    contentType = null;
  }
  options.contentType = contentType;
}
```

Finally, following the [official guide](https://dart.dev/tools/webdev#test) on how to run web tests - and modifying `MockAdapter` to not use `dart:io` -, I've managed to make it throw when running the "ignores unsupported data type" test added in https://github.com/cfug/dio/pull/1697.

dio/lib/src/interceptors/imply_content_type.dart
```diff
@@ -5,6 +5,8 @@ import '../form_data.dart';
 import '../headers.dart';
 import '../options.dart';
 
+const _kReleaseMode = bool.fromEnvironment('dart.vm.product');
```
ueman, line 8: Can you add the PR description as a comment here? That way nobody changes it back, hopefully, because they were warned.

ueman (commented):
Thanks for this and the throughout research! I appreciate it.

dio/lib/src/interceptors/imply_content_type.dart
skreborn: Of course, that makes sense.

ueman (approved):

mauriziopinotti:
@ueman @skreborn any plan to release 5.0.2 with this fix in the near future?
Thanks!

AlexV525:
> @ueman @skreborn any plan to release 5.0.2 with this fix in the near future? Thanks!

5.0.2 is up.

PR #1710, opened 2023-03-05
🐛 Fix `DioErrorType.cancel` in `Interceptors`

### LINKED ISSUE #1512: Interceptor doesn't work for DioError.type = cancel
Interceptor have no effect on dio error when you cancel request.

**version**: 4.0.6

**Example**

```
class ErrorInterceptor extends Interceptor {
  @override
  void onError(DioError err, ErrorInterceptorHandler handler) {
    if (err.type == DioErrorType.cancel) {
      handler.next(CancelledException());
    } else {
      handler.next(err);
    }
  }
}
```

When register interceptor and try to catch an error it always has a type = `DioError` instead of `CancelledException` whenever request is cancelled.

AlexV525:
Fixes #1512.

Previously, `listenCancelForAsyncTask` will wait for `CancelToken.whenCancel` in certain scenarios:
https://github.com/cfug/dio/blob/6544701ed370e6ceaf81576f393459605e9a612f/dio/lib/src/dio_mixin.dart#L457-L466

These scenarios include `DioErrorType.cancel`, in which the error handler has been put into the event loop rather than executed immediately, which means the handler will handle the exception after the request is completed.

### DIFF
--- a/dio/test/interceptor_test.dart
+++ b/dio/test/interceptor_test.dart
@@ -469,6 +469,40 @@ void main() {
     });
   });
 
+  group('Error Interceptor', () {
+    test('handled when request cancelled', () async {
+      final cancelToken = CancelToken();
+      DioError? iError, qError;
+      final dio = Dio()
+        ..httpClientAdapter = MockAdapter()
+        ..options.baseUrl = MockAdapter.mockBase
+        ..interceptors.add(
+          InterceptorsWrapper(
+            onError: (DioError e, ErrorInterceptorHandler handler) {
+              iError = e;
+              handler.next(e);
+            },
+          ),
+        )
+        ..interceptors.add(
+          QueuedInterceptorsWrapper(
+            onError: (DioError e, ErrorInterceptorHandler handler) {
+              qError = e;
+              handler.next(e);
+            },
+          ),
+        );
+      Future.delayed(const Duration(seconds: 1)).then((_) {
+        cancelToken.cancel('test');
+      });
+      await dio
+          .get('/test-timeout', cancelToken: cancelToken)
+          .then((_) {}, onError: (_) {});
+      expect(iError, isA<DioError>());
+      expect(qError, isA<DioError>());
+    });
+  });
+
   group('QueuedInterceptor', () {
     test('requests ', () async {
       String? csrfToken;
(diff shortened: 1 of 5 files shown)


dio/test/mock/adapters.dart
```diff
@@ -32,32 +32,30 @@ class MockAdapter implements HttpClientAdapter {
             },
           );
         case '/test-auth':
-          {
```
AlexV525, line 35: A format change to remove the `Set` caller.

dio/test/mock/adapters.dart
```diff
@@ -70,23 +68,23 @@ class MockAdapter implements HttpClientAdapter {
               },
             );
           });
-
         case '/token':
-          {
```
AlexV525, line 75: Same as above.

dio/test/mock/adapters.dart
```diff
@@ -70,23 +68,23 @@ class MockAdapter implements HttpClientAdapter {
...
+            200,
+            headers: {
+              Headers.contentTypeHeader: [Headers.jsonContentType],
+            },
+          );
         case '/test-force-convert':
           return ResponseBody.fromString('{"code":0,"result":"ok"}', 200);
+        case '/test-timeout':
+          await Future.delayed(const Duration(days: 365));
+          return ResponseBody.fromString('', 200);
```
AlexV525, line 87: Adds a timeout handler that is intended to be timed out in the whole year.

dio/test/interceptor_test.dart
```diff
@@ -469,6 +469,40 @@ void main() {
...
+      DioError? iError, qError;
+      final dio = Dio()
+        ..httpClientAdapter = MockAdapter()
+        ..options.baseUrl = MockAdapter.mockBase
+        ..interceptors.add(
+          InterceptorsWrapper(
+            onError: (DioError e, ErrorInterceptorHandler handler) {
+              iError = e;
+              handler.next(e);
+            },
```
AlexV525, line 484: Actually, the caller will be called previously, but it'll be slightly after the expectation. If we add any delay after awaiting the request, the test will be passed.

chenglu (approved):

PR #1719, opened 2023-03-07
🎨 Using shared analysis options for all packages

AlexV525:
Uses the base analysis options for all packages to avoid redundant definitions.


AlexV525:
Unfortunately, I don't have ideas about how to test compute with these changes.

kuhnroyal:
It may be better to just use https://pub.dev/packages/lints and adjust a few rule that we want additionally or don't want? This way new rules would be used automatically.

AlexV525:
> It may be better to just use https://pub.dev/packages/lints and adjust a few rule that we want additionally or don't want? This way new rules would be used automatically.

That's for sure, but it cannot be used in a base config like this, since we don't have a pubspec here. Otherwise we may revert the config change and change dart files only in this PR.

AlexV525:
I walked around other monorepos like fluttefire, flutter_local_notification, and melos itself, adding a pubspec seems fine here.

ueman:
We could symlink the same file to all repos. Then we could use imports. I am not sure whether that’s a good idea.

AlexV525:
> We could symlink the same file to all repos. Then we could use imports. I am not sure whether that’s a good idea.

We should avoid symlinks because it usually breaks the development on Windows. Added new commits.

pubspec.yaml
```diff
@@ -0,0 +1,10 @@
+name: dio_workspace
```
AlexV525, line 1: - https://github.com/invertase/melos/blob/main/pubspec.yaml
- https://github.com/firebase/flutterfire/blob/master/pubspec.yaml

kuhnroyal: Yea they all do this because they use `melos`. But I think it is a good idea.

kuhnroyal: But this involves us changing Github workflows to run `pub get` in the root repository before anything else.

kuhnroyal:
Tbh I think we can just add `lints` to every package. There should not be that much customization of rules required.

AlexV525:
> But this involves us changing Github workflows to run `pub get` in the root repository before anything else.

I'm surprised not seing any exceptions in tests. Maybe it only affects the local development.

> Tbh I think we can just add `lints` to every package. There should not be that much customization of rules required.

That's what we have done previously.

AlexV525:
> It works perfectly fine without the `pubspec.yaml` in root for me.

This is what happened if you don't have a pubspec in the root which doesn't have `.dart_tool` too.
image

kuhnroyal:
> > It works perfectly fine without the `pubspec.yaml` in root for me.
> 
> This is what happened if you don't have a pubspec in the root which doesn't have `.dart_tool` too. image

Yea but this is only when you open it right? We can just name it differently and it is probably fine. Something like `analysis_options_base.yaml` might work.

AlexV525:
Deleted and scheduled for another run.

chenglu (approved):
LGTM!

PR #1721, opened 2023-03-07
Interceptors now gets the StackTrace in the DioError objects

lohnn:
### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [x] I have added the required tests to prove the fix/feature I'm adding
- [x] I have updated the documentation (if necessary)
- [x] I have run the tests without failures
- [x] I have updated the `CHANGELOG.md` in the corresponding package

### Additional context and info (if any)

The StackTrace will be the same one as you get when surrounding the request with try-catch.

Fixes #1173.

PR #1722, opened 2023-03-07
[dio] Fix missing source stacktrace in DioError

kuhnroyal:
[Improve nullability in DioMixin.assureResponse](https://github.com/cfug/dio/commit/8418c72a7d78b944bcfcd54d022de996fc717ed3)
* there should never be a case where there are not `RequestOptions` available
* this is technically a breaking change but we should instead mark this `@internal` in the future

[Fix missing source stacktrace in DioError](https://github.com/cfug/dio/commit/4f4ffe44b12f444fda9c912dec8676df1a3aa215) 

* a `DioError` should always have a meaningful stacktrace which points to the actual invocation of `Dio.get/post/xxx`
* `Dio` historically lost a lot of this source information due to a multitude of asynchronous calls
* now the source stacktrace is consistently being set into the `RequestOptions` instance and can later be retrieved and used when a `DioError` is constructed
* additionally a `DioError` may contain another cause which may itself contain a separate stackTrace with more detailed information

The existing behavior which was somewhat correct got lost in https://github.com/cfug/dio/pull/1405 which was merged in the temporary diox repository.


AlexV525:
Is it another approach to fixing stack trace in errors or it contains the fix that #1721 does overall?

dio/lib/src/dio_mixin.dart
```diff
@@ -706,13 +701,13 @@ abstract class DioMixin implements Dio {
   }
 
   static Response<T> assureResponse<T>(
-    Object response, [
-    RequestOptions? requestOptions,
-  ]) {
+    Object response,
+    RequestOptions requestOptions,
+  ) {
```
AlexV525, line 706: I cannot evaluate the affection by either breaking this or making it `@internal`. However, an exposed API should be considered as used by everyone. Can we keep the method and add another underscored method to pass through?

ueman:
I feel like we would improve the current code base and the exceptions if we improve the inner workings of dio, instead of trying to attach the correct stacktraces

kuhnroyal:
I agree but currently it is just not possible to trace any error to it's source. And this broke in 5.0.0
This is a minimal invasive approach to add the old behavior back.
I don't see anyone rewriting the inner workings anytime soon and even if, then you will still have a lot of async gaps that break the stack traces.

dio/lib/src/dio_mixin.dart
```diff
@@ -359,19 +359,16 @@ abstract class DioMixin implements Dio {
...
-    final requestOptions = options.compose(
+    final requestOptions = (options ??= Options()).compose(
       this.options,
       path,
       data: data,
       queryParameters: queryParameters,
       onReceiveProgress: onReceiveProgress,
       onSendProgress: onSendProgress,
       cancelToken: cancelToken,
+      sourceStackTrace: StackTrace.current,
```
kuhnroyal, line 370: This is the actual stack trace that matters in order to find out where the client code initiated the request.
After this point inside `fetch` there are async gaps and then the stack trace will look somewhat like this:

```
  File "dio_mixin.dart", line 643, in DioMixin.fetch.<fn>
  File "zone.dart", line 1422, in _rootRunBinary
  File "zone.dart", line 1314, in _CustomZone.runBinary
  File "future_impl.dart", line 162, in _FutureListener.handleError
  File "future_impl.dart", line 779, in Future._propagateToListeners.handleError
  File "future_impl.dart", line 800, in Future._propagateToListeners
  File "future_impl.dart", line 575, in Future._completeError
  File "future_impl.dart", line 51, in _SyncCompleter._completeError
  File "future_impl.dart", line 23, in _Completer.completeError
  File "future.dart", line 617, in Future.any.onError
  File "zone.dart", line 1422, in _rootRunBinary
  File "zone.dart", line 1314, in _CustomZone.runBinary
  File "future_impl.dart", line 162, in _FutureListener.handleError
  File "future_impl.dart", line 779, in Future._propagateToListeners.handleError
  File "future_impl.dart", line 800, in Future._propagateToListeners
  File "future_impl.dart", line 575, in Future._completeError
  File "future_impl.dart", line 520, in Future._chainForeignFuture.<fn>
  File "zone.dart", line 1398, in _rootRun
  File "zone.dart", line 1300, in _CustomZone.run
  File "zone.dart", line 1208, in _CustomZone.runGuarded
  File "zone.dart", line 1248, in _CustomZone.bindCallbackGuarded.<fn>
  File "schedule_microtask.dart", line 40, in _microtaskLoop
  File "schedule_microtask.dart", line 49, in _startMicrotaskLoop
  ```
  
  And you have no idea where it came from.

**Comment by kuhnroyal** [comment]
But yea this will currently not work when any interceptor manually creates a new instance of `RequestOptions`.

**Comment by kuhnroyal** [comment]
I read a little bit last night. This has nothing to do with how dio is written. There is just no way to avoid the async gaps when `async/await` is used. There is always a new stacktrace from the moment the async call is started and at that point the context of the original call site of `Dio().xxx` is lost.
But this is the only important information any dio client/user needs, a stacktrace which starts somewhere inside `DioMixin.fetch` helps nobody.

> I feel like we would improve the current code base and the exceptions if we improve the inner workings of dio

So as a result I don't think this is an option.

There is https://pub.dev/packages/stack_trace which uses zone magic to combine stacktraces across async gaps. This can combine the stacktrace(s) correctly but it also comes with the performance overhead of zones and a different string representation of the stacktraces. I created https://github.com/cfug/dio/pull/1723 tho showcase this.

**Comment by kuhnroyal** [comment]
We need to discuss how to proceed here. https://github.com/cfug/dio/pull/1721 also attempts to fix a smaller part of the same problem. There is no simple solution for the async gaps but IMHO dio needs to ensure that the stacktraces in DioError are available and useful.

**Review by ueman — COMMENTED** [review]
I prefer this approach over https://github.com/cfug/dio/pull/1723 because it's less invasive. Chains introduce zones which have a negative impact on the performance and zones are hard to deal with. Tho, I would love the improved stacktraces it brings with it.

**Comment by kuhnroyal** [comment]
I will see if I can add some tests similar to those in #1723 so we can make sure that stacktraces are not lost again.

**Comment by kuhnroyal** [comment]
I think this is ready

File: dio/lib/src/dio_mixin.dart [hunk 3]
```diff
@@ -359,19 +361,16 @@ abstract class DioMixin implements Dio {
...
     );
-    requestOptions.onReceiveProgress = onReceiveProgress;
-    requestOptions.onSendProgress = onSendProgress;
-    requestOptions.cancelToken = cancelToken;
```
AlexV525, line 374: Are these unnecessary anymore? I didn't see new coding applying this.l

dio/test/dio_mixin_test.dart
```diff
@@ -3,13 +3,15 @@ import 'package:test/test.dart';
...
     final untypedResponse = Response<dynamic>(
-      requestOptions: RequestOptions(path: ''),
+      requestOptions: requestOptions,
       data: null,
     );
     expect(untypedResponse is Response<int?>, isFalse);
 
-    final typedResponse = DioMixin.assureResponse<int?>(untypedResponse);
+    final typedResponse =
+        DioMixin.assureResponse<int?>(untypedResponse, requestOptions);
```
AlexV525, line 14: I should definitely apply the `require_trailing_commas` lint rule. :)

dio/lib/src/cancel_token.dart
```diff
@@ -31,8 +31,12 @@ class CancelToken {
 
   /// Cancel the request with the given [reason].
   void cancel([Object? reason]) {
+    if (requestOptions == null) {
+      throw StateError(
+          'CancelToken was canceled before being used in a request!');
+    }
     _cancelError = DioError.requestCancelled(
-      requestOptions: requestOptions ?? RequestOptions(),
+      requestOptions: requestOptions!,
```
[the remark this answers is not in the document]
kuhnroyal: I don't think that would be good. This is the only place where this could be relevant but I think above solution is fine.

dio/lib/src/dio_mixin.dart
kuhnroyal: They are already passed to `Options.compose` just above.

dio/lib/src/dio_mixin.dart
```diff
@@ -359,19 +361,16 @@ abstract class DioMixin implements Dio {
     ProgressCallback? onSendProgress,
     ProgressCallback? onReceiveProgress,
   }) async {
-    options ??= Options();
-    final requestOptions = options.compose(
+    final requestOptions = (options ??= Options()).compose(
```
kuhnroyal, line 364:
```suggestion
    final requestOptions = (options ?? Options()).compose(
```

dio/lib/src/dio_error.dart
```diff
@@ -56,78 +56,99 @@ extension _DioErrorTypeExtension on DioErrorType {
...
     this.type = DioErrorType.unknown,
     this.error,
     StackTrace? stackTrace,
     this.message,
-  }) : stackTrace = identical(stackTrace, StackTrace.empty) ? null : stackTrace;
+  }) : stackTrace = identical(stackTrace, StackTrace.empty)
+            ? (requestOptions.sourceStackTrace ?? StackTrace.current)
+            : (stackTrace ??
+                requestOptions.sourceStackTrace ??
+                StackTrace.current);
```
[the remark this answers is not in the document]
kuhnroyal: Fixed

dio/test/dio_mixin_test.dart
kuhnroyal: I added it manually here.

dio/lib/src/cancel_token.dart
```diff
@@ -31,8 +31,12 @@ class CancelToken {
 
   /// Cancel the request with the given [reason].
   void cancel([Object? reason]) {
+    if (requestOptions == null) {
+      throw StateError(
+          'CancelToken was canceled before being used in a request!');
```
AlexV525, line 36:
```suggestion
      throw StateError(
        'CancelToken was canceled before being used in a request.',
      );
```

dio/lib/src/dio_error.dart
```diff
@@ -56,78 +56,99 @@ extension _DioErrorTypeExtension on DioErrorType {
...
+  DioError({
     required this.requestOptions,
     this.response,
     this.type = DioErrorType.unknown,
     this.error,
     StackTrace? stackTrace,
     this.message,
-  }) : stackTrace = identical(stackTrace, StackTrace.empty) ? null : stackTrace;
+  }) : stackTrace = identical(stackTrace, StackTrace.empty)
+            ? (requestOptions.sourceStackTrace ?? StackTrace.current)
```
AlexV525, line 67: These parentheses are still redundant. :P

dio/test/cancel_token_test.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:dio/dio.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group(CancelToken, () {
+    test('cancel returns the correct DioError', () async {
+      final token = CancelToken();
+      const reason = 'cancel';
+
+      expectLater(token.whenCancel, completion((error) {
```
AlexV525, line 10: I've seen multiple forms with the predicate statement in our tests, such as `equals`, `(e) => e...`, and `completion`. 🤣

AlexV525 (approved):
LGTM with above nits

lohnn:
As this PR seems to handle the issue I intended to solve in #1721, but also solves a bigger issue as well, maybe you could take the two tests from my PR to hopefully catch future regressions in regards to stack traces in interceptors?

AlexV525:
I still got an invalid stack using the branch to test with #1749.

With the branch:
image

The real stack should be:
image

kuhnroyal:
> I still got an invalid stack using the branch to test with #1749.
> 
> With the branch: image
> 
> The real stack should be: image

I don't now, should it really be that stackTrace? You don't have any indication of the call of your code in it. You see that `jsonDecode` failed but nothing more.

AlexV525:
> I don't know, should it really be that stackTrace? You don't have any indication of the call of your code in it. You see that `jsonDecode` failed but nothing more.

IMO it should be. When you trying with #1749 with the previous code, `catchError((e, s) {})`'s stack can obtain the correct stack (as in my last image). But with this fix, the stack is completely gone even if we add back the stack trace argument and I have no idea who cause the format fail.

kuhnroyal:
Well, that is the trade off. It would be the same with https://github.com/cfug/dio/pull/1721 - in the interceptor you would not see the `jsonEncode` stack trace.

I guess we need a discussion about this:
> a `DioError` should always have a meaningful stacktrace which points to the actual invocation of `Dio.get/post/xxx`

AlexV525:
It just seems not pointed to the real request.

image

> * a `DioError` should always have a meaningful stacktrace which points to the actual invocation of `Dio.get/post/xxx`

kuhnroyal:
> It just seems not pointed to the real request.

Yea that will not work, unless we can use `Error.throwWithStackTrace`. If we `throw DioError(...)` then the stackTrace that you catch will always point to this line. But `DioError.stackTrace` should be correct and reference your test file.

kuhnroyal:
But we still have a problem if `DioError.error is Exception` - in this case we lose the original stackTrace from where the exception was thrown. The only way I see is, we wrap it again, which is horrible.

AlexV525:
Sounds like we should have an issue to track for problems with exceptions, about what shall we do, what we've done, and when to do something.

I'll leave this PR to @ueman to decide what's next.

ueman:
I recognize that we're right probably not able to always have a useful stacktrace due to the various reasons already laid out. I also think we (well not me but you guys, @kuhnroyal and @AlexV525) tried our best at getting a useful stacktrace here, but it might simply not be possible to satisfy every use case.

FWIW, I think the stacktrace of the dioerror should point to the invocation of get, post, etc and the inner stacktrace should always be the stacktrace of the inner exception. But since we can't use `Error.throwWithStackTrace` yet.

To conclude, as long as the errors got at least somewhat better than before, I'm happy with this PR. We don't need to get it perfect right now, since we can always iterate and improve later on.

Especially with Dart 3 right around the corner we have a good opportunity to revamp how we deal with exception and stacktraces (as well as other parts if needed). @kuhnroyal already started a discussion for that over at https://github.com/cfug/dio/discussions/1753

kuhnroyal:
I will resolve the conflicts and add the tests suggested by @lohnn - then we can release this as 5.1.0 I would suggest.

dio/lib/src/dio_mixin.dart
```diff
@@ -447,10 +444,10 @@ abstract class DioMixin implements Dio {
     FutureOr<dynamic> Function(dynamic, StackTrace) errorInterceptorWrapper(
```
AlexV525, line 444: The stack trace parameter is no longer needed. And the error type can be promoted to `Object`.

```suggestion
    FutureOr<dynamic> Function(Object) errorInterceptorWrapper(
```

kuhnroyal: No, this doesn't work, there are dynamic calls after this..

AlexV525: Oh right, but the stack trace should be available to remove.

kuhnroyal: I refactored this a bit, so now it works.

dio/lib/src/dio_error.dart
kuhnroyal: Removed

kuhnroyal:
We should wait with the merge until we have released another hotfix 5.0.4 with some encoding fixes.

ueman (approved):
Thanks @kuhnroyal

AlexV525 (approved):
Great work with this long history!

dio/lib/src/cancel_token.dart
```diff
@@ -31,8 +31,12 @@ class CancelToken {
 
   /// Cancel the request with the given [reason].
   void cancel([Object? reason]) {
+    if (requestOptions == null) {
```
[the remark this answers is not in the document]
kuhnroyal: https://github.com/cfug/dio/issues/1765 - @ueman What do you think now :)

PR #1738, opened 2023-03-14
Avoid new cookies overwritten by previous cookies

jasonlaw:
Fixed the bug that the new cookie will be overwritten by the previous cookie.

### DIFF
--- a/plugins/cookie_manager/CHANGELOG.md
+++ b/plugins/cookie_manager/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-*None.*
+- Allow `Set-Cookie` to be parsed in redirect responses.
 
 ## 2.1.2
 
--- a/plugins/cookie_manager/CHANGELOG.md
+++ b/plugins/cookie_manager/CHANGELOG.md
@@ -3,10 +3,7 @@
 ## Unreleased
 
 - Allow `Set-Cookie` to be parsed in redirect responses.
-
-## 2.1.3
-
-- Fix new cookie being replaced by old cookie with the same name.
+- Fix new cookies being replaced by old cookies with the same name.
 - Sort the cookie by path (longer path first).
 
 ## 2.1.2
--- a/plugins/cookie_manager/CHANGELOG.md
+++ b/plugins/cookie_manager/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 - Allow `Set-Cookie` to be parsed in redirect responses.
 
+## 2.1.3
+
+- Fix new cookie being replaced by old cookie with the same name.
+- Sort the cookie by path (longer path first).
+
 ## 2.1.2
 
 - Fix empty cookie parsing and header value set.
--- a/plugins/cookie_manager/lib/src/cookie_mgr.dart
+++ b/plugins/cookie_manager/lib/src/cookie_mgr.dart
@@ -38,11 +38,11 @@ class CookieManager extends Interceptor {
       final previousCookies =
           options.headers[HttpHeaders.cookieHeader] as String?;
       final newCookies = getCookies([
-        ...cookies,
         ...?previousCookies
             ?.split(';')
             .where((e) => e.isNotEmpty)
             .map((c) => Cookie.fromSetCookieValue(c)),
+        ...cookies,
       ]);
       options.headers[HttpHeaders.cookieHeader] =
           newCookies.isNotEmpty ? newCookies : null;
--- a/plugins/cookie_manager/lib/src/cookie_mgr.dart
+++ b/plugins/cookie_manager/lib/src/cookie_mgr.dart
@@ -30,13 +30,19 @@ class CookieManager extends Interceptor {
   /// Merge cookies into a Cookie string.
   /// Cookies with longer paths are listed before cookies with shorter paths.
   static String getCookies(List<Cookie> cookies) {
-    final uniqueNames = <String>{};
-    // Filter out cookie with duplicate name.
-    cookies = cookies.where((cookie) => uniqueNames.add(cookie.name)).toList();
     // Sort cookies by path (longer path first).
-    cookies.sort(
-      (a, b) => (b.path?.length ?? 0).compareTo(a.path?.length ?? 0),
-    );
+    cookies.sort((a, b) {
+      if (a.path == null && b.path == null) {
+        return 0;
+      } else if (a.path == null) {
+        return -1;
+      } else if (b.path == null) {
+        return 1;
+      } else {
+        return (b.path!.length).compareTo(a.path!.length);
+      }
+    });
+
     return cookies.map((cookie) => '${cookie.name}=${cookie.value}').join('; ');
   }
 
@@ -46,11 +52,11 @@ class CookieManager extends Interceptor {
       final previousCookies =
           options.headers[HttpHeaders.cookieHeader] as String?;
       final newCookies = getCookies([
-        ...cookies,
         ...?previousCookies
             ?.split(';')
             .where((e) => e.isNotEmpty)
             .map((c) => Cookie.fromSetCookieValue(c)),
+        ...cookies,
       ]);
       options.headers[HttpHeaders.cookieHeader] =
           newCookies.isNotEmpty ? newCookies : null;
--- a/plugins/cookie_manager/lib/src/cookie_mgr.dart
+++ b/plugins/cookie_manager/lib/src/cookie_mgr.dart
@@ -33,12 +33,10 @@ class CookieManager extends Interceptor {
     final uniqueNames = <String>{};
     // Filter out cookie with duplicate name.
     cookies = cookies.where((cookie) => uniqueNames.add(cookie.name)).toList();
-
     // Sort cookies by path (longer path first).
     cookies.sort(
       (a, b) => (b.path?.length ?? 0).compareTo(a.path?.length ?? 0),
     );
-
     return cookies.map((cookie) => '${cookie.name}=${cookie.value}').join('; ');
   }
 
--- a/plugins/cookie_manager/lib/src/cookie_mgr.dart
+++ b/plugins/cookie_manager/lib/src/cookie_mgr.dart
@@ -29,6 +29,10 @@ class CookieManager extends Interceptor {
 
   /// Merge cookies into a Cookie string.
   static String getCookies(List<Cookie> cookies) {
+    // remove previous cookie with same name
+    cookies.removeWhere((cookie) =>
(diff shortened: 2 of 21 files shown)


AlexV525:
Hi @jasonlaw, I saw you didn't check any checkboxes so I turn this into a draft. Please add a unit test for the change. Thanks!

plugins/cookie_manager/lib/src/cookie_mgr.dart
```diff
@@ -29,6 +29,10 @@ class CookieManager extends Interceptor {
 
   /// Merge cookies into a Cookie string.
   static String getCookies(List<Cookie> cookies) {
+    // remove previous cookie with same name
+    cookies.removeWhere((cookie) =>
+        cookies.indexWhere((c) => c.name == cookie.name) <
+        cookies.indexOf(cookie));
```
AlexV525, line 35: I haven't got this before, but as I look into the [RFC 6265](https://www.rfc-editor.org/rfc/rfc6265), we should probably sort cookies by longer paths too, as described in the [Section 5.4](https://www.rfc-editor.org/rfc/rfc6265#section-5.4).

jasonlaw: Agreed. 
Btw, in the same document, it also stated that we should sort by creation-times in case there is equal length path fields. I don't aware there is a creation-times in cookies. I am wondering how should we handle this?

AlexV525: We don't, which means cookies first in should stay their position if the later one has the same length of the path.

jasonlaw: I have added the sorting logic and new unit test. Please review.

plugins/cookie_manager/lib/src/cookie_mgr.dart
```diff
@@ -29,6 +29,17 @@ class CookieManager extends Interceptor {
 
   /// Merge cookies into a Cookie string.
   static String getCookies(List<Cookie> cookies) {
+    // remove previous cookie with same name
```
AlexV525, line 32: Capitalize the first word for comments and `.` at the trailing.

plugins/cookie_manager/lib/src/cookie_mgr.dart
```diff
@@ -29,6 +29,17 @@ class CookieManager extends Interceptor {
...
+    cookies = cookies.fold<List<Cookie>>(
+        [],
+        (prev, current) => prev.indexWhere((c) => c.name == current.name) == -1
+            ? [...prev, current]
+            : prev);
```
AlexV525, line 37: Can you use control flows rather than inline APIs, which will make the code more readable and reliable?

plugins/cookie_manager/lib/src/cookie_mgr.dart
```diff
@@ -29,6 +29,17 @@ class CookieManager extends Interceptor {
...
+
+    // sort cookies by path (longer path first)
+    cookies
+        .sort((a, b) => (b.path?.length ?? 0).compareTo(a.path?.length ?? 0));
```
AlexV525, line 41:
```suggestion
    cookies.sort(
      (a, b) => (b.path?.length ?? 0).compareTo(a.path?.length ?? 0),
    );
```

plugins/cookie_manager/test/cookies_test.dart
```diff
@@ -111,4 +111,35 @@ void main() {
       expect(response.requestOptions.headers[HttpHeaders.cookieHeader], null);
     });
   });
+  test('testing redundant cookies', () async {
```
AlexV525, line 114:
```suggestion

  test('cookies replacement', () async {
```

plugins/cookie_manager/lib/src/cookie_mgr.dart
```diff
@@ -28,7 +28,25 @@ class CookieManager extends Interceptor {
...
 
   /// Merge cookies into a Cookie string.
+  /// Cookies with longer paths are listed before cookies with shorter paths.
   static String getCookies(List<Cookie> cookies) {
+    // Remove previous cookie with same name.
+    cookies = cookies.fold<List<Cookie>>([], (prev, current) {
+      // Check if the current cookie's name is already in the list.
+      if (prev.any((c) => c.name == current.name)) {
+        // Skip this cookie.
+        return prev;
```
AlexV525, line 38: Deeply reviewing these codes, I think it'll be better if we use `[...previousCookies, ...cookies]` and return the new cookie here. Even though they are the same theoretically at present, it still causes confusion when reading these codes.

plugins/cookie_manager/lib/src/cookie_mgr.dart
```diff
@@ -28,7 +28,25 @@ class CookieManager extends Interceptor {
...
+      } else {
+        // Add this cookie to the list.
+        return [...prev, current];
+      }
+    });
+
```
AlexV525, line 44: nit: We don't need so many blank lines for inline codes, only for blocks like `test(...)`. Also consider to apply the suggestion to other codes.

plugins/cookie_manager/test/cookies_test.dart
```diff
@@ -139,6 +139,41 @@ void main() {
...
+    ];
+
+    final newCookies = CookieManager.getCookies(
+      [...cookies, ...previousCookies],
+    );
+
+    expect(newCookies, 'foo=bar; a=c; d=e; e=f');
+  });
+
+  test('testing cookies RFC6265 5.4 #2 sorting', () async {
```
AlexV525, line 162: A test named "test" is inefficient.

```suggestion
  test('RFC6265 5.4 #2 sorting', () async {
```

plugins/cookie_manager/lib/src/cookie_mgr.dart
jasonlaw: Sorry I am a bit confuse here, is it the **.fold** that causing the confusion?
If that is the case, we may use a Set variable to avoid the cookie with duplicate name.
~~~
 final uniqueNames = <String>{};
 // Filter out duplicate cookie name.
 cookies = cookies.where((cookie) => uniqueNames.add(cookie.name)).toList();
~~~
Is this has better readability?

plugins/cookie_manager/lib/src/cookie_mgr.dart
jasonlaw: noted.

plugins/cookie_manager/lib/src/cookie_mgr.dart
AlexV525: https://github.com/cfug/dio/blob/6b1f9edcfb575a6f4fdfc588cd5170e62bb04439/plugins/cookie_manager/lib/src/cookie_mgr.dart#L48-L54

Here we placed new cookies at the front, can we consider placing previous cookies first?1
```dart
 final newCookies = getCookies([
   ...?previousCookies
       ?.split(';')
       .where((e) => e.isNotEmpty)
       .map((c) => Cookie.fromSetCookieValue(c)),
   ...cookies,
 ]);
```

jasonlaw: This was working before we added the sorting logic, I have no idea why since the order should be remained the same with all are having the same path. Let me further dig out why it is not working and update here later.

AlexV525: I brought this up because this is the change that happened in your first commit.

jasonlaw: Indeed that was my first commit, and I found why it is not working with the sorting logic.
The previous cookies are constructed with path = null, and the sorting logic will place it to the last in order. 
To fix:
1. Fix the null path issue, but the previous cookies are retrieved from the header which has no path info.
2. Sort the null path in front.
3. Remove duplicate name.

What do you think?

AlexV525: Leaving the path as not nullable can be good in this case but good in compatibilities. The thing is `getCookies` is a public API, which can be used with `[...previous, ...new]` by users which makes it valid.

jasonlaw: Meanwhile, I will fix with option #2, which sorts the null path in front. Is it ok for you?

AlexV525: > Meanwhile, I will fix with option #2, which sorts the null path in front. Is it ok for you?

Yeah sure. Thanks!

AlexV525: > Leaving the path as not nullable can be good in this case but good in compatibilities.

Sorry about the typos, here I mean we should correct the `getCookies` rather than change fields like paths.

AlexV525:
CHANGELOG needs to be added, and we can have the final review.

plugins/cookie_manager/test/cookies_test.dart
```diff
@@ -139,6 +136,40 @@ void main() {
...
+      Cookie('d', 'e')..path = '/',
+      Cookie('e', 'f')..path = '/',
+    ];
+    final newCookies = CookieManager.getCookies(
+      [
+        ...previousCookies,
+        ...cookies,
+      ],
+    );
+    expect(newCookies, 'foo=oldbar; d=e; e=f; foo=bar; a=c');
```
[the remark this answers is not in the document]
jasonlaw: Yes you are right, it suppose to mimic the scenario with both previous and new cookies exists together.
I have updated the test code again so that the previous cookies will have null path, in order to match the actual scenario.

plugins/cookie_manager/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 - Allow `Set-Cookie` to be parsed in redirect responses.
 
+## 2.1.3
+
+- Fix new cookie being replaced by old cookie with the same name.
+- Sort the cookie by path (longer path first).
+
```
AlexV525, line 11:
```suggestion

- Allow `Set-Cookie` to be parsed in redirect responses.
- Fix new cookies being replaced by old cookies with the same name.
- Sort the cookie by path (longer path first).

```

AlexV525 (approved):

plugins/cookie_manager/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: dio_cookie_manager
 description: A cookie manager for Dio, which supports persistent cookies in RAM and file.
-version: 2.1.2
+version: 2.1.3
```
AlexV525, line 3:
```suggestion
version: 2.1.2
```

PR #1754, opened 2023-03-22
🚀 Allow defining adapter methods through their constructors

AlexV525:

Methods in `BrowserHttpClientAdapter` and `IOHttpClientAdapter` can be defined directly through their constructor instead of updates after the creation.

### DIFF
--- a/example/lib/proxy.dart
+++ b/example/lib/proxy.dart
@@ -8,18 +8,19 @@ void main() async {
   dio.options
     ..headers['user-agent'] = 'xxx'
     ..contentType = 'text';
-  // dio.options.connectTimeout = 2000;
-  // More about HttpClient proxy topic please refer to Dart SDK doc.
-  dio.httpClientAdapter = IOHttpClientAdapter()
-    ..onHttpClientCreate = (HttpClient client) {
+  dio.httpClientAdapter = IOHttpClientAdapter(
+    onHttpClientCreate: (HttpClient client) {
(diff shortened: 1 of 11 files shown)


ueman:
Isn't this technically a breaking change? It would break everyone who copied the example from the readme.

AlexV525:
> Isn't this technically a breaking change? It would break everyone who copied the example from the readme. Apart from that, I like this change.

Not really because you can still write like the previous code, it basically just provides another coding style.

ueman (approved):

example/lib/proxy.dart
```diff
@@ -8,18 +8,19 @@ void main() async {
...
-  // More about HttpClient proxy topic please refer to Dart SDK doc.
-  dio.httpClientAdapter = IOHttpClientAdapter()
-    ..onHttpClientCreate = (HttpClient client) {
+  dio.httpClientAdapter = IOHttpClientAdapter(
+    onHttpClientCreate: (HttpClient client) {
       client.findProxy = (uri) {
-        //proxy all request to localhost:8888
+        // Proxy all request to localhost:8888.
+        // Be aware, the proxy should went through you running device,
+        // not the host platform.
```
AlexV525, line 16: I added this because there are multiple times that some users ask: "Why I cannot access my proxy server at 127.0.0.1?". This might help them quickly realize they are on different devices.

PR #1755, opened 2023-03-22
[dio] Test typed responses

kuhnroyal:
This PR relates to https://github.com/cfug/dio/discussions/1335 and is for now just meant as idea and for further testing.
But correctly typed responses do work.


kuhnroyal:
@AlexV525 @ueman I had this open for a while and would like some feedback for 6.0.0 :)

kuhnroyal:
Really would love some feedback here :)

ueman (approved):
I like it, good work

AlexV525 (commented):
This is absolutely a significant step forward for the package. So everyone should write `request<String?>` instead of `request<String>`, right? If they upgraded without thinking, this might produce much NPE during the runtime?

AlexV525:
The change is great, but we might evaluate the cost for the whole ecosystem.

kuhnroyal:
I agree that this can break. So we need good examples in the 6.0.0 migration docs.

But users who have handling for nullable responses via `response?.data` or `if (response.data != null)` will receive lints at these points and need to figure out what they want or expect. Good samples are the key here.

kuhnroyal:
I added a changelog and some basic samples to the migration docs.
Really think we should do this. We have to do it at some point, might as well get it over with now :D

AlexV525 (approved):
🚀

PR #1777, opened 2023-03-31
[dio] upgrade plugins version

p709723778:

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ## 5.1.2
 
-- Update plugins version
+- Bump cookie_jar from 3.0.0 to 4.0.1
 
 ## 5.1.1
 
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 *None.*
 
+## 5.1.2
+
+- Update plugins version
+
 ## 5.1.1
 
 - Revert changes to `CancelToken.cancel()` behavior, as a result the `DioError`
--- a/dio/pubspec.yaml
+++ b/dio/pubspec.yaml
@@ -15,7 +15,7 @@ environment:
   sdk: '>=2.15.0 <3.0.0'
 
 dependencies:
-  http_parser: ^4.0.2
+  http_parser: ^4.0.0
   meta: ^1.5.0
   path: ^1.8.0
 
--- a/dio/pubspec.yaml
+++ b/dio/pubspec.yaml
@@ -9,20 +9,20 @@ description: |
 homepage: https://github.com/cfug/dio
 repository: https://github.com/cfug/dio/blob/main/dio
 issue_tracker: https://github.com/cfug/dio/issues
-version: 5.1.1
+version: 5.1.2
 
 environment:
   sdk: '>=2.15.0 <3.0.0'
 
 dependencies:
-  http_parser: ^4.0.0
+  http_parser: ^4.0.2
   meta: ^1.5.0
   path: ^1.8.0
 
 dev_dependencies:
   lints: any
-  test: ^1.5.1
-  coverage: ^1.0.3
+  test: ^1.24.0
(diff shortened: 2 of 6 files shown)


plugins/cookie_manager/pubspec.yaml
```diff
@@ -9,9 +9,9 @@ environment:
   sdk: '>=2.15.0 <3.0.0'
 
 dependencies:
-  cookie_jar: ^3.0.0
-  dio: ^5.0.0
+  cookie_jar: ^4.0.1
+  dio: ^5.1.1
```
ueman, line 13: For major version updates, it makes sense.

Please add a changelog entry which mentions this and bump the version number.
.

plugins/http2_adapter/pubspec.yaml
```diff
@@ -9,10 +9,10 @@ environment:
   sdk: '>=2.15.0 <3.0.0'
 
 dependencies:
-  http2: ^2.0.0
-  dio: ^5.0.0
+  http2: ^2.0.1
+  dio: ^5.1.1
```
ueman, line 13: These changes are unnecessary. Please revert them and the version bump as well.

dio/pubspec.yaml
```diff
@@ -9,20 +9,20 @@ description: |
...
 issue_tracker: https://github.com/cfug/dio/issues
-version: 5.1.1
+version: 5.1.2
 
 environment:
   sdk: '>=2.15.0 <3.0.0'
 
 dependencies:
-  http_parser: ^4.0.0
+  http_parser: ^4.0.2
```
ueman, line 18: There's no need to unnecessarily constrain this to the newest version. It would force downstream user to use the newest version even if we're still compatible with previous versions.

In general it's a good idea to constrain our dependencies as loosely as possible so that we don't cause dependency conflicts with other libraries or the consumer of this library.

If a user wants to use the latest http parser version, he can already do so. We don't need to update that on our side.

Constraining dev dependencies to the latest version makes sense, since we can make use of more lints etc and because they can't create dependency conflicts.

So please revert this change and the version number bump here.

dio/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
 *None.*
 
+## 5.1.2
+
+- Bump cookie_jar from 3.0.0 to 4.0.1
```
ueman, line 9: This should be in the cookie manager changelog.

dio/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ description: |
 homepage: https://github.com/cfug/dio
 repository: https://github.com/cfug/dio/blob/main/dio
 issue_tracker: https://github.com/cfug/dio/issues
-version: 5.1.1
```
ueman, line 12: This change are not needed

plugins/cookie_manager/pubspec.yaml
ueman: This is still not done

ueman (commented):
Thanks for your contribution 👍

dio/pubspec.yaml
kuhnroyal: AFAIK dev dependencies do create conflicts, just thinking of Flutter and `flutter_test`. I think we should revert all the changes to dev dependencies or align them with the minimum supported Flutter version.

ueman: but we don't depend on Flutter test for the changed projects?

kuhnroyal: But if I want to use Dio in a Flutter project, this might raise the test package higher than supported by Flutter.

ueman: Why would it? dev_dependencies aren't resolved by consumers of a particular library. See https://dart.dev/tools/pub/dependencies#dev-dependencies

kuhnroyal: Hmm yeah, seems to bi that way. I remember having conflicts somehow. Nvm.

PR #1780, opened 2023-04-06
improve log interceptor

LucasMatheusDev:

### Motivation 
When running the app on systems focused on application analysis, such as [BrowserStack](https://www.browserstack.com) , same in profile mode , the request logs can leak unwanted information, even if the developer has the option to remove the print, most of them are not aware of these "leaks" of requests.

This opens the way for an attacker who has an apk , aab or any other file that the browserStack
accept, be able to run the app on the website and view the logs

So I suggest the inclusion of this little code that by default only executes the prints in debug mode

### DIFF
--- a/dio/lib/src/interceptors/log.dart
+++ b/dio/lib/src/interceptors/log.dart
@@ -15,7 +15,7 @@ class LogInterceptor extends Interceptor {
     this.responseHeader = true,
     this.responseBody = false,
     this.error = true,
-    this.logPrint = printOnlyDebug,
+    this.logPrint = debugPrint,
   });
 
   /// Print request [Options]
@@ -133,7 +133,7 @@ class LogInterceptor extends Interceptor {
   }
 }
 
-void printOnlyDebug(Object? object) {
+void debugPrint(Object? object) {
   assert(() {
     print(object);
     return true;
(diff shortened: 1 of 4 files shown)


AlexV525 (dismissed):
This is a breaking change that previous users might also suffer from the change that logs are not printed by default, and I'm pretty sure that we can't make changes like this at the moment.
I'd suggest warning users when constructing the interceptor.

LucasMatheusDev:
> This is a breaking change that previous users might also suffer from the change that logs are not printed by default, and I'm pretty sure that we can't make changes like this at the moment. I'd suggest warning users when constructing the interceptor.

Hello @AlexV525 , this change only prevents prints in Production mode, and prints in Debug mode remain by default.

Between avoiding and warning about a possible leak of sensitive data, I believe it's better to avoid.


I believe that no developer expects their app to show logs in Production mode, so this suggestion does not disrupt the workflow of those already using the package.

What do you think about it?

dio/lib/src/interceptors/log.dart
```diff
@@ -132,3 +132,10 @@ class LogInterceptor extends Interceptor {
     msg.toString().split('\n').forEach(logPrint);
   }
 }
+
+void printOnlyDebug(Object? object) {
```
ueman, line 136:
```suggestion
void debugPrint(Object? object) {
```

ueman:
I agree with @LucasMatheusDev. Though, we should enable people to use print even in production. We should also point out this change prominently in the changelog and readme.

Can you please add it to the changelog and the readme?

AlexV525:
cc @ueman to review this again.

ueman (approved):
LGTM

PR #1704, opened 2023-03-01
Fix encoding of query parameters

kuhnroyal:
Changes and test for correct encoding of collections in query parameters.

I am totally not sure if the encoding for `x-www-form-urlencoded` is supposed to be different from query parameters.

### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [x] I have added the required tests to prove the fix/feature I'm adding
- [ ] I have updated the documentation (if necessary)
- [x] I have run the tests without failures
- [x] I have updated the `CHANGELOG.md` in the corresponding package

PR #1794, opened 2023-04-21
🚀 Backward compatible with `Map<dynamic, dynamic>` encodes

AlexV525:
Partially reverts #1704.

When constructing the data like `{key: value}`, it can be easy that the type is inferred to a dynamic map, rather than `Map<String, dynamic>`. So #1704 breaks part of previous usages and those usages become invalid after upgraded.


kuhnroyal:
I don't think we should allow this. I surely didn't think of that cases when I fixed the encoding problems :)
Other than that, it looks ok.

AlexV525:
> I don't think we should allow this.

Any intentions or reasons about this?

kuhnroyal:
> > I don't think we should allow this.
> 
> Any intentions or reasons about this?

It will just lead to problems and hard to trace issues with people having weirdly encoded keys.

PR #1803, opened 2023-05-04
Deprecate DioError in favor of DioException

### LINKED ISSUE #1789: DioError is named incorrectly
### Request Statement

This has bothered me since a while. 
By dart convention, `DioError` is named incorrectly, as it is not an `Error`, but an `Exception`.

The distinction is important, as `Exception` is meant to be caught but `Error` is not.
`Error` signals a mistake from the programmer, that is usually not recoverable, 
while `Exception` is simply a failure of an operation, that can be handled by the program.

As per:
https://api.flutter.dev/flutter/dart-core/Error-class.html
https://api.flutter.dev/flutter/dart-core/Exception-class.html

### Solution Brainstorm

I understand that renaming `DioError` to `DioException` would be a very large breaking change.
This change could be introduced gradually by adding a typedef:

```dart
@Deprecated('Use DioException instead')
typedef DioError = DioException;
```

It's also noteworthy that while `DioException` is more correct based on convention, it is a bit longer to write than `DioError`, though this will probably be eliviated by IDE autocompletion.

It would perhaps be useful to provide a codemod for this transition, as there probably will be many projects that use `DioError` in many places.

clragon:

Fixes #1789.

#### Quick summary

`DioError` is named incorrectly, as it is of type `Exception` and not `Error`.
The fix for this is to rename `DioError` to `DioException` and provide a typdef alias for `DioError` that is marked deprecated.
The typedef offers a grace period for migration and makes this a non-breaking change that can be introduced as soon as possible to notify users of the deprecation.

This PR makes these changes. Its worth considering also offering extra help with migration to users, for example by providing a lint with an automatic fix that turns `DioError` into `DioException` and `DioErrorType` into `DioExceptionType`.

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Make `LogInterceptor` prints in DEBUG mode (when the assertion is enabled) by default.
+- Deprecate `DioError` in favor of `DioException`.
 
 ## 5.1.2
 
@@ -82,7 +83,6 @@
 - refactor timeout logic
 - use 'arraybuffer' instead of 'blob' for xhr requests in web platform
 
-
 ## 4.0.4
 
 - Fix fetching null data in a response
@@ -154,7 +154,6 @@ the subsequent interceptors processing logic more finely (whether to skip them o
 2. fix #851
 3. fix #641
 
-
 ## 3.0.9 2020.2.24
 
 - Add test cases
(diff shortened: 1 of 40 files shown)


dio/lib/src/dio_error.dart
```diff
@@ -1,200 +1,10 @@
...
-  String toString() {
-    String msg = 'DioError [${type.toPrettyDescription()}]: $message';
-    if (error != null) {
-      msg += '\nError: $error';
-    }
-    return msg;
-  }
-}
+/// Deprecated in favor of [DioException] and will be removed in future major versions.
+@Deprecated('Use DioException instead')
```
AlexV525, line 9:
```suggestion
@Deprecated('Use DioException instead. This will be removed in 6.0.0')
```

dio/lib/src/dio_error.dart
```diff
@@ -1,200 +1,10 @@
...
-        return 'request cancelled';
-      case DioErrorType.connectionError:
-        return 'connection error';
-      case DioErrorType.unknown:
-        return 'unknown';
-    }
-  }
-}
+/// Deprecated in favor of [DioExceptionType] and will be removed in future major versions.
+@Deprecated('Use DioExceptionType instead')
```
AlexV525, line 4:
```suggestion
@Deprecated('Use DioExceptionType instead. This will be removed in 6.0.0')
```

dio/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ## 5.1.2
 
 - Allow `FormData` to send a null entry value as an empty string.
+- Deprecate `DioError` in favor of `DioException`.
```
AlexV525, line 10: The entry is placed in the wrong place. It should be under the unreleased section.

clragon:
The test fail right now because I intentionally did not update the plugins with this change yet.
Should I do that anyways?

kuhnroyal:
This would go into 5.2.0? And all the plugins need to also bump constraints and a new minor version?

clragon:
Right, this change would go into 5.2.0, as its non breaking and it will tell users as soon as possible that `DioError` is deprecated.
I have updated the PR to include all plugins and their changelogs.

clragon:
I assume that would happen with the release of the versions itself, and not with this PR.
But I can certainly include that.

AlexV525:
> But then you also need to raise the constraint in the plugins to `dio: ^5.2.0`.

That should be unnecessary IMO. No codes have been deleted actually.

AlexV525:
> But the plugins now use the new `DioException` which is being added in `5.2.0`.

Oh indeed. They should have new constraint then.

AlexV525 (changes requested):

AlexV525 (approved):

Issue #1811, opened 2023-05-11
Rework CancelToken handling and HttpClient creation for IOHttpClientAdapter


kuhnroyal:
### Request Statement

1. We should change the return type of `typedef OnHttpClientCreate = HttpClient? Function(HttpClient client);` not non-nullable `HttpClient`. Keeping it nullable makes no sense is probably a leftover of the NNBD migration.
2. Currently we create a new `HttpClient` for every request where a `CancelToken` is provided. In Dart 2.10 an API to abort requests was added, we should use `HttpClientRequest.abort` when the `CancelToken` is cancelled

PR #1810, opened 2023-05-11
Fix IOHttpClientAdapter.onHttpClientCreate Repeated calls

Passer-by:
Fix https://github.com/cfug/dio/issues/1809
### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [x] I have added the required tests to prove the fix/feature I'm adding
- [x] I have updated the documentation (if necessary)
- [x] I have run the tests without failures
- [x] I have updated the `CHANGELOG.md` in the corresponding package

PR #1812, opened 2023-05-11
Fix return type of OnHttpClientCreate

kuhnroyal:
The return type of `OnHttpClientCreate` has been changed from `HttpClient?` to `HttpClient`.
While technically a breaking change, it is considered a bug because the callback was actually
a no-op when the provided client was modified but not returned.

Fixes 1. of #1811

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -4,8 +4,7 @@
 
 - Make `LogInterceptor` prints in DEBUG mode (when the assertion is enabled) by default.
 - Fix `IOHttpClientAdapter.onHttpClientCreate` Repeated calls
-- `IOHttpClientAdapter.onHttpClientCreate` has been deprecated because of unclear handling
-  when the provided client was modified but not returned. This is scheduled for removal in
+- `IOHttpClientAdapter.onHttpClientCreate` has been deprecated and is scheduled for removal in
   Dio 6.0.0 - Please use the replacement `IOHttpClientAdapter.createHttpClient` instead.
 
 ## 5.1.2
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 - Make `LogInterceptor` prints in DEBUG mode (when the assertion is enabled) by default.
 - Fix `IOHttpClientAdapter.onHttpClientCreate` Repeated calls
+- `IOHttpClientAdapter.onHttpClientCreate` has been deprecated because of unclear handling
+  when the provided client was modified but not returned. This is scheduled for removal in
+  Dio 6.0.0 - Please use the replacement `IOHttpClientAdapter.createHttpClient` instead.
 
 ## 5.1.2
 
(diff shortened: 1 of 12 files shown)


dio/lib/src/adapters/io_adapter.dart
```diff
@@ -197,7 +200,7 @@ class IOHttpClientAdapter implements HttpClientAdapter {
     if (_defaultHttpClient == null) {
       client.idleTimeout = Duration(seconds: 3);
       if (onHttpClientCreate?.call(client) != null) {
```
ueman, line 202:
```suggestion
      if (onHttpClientCreate != null) {
```

kuhnroyal: Yea I want to rebase this after https://github.com/cfug/dio/pull/1810

AlexV525:
I just came out with an idea to avoid too breaking: Deprecate the `onHttpClientCreate` and add `createHttpClient` which is similar to `HttpOverrides`.

AlexV525:
We need to consider how many people are using the package...

kuhnroyal:
Well then we have only deprecated the bug :) But I am fine with this.

kuhnroyal:
This now also solves the same problem that #1810 does. Missing a test.

AlexV525:
I just thought about this last night, we can't really guarantee the fix if we keep the signature. I'm agree with changing the return type now. Sorry for the previous advice.

kuhnroyal:
Hmm now I am unsure :D
The deprecation is fine I think and we should start a 6.0.0 branch in parallel where we remove all the deprecated things.

dio/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@
 
 - Make `LogInterceptor` prints in DEBUG mode (when the assertion is enabled) by default.
 - Fix `IOHttpClientAdapter.onHttpClientCreate` Repeated calls
+- `IOHttpClientAdapter.onHttpClientCreate` has been deprecated because of unclear handling
```
AlexV525, line 7: Can we keep the CHANGLOG with critical info only and move details into the migration guide?

dio/lib/src/utils.dart
```diff
@@ -5,7 +5,7 @@ import 'dart:convert';
 import 'options.dart';
 import 'parameter.dart';
 
-/// Pipes all data and errors from [stream] into [sink]. Completes [Future] once
+/// Pipes all data and errors from [stream] into [sink]. Completes [Future]HttpClonce
```
AlexV525, line 8: Seems to be an accidently paste.

dio/CHANGELOG.md
AlexV525, line 1: Will the HTTP/2 adapter affected by this change?

dio/CHANGELOG.md
kuhnroyal: We only have breaking information there

dio/lib/src/utils.dart
kuhnroyal: oops

dio/CHANGELOG.md
kuhnroyal: Nah I don't see it being used there.

AlexV525 (approved):

PR #1818, opened 2023-05-15
Abort HttpRequests when CancelToken is cancelled

### LINKED ISSUE #1811: Rework CancelToken handling and HttpClient creation for IOHttpClientAdapter

kuhnroyal:
No longer close the `HttpClient` but instead abort individual requests. This has the added benefit of fewer `HttpClient`s being constructed. If `CancelToken.cancel` is called before the request has started, cancel the whole operation.
The `HttpClientRequest.abort` method was added in Dart 2.10.

Fixes #1811.


dio/test/cancel_token_test.dart
```diff
@@ -19,5 +19,27 @@ void main() {
     test('cancel without use does not throw (#1765)', () async {
       CancelToken().cancel();
     });
+
+    test('cancels multiple requests', () async {
```
AlexV525, line 23: These tests didn't seem covered clients closed status?

kuhnroyal: No they didn't. And I am currently also not testing if `abort()` is called - probably need a mock `HttpClient` and mocked requests for this.

AlexV525 (approved):

dio/test/cancel_token_test.dart
kuhnroyal: I updated the tests to use mocks and verify that `abort` is actually being called.

AlexV525 (approved):

PR #1826, opened 2023-05-16
Remove custom idleTimeout in IOHttpClientAdapter

### LINKED ISSUE #1801: Make `idleTimeout` in `IOHttpClientAdapter` configurable
You are probably not interested anymore but incase someone else is searching for it I will post my findings here.

It looks like there is a `idleTimeout` (see https://github.com/cfug/dio/blob/main/dio/lib/src/adapters/io_adapter.dart#L198).
Default is 3 seconds.
So if you don't do anything for 3 seconds the connection will get closed and it will use a new connection next time.
Sadly I believe that this also seems to force a reconnection if your rest-service itself needs more than 3 seconds.
(atleast thats my observation with flutter dev tools)
Although this won't disrupt an active connection, it can still be annoying.

I'm not shure, if there is a "normal" way to configure it.
But for servers that understand http 2.0 you could probably use the example there:
[dio_http2_adapter](https://github.com/cfug/dio/blob/main/plugins/http2_adapter/README.md)
to increase the timeout.

_Originally posted by @TheCrazyT in https://github.com/cfug/dio/issues/912#issuecomment-1533211269_

kuhnroyal:
* fixes #1801 - I think we should not provide a customisation option directly, all customisation can be done via `createHttpClient` instead. I also don't understand why we did default to 3 seconds.

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -7,6 +7,8 @@
 - Remove `IOHttpClientAdapter.onHttpClientCreate` which was deprecated in `5.2.0`
 - Remove `DioError` and `DioErrorType` which was deprecated in `5.2.0`.
 - Remove `DefaultTransformer` which was deprecated in `5.0.0`.
+- `IOHttpClientAdapter` no longer sets a custom `HttpClient.idleTimeout`. A custom `HttpClient` can be provided via
+  `IOHttpClientAdapter.createHttpClient` if customisation is required.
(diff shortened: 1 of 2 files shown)


ueman:
I think this should go into the next major version. A changed default timeout seems like a significant behavior change.

Other than that, this seems like a worthwhile change.

ueman (approved):

PR #1836, opened 2023-05-27
Added handling for xhr.onTimeout

morhaviv:
I've been trying to set connectionTimeout and receiveTimeout for my http requests on chrome, and it didn't seem to be working. After further investigation, it seems like the request was actually being canceled after the timeout passed, but after the cancellation nothing happened (the request was still blocked and the completer wasn't called). 

So I found out xhr has a onTimeout method, I used it and raised the required exception and now it works just fine.


kuhnroyal:
Thanks, looks good. I can add a test in a separate branch.

dio/CHANGELOG.md
```diff
@@ -7,7 +7,8 @@
 - Fix `IOHttpClientAdapter.onHttpClientCreate` Repeated calls
 - `IOHttpClientAdapter.onHttpClientCreate` has been deprecated and is scheduled for removal in
   Dio 6.0.0 - Please use the replacement `IOHttpClientAdapter.createHttpClient` instead.
-- Using `CancelToken` no longer closes and re-creates `HttpClient` for each request when `IOHttpClientAdapter` is used. 
+- Using `CancelToken` no longer closes and re-creates `HttpClient` for each request when `IOHttpClientAdapter` is used.
+- Fix timeout handling for browsers on `connectTimeout` and `receiveTimeout`.
```
kuhnroyal, line 11:
```suggestion
- Fix timeout handling for browser `receiveTimeout`.
```
This PR actually should not have any effect on the connectTimeout which is handled manually.

morhaviv:
> Thanks, looks good. I can add a test in a separate branch.

Sounds great!
I accepted your CHANGELOG.md suggestion :)

kuhnroyal (approved):

AlexV525 (approved):

dio/lib/src/adapters/browser_adapter.dart
```diff
@@ -178,6 +178,24 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
...
+    xhr.onTimeout.first.then((_) {
+      int timeoutMilliseconds = 0;
+      if (connectTimeout != null) {
+        timeoutMilliseconds += connectTimeout.inMilliseconds;
+      }
+      if (receiveTimeout != null) {
+        timeoutMilliseconds += receiveTimeout.inMilliseconds;
+      }
+      connectTimeoutTimer?.cancel();
+      completer.completeError(
```
AlexV525, line 190: One thing I've missed is we should check if the completer is completed before calling this.

morhaviv:
@kuhnroyal @AlexV525 So I've just looked into this piece of code a little bit more, and I noticed something odd. If `connectTimeout` *OR* `receiveTimeout` is set to zero, it seems like `xhr.timeout` is not set at all. Which means if you don't set both `connectTimeout` and `receiveTimeout` the timeout won't be set at all. 

For me it seems like a mistake, correct me if I'm wrong, so I improved the logic there a little bit and added a variable for `xhrTimeout` instead of calculating the timeout twice.

dio/lib/src/adapters/browser_adapter.dart
morhaviv: @kuhnroyal Added :) Thanks!

AlexV525: This seems still not addressed.

morhaviv: @AlexV525 I think the preview here still displays the old code, go to the "Files changed" section, you'll see on the latest code version:
```
      if (connectTimeoutTimer != null) {
        connectTimeoutTimer?.cancel();
      }
```

morhaviv: @AlexV525 Sorry my bad, misunderstood what you meant, now I think I handled it properly :)

kuhnroyal:
> > For me it seems like a mistake, correct me if I'm wrong, so I improved the logic there a little bit and added a variable for `xhrTimeout` instead of calculating the timeout twice.
> 
> Great catch. I also think it might be a mistake too.

As crazy as it sounds at first, but this is NOT a bug!
`xhr` only has one timeout setting which is neither directly a connect nor a receive timeout, so this seem to be a hacky way to support these. If one of those timeouts is set to `Duration.zero` (which basically means no timeout) then we must not set the other timeout because it would be set as the total `xhr.timeout` which is not the intention.

morhaviv:
> > > For me it seems like a mistake, correct me if I'm wrong, so I improved the logic there a little bit and added a variable for `xhrTimeout` instead of calculating the timeout twice.
> > 
> > 
> > Great catch. I also think it might be a mistake too.
> 
> As crazy as it sounds at first, but this is NOT a bug! `xhr` only has one timeout setting which is neither directly a connect nor a receive timeout, so this seem to be a hacky way to support these. If one of those timeouts is set to `Duration.zero` (which basically means no timeout) then we must not set the other timeout because it would be set as the total `xhr.timeout` which is not the intention.


Ok, I got why it's intended, but I must admit it took me some time to understand and I'm not sure that every developer will understand what's going on at first, even though there is sense with not limiting the whole connection to the same amount of time as the `connectTimeout` and etc.

Perhaps in this specific instance, adding a different setting such as `browserTimeout` is the right thing to do? 

Anyways, updated the code to keep the previous logic :)

kuhnroyal:
Yea we need to think about that separately.

AlexV525 (approved):

AlexV525:
Tests seem to be failed though.

kuhnroyal:
Yea this is too flaky based on network conditions and we can not run a custom server for the browser test, except if we would start some Docker service container in CI.

PR #1861, opened 2023-06-11
Fix changelog for 5.2.1

kuhnroyal:
Just fixes the changelog version


ueman (approved):

kuhnroyal:
Ok so it looks like the release action doesn't work like I expected it to. Making another PR.

PR #1873, opened 2023-06-16
[dio] Ensure consistent handling of "/" in concatenation of baseUrl with path

KernelPanic92:
This PR ensures consistent handling of "/" when concatenating the baseUrl with the path.
Previously, if the baseUrl had a trailing "/" or the path had a leading "/", the resulting URL would contain duplicate slashes.
To address this issue, the algorithm has been modified to insert only a single "/" regardless of whether the baseUrl has a trailing slash or the path has a leading slash. This improvement enhance the reliability and correctness of the concatenation process, providing a more robust solution.

### DIFF
--- a/dio/lib/src/options.dart
+++ b/dio/lib/src/options.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:meta/meta.dart';
 
 import 'adapter.dart';
@@ -585,20 +587,21 @@ class RequestOptions extends _RequestConfig with OptionsMixin {
 
   /// generate uri
   Uri get uri {
-    String url = path;
-    if (!url.startsWith(RegExp(r'https?:'))) {
-      url = '$baseUrl/$url';
-      final s = url.split(':/');
-      if (s.length == 2) {
-        url = '${s[0]}:/${s[1].replaceAll(RegExp(r'///?'), '/')}';
-      }
+    Uri uri = Uri.parse(path);
+
+    if (!uri.isAbsolute) {
+      uri = Uri.parse(baseUrl).resolveUri(uri);
     }
+
     final query = Transformer.urlEncodeQueryMap(queryParameters, listFormat);
     if (query.isNotEmpty) {
-      url += (url.contains('?') ? '&' : '?') + query;
+      final separator = uri.query.isNotEmpty ? '&' : '';
+      final mergedQuery = uri.query + separator + query;
+      uri = uri.replace(query: mergedQuery);
     }
+
     // Normalize the url.
-    return Uri.parse(url).normalizePath();
+    return uri.normalizePath();
   }
 
   /// Request data, can be any type.
--- a/dio/lib/src/options.dart
+++ b/dio/lib/src/options.dart
@@ -587,10 +587,10 @@ class RequestOptions extends _RequestConfig with OptionsMixin {
   Uri get uri {
     String url = path;
     if (!url.startsWith(RegExp(r'https?:'))) {
-      url = baseUrl + url;
+      url = '$baseUrl/$url';
       final s = url.split(':/');
       if (s.length == 2) {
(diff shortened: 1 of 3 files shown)


dio/lib/src/options.dart
```diff
@@ -587,10 +587,10 @@ class RequestOptions extends _RequestConfig with OptionsMixin {
   Uri get uri {
     String url = path;
     if (!url.startsWith(RegExp(r'https?:'))) {
-      url = baseUrl + url;
+      url = '$baseUrl/$url';
```
ueman, line 590: Can you try to build the URL with the Uri class? I assume that should probably result in a safer implementation. It's constructor takes a list of path segments, so I guess it should also be easy to use.

KernelPanic92: @ueman Sure! Since this is my first contribution to this library, I didn't want to be too intrusive, but your idea is really great! Therefore, to make the most of the Uri object, I would completely refactor the algorithm in the following way:

```dart
Uri get uri {
    Uri uri = Uri.parse(path);

    if (!uri.isAbsolute) {
      uri = Uri.parse(baseUrl).resolveUri(uri);
    }

    final query = Transformer.urlEncodeQueryMap(queryParameters, listFormat);
    if (query.isNotEmpty) {
      final querySeparator = uri.query.isNotEmpty ? '&' : '';
      final replacedQuery = '${uri.query}$querySeparator$query';
      uri = uri.replace(query: replacedQuery);
    }

    // Normalize the url.
    return uri.normalizePath();
  }
```

(Take this algorithm with a grain of salt: I haven't tested it yet 😅)

ueman: That approach looks indeed much better. Thank you

KernelPanic92: After applying the refactoring, I noticed a step in the old algorithm that may require further consideration. The following code snippet exists in the previous approach:

```dart
    if (query.isNotEmpty) {
      url += (url.contains('?') ? '&' : '?') + query;
    }
```

This logic checks if the URL already contains parameters, but it may not make sense to check for parameter presence in the URL since it should always be a path, and parameter handling is delegated to the `queryParameters` field. In other words, is it expected to have parameters included in the `baseUrl` or the `path`, or is it simply a defensive approach?

Reintroducing this logic raises a couple of concerns:

1. What should we do if a parameter in the path is encoded differently from `listFormat`?
2. What should we do if a parameter in the path also exists in `queryParameters`? Should we attempt to merge the two values or replace the value in `queryParameters`?

It would be prudent to carefully evaluate these issues and make an informed decision before reintroducing this logic in the refactoring.

@ueman @cfug-dev

ueman: Are there existing tests for that part of the logic?

KernelPanic92: @ueman no, unfortunately not...

ueman: @kuhnroyal @AlexV525 What do you think?

AlexV525: I haven't read this completely, but we should be cautious when writing new implementations that abandon old behavior. I would suggest that you take steps to change (in many PRs maybe):
1. Resolves duplicate `/` only, without additional changes.
2. Before the next major release, extract how we handle `ListFormat` as configurable.
3. In the next major release, apply the same handling to the `baseUrl`, and convert the implementation to `Uri`.

ueman (approved):

ueman:
@KernelPanic92 If you want you can open a PR against the v6.0.0 branch with your improvements via the URI class. Thanks for the PR.

KernelPanic92:
@ueman Attention, I hadn't yet found the time to restore the initial version of the code, so in this branch, the management using the Uri object was still present. 😱
Are you sure it's acceptable? If it's fine and wasn't an oversight, we simply need to rebase the v6.0.0 branch 🚀

AlexV525:
It's been reverted, so you'll need to create a new one. @KernelPanic92

PR #1874, opened 2023-06-17
🐛 Only produce null response body when the response is a JSON type

### LINKED ISSUE #1834: Empty response body is read as null
### Package

dio

### Version

5.1.2

```shell
[✓] Flutter (Channel stable, 3.10.2, on macOS 13.3.1 22E772610a darwin-arm64, locale en-KR)
    • Flutter version 3.10.2 on channel stable at /Users/jeiea/.asdf/installs/flutter/2.10.3-stable
    • Framework revision 9cd3d0d9ff (28 hours ago), 2023-05-23 20:57:28 -0700
    • Engine revision 90fa3ae28f
    • Dart version 3.0.2
    • DevTools version 2.23.1

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    • Android SDK at /Users/jeiea/Library/Android/sdk
    • Platform android-33, build-tools 33.0.2
    • Java binary at: /Users/jeiea/.asdf/shims/java
    • Java version OpenJDK Runtime Environment Temurin-11.0.15+10 (build 11.0.15+10)

[✓] Xcode - develop for iOS and macOS (Xcode 14.3)
    • Build 14E222b
    • CocoaPods version 1.12.1

[!] Android Studio
    • Android Studio at /Applications/Android Studio Preview.app/Contents
    ✗ Unable to find bundled Java version.
    • Try updating or re-installing Android Studio.

[✓] VS Code (version 1.78.2)
    • Flutter extension version 3.64.0

[✓] Connected device (3 available)
    • HojuniPad (mobile) • 00008101-001D44A63E29001E • ios            • iOS 16.4.1 20E252
    • macOS (desktop)    • macos                     • darwin-arm64   • macOS 13.3.1 22E772610a darwin-arm64
    • Chrome (web)       • chrome                    • web-javascript • Google Chrome 113.0.5672.126

[✓] Network resources

! Doctor found issues in 1 category.
```

### Dart Version

_No response_

### Steps to Reproduce

Run the following dart source.

```dart

Future<void> main() async {
  final dio = Dio();
  final response = await dio.get('https://jeiea.free.beeceptor.com/empty');
  final data = response.data;
  print(data);
}
```

### Expected Result

Print empty string.

I found this from my app's breaking behavior so think v4 would return empty string in this situation.

Also I think regarding as empty string is correct because the response has text/html content type and 200 OK, not no content.

### Actual Result

```
> dart run src/main.dart
null
```

AlexV525:
Resolves #1834.


dio/lib/src/transformers/sync_transformer.dart
```diff
@@ -42,86 +42,96 @@ class SyncTransformer extends Transformer {
...
+        streamCompleter.completeError(error, stackTrace);
+      },
+      onDone: () {
+        streamCompleter.complete();
       },
-      onDone: () => completer.complete(),
       cancelOnError: true,
     );
     options.cancelToken?.whenCancel.then((_) {
       return subscription.cancel();
```
AlexV525, line 97: I wasn't sure about how this line will behave (whether we should complete the completer or not). It's a void future and we await for the completer in the outside scope.

dio/lib/src/dio_mixin.dart
```diff
@@ -559,7 +558,19 @@ abstract class DioMixin implements Dio {
       );
       final statusOk = reqOpt.validateStatus(responseBody.statusCode);
       if (statusOk || reqOpt.receiveDataWhenStatusError == true) {
-        ret.data = await transformer.transformResponse(reqOpt, responseBody);
+        Object? data = await transformer.transformResponse(
+          reqOpt,
+          responseBody,
+        );
+        // Make the response as null before returned as JSON.
+        if (data is String &&
```
bgenidy, line 566: wouldn't this not give users the ability to override the response transformer to keep existing logic of returning string instead of null?

AlexV525: `""` would not be parsed as a JSON object (mostly written in `Map`) successfully. So, overwrite the result to could avoid exceptions like `String is not a subtype of Map`.

This also can be improved by adding the generic type to `transformResponse`, then it should be in the next major release. I've put it on my list and will apply it later.

bgenidy: I had the opposite experience where generating my updated flutter api using openapi with dio 5.x started throwing the following exception because my no content api decorators were strings "**Error: type 'Null' is not a subtype of type 'String' in type cast**"

thanks, will wait for the next major release to test things out!

AlexV525: > I had the opposite experience where generating my updated flutter api using openapi with dio 5.x started throwing the following exception because my no content api decorators were strings "**Error: type 'Null' is not a subtype of type 'String' in type cast**"

So the PR should fix your issue. Could you also help us to test if it's working?

bgenidy: Yes can definitely test it out, but it wouldn't help in my case because I have a json rest API, I just used the incorrect data type in the swagger spec as sometimes void acts weird with some language generation sdks.

Here is the code I had in Dio 4.x that was working (backend is .net stack)

```csharp
[HttpPost("foo/")]
        [SwaggerResponse((int)System.Net.HttpStatusCode.OK, type: typeof(string))]
        public async Task FooAsync()
```

I've updated it to be the correct return type
```csharp
[HttpPost("foo/")]
        [SwaggerResponse((int)System.Net.HttpStatusCode.OK, type: typeof(void))]
        public async Task FooAsync()
```

AlexV525: Sorry I cannot read .NET code. Could you simply your use case as a minimal example? I would like to know if the PR is valid in cases I can cover, and maybe add more tests.

bgenidy: apologies let me clarify, my server response type is empty (no content) json type. However the way I had my flutter code written was as follows

```dart
Future<Response<String>> apiFooPost({ 
    required String version,
    CancelToken? cancelToken,
    Map<String, dynamic>? headers,
    Map<String, dynamic>? extra,
    ValidateStatus? validateStatus,
    ProgressCallback? onSendProgress,
  }) async {
    final _path = r'/api/v{version}/foo/{userId}'.replaceAll('{' r'version' '}', version.toString());
    final _options = Options(
      method: r'POST',
      headers: <String, dynamic>{
        ...?headers,
      },
      extra: <String, dynamic>{
        'secure': <Map<String, String>>[
          {
            'type': 'http',
            'scheme': 'bearer',
            'name': 'Bearer',
          },
        ],
        ...?extra,
      },
      validateStatus: validateStatus,
    );
    final _response = await _dio.request<Object>(
      _path,
      options: _options,
      cancelToken: cancelToken,
      onSendProgress: onSendProgress,
      onReceiveProgress: onReceiveProgress,
    );

    String _responseData;

    try {
      _responseData = _response.data as String;

    } catch (error, stackTrace) {
      throw DioException(
        requestOptions: _response.requestOptions,
        response: _response,
        type: DioExceptionType.unknown,
        error: error,
        stackTrace: stackTrace,
      );
    }
```

when I updated to Dio 5.x the above code started throwing a null cast exception. So I switched it to 

```dart
Future<Response<void>> apiFooPost({ 
    required String version,
    CancelToken? cancelToken,
    Map<String, dynamic>? headers,
    Map<String, dynamic>? extra,
    ValidateStatus? validateStatus,
    ProgressCallback? onSendProgress,
  }) async {
    final _path = r'/api/v{version}/foo/{userId}'.replaceAll('{' r'version' '}', version.toString());
    final _options = Options(
      method: r'POST',
      headers: <String, dynamic>{
        ...?headers,
      },
      extra: <String, dynamic>{
        'secure': <Map<String, String>>[
          {
            'type': 'http',
            'scheme': 'bearer',
            'name': 'Bearer',
          },
        ],
        ...?extra,
      },
      validateStatus: validateStatus,
    );
    return await _dio.request<Object>(
      _path,
      options: _options,
      cancelToken: cancelToken,
      onSendProgress: onSendProgress,
      onReceiveProgress: onReceiveProgress,
    );
    }
```

if I understand your change correctly I would still need to keep the above logic or override the default response decoder otherwise I would still get a null in Dio 5.x since my response is of type json. Let me know if I misunderstood this PR?

AlexV525: After this change, the response will only produce null (compared to the targeting flow) when:
1. The response is JSON and is not empty.
2. The request generic type is not `String` (`T != dynamic && T != String`).

Your previous code requests `Object`, so the response type would be applied as JSON, but `String` is a subtype of `Object`, so empty strings are allowed. And the PR should allow this again.

bgenidy: tested it out and still get null data

image

This is what I have for my pubspec incase you see any issues with my setup
```yaml
  dio:
    git:
      url: https://github.com/cfug/dio
      path: dio
      ref: fix/null-response-body-only-when-response-type-json
```

AlexV525: Is your content type `application/json`? If so, the response will be manipulate too.

bgenidy: yes request is "application/json" response is a 200 with no content (I don't return an empty string or anything just a 200 success)

AlexV525:
cc @kuhnroyal @ueman

kuhnroyal (approved):

PR #1889, opened 2023-07-06
✨ Make MultipartFile recoverable to enable retrying FormData requests

gabrielaraujoz:
- Add restoration capability to MultipartFile
- Add tests to ensure MultipartFile is not changed during cloning, it just gets a new instance that can be consumed by FormData

Possible next steps: Create method in FormData to make it recoverable/clonable as well.

Hello there!

We at Revelo use Dio for all of our http requests. We also use MultiPart a lot and have gotten a significant amount of errors when retrying to send a MultiPart File because the stream has already been listened to/finalized.

exception: Bad state: The FormData has already been finalized. This typically means you are using the same FormData in repeated requests.. Error thrown .

This is a common use case for ourselves (and we think for others as well) due to access token's short duration. As it expires, we automatically renew it with the refresh token, and finally do the original request again, that's why we fall into this problem.

I thought of a possible way to make retrying these requests possible and it involves, as a first step, making MultipartFile store the original source path/value and be able to "clone" itself with a new stream generated from the same source path. 

Even with this change, to retry the request we will still need to recreate FormData. But this small change already makes it possible to recreate it without the need of a secondary interface that would just be used to store the original source path. This proposed change avoids a lot of boilerplate code and extra state management, since we can "reset" the MultipartFile to it's initial state by simply calling `restoreMultipartFile()`.

With this, we could retrieve add a new restored `FormData` in the RequestOptions object, and would not need to have a secondary solution just to get the file paths for each MultipartFile and each request.

WDYT? Would this ideia be a good addition to Dio? I plan to make another PR with a FormData.clone() convenience method to avoid even more boilerplate code once I have a little more spare time, but this second PR would require us to first apply the proposed change in this one.

Thanks a lot, best regards!


dio/lib/src/multipart_file.dart
```diff
@@ -44,9 +50,13 @@ class MultipartFile {
     this.filename,
     MediaType? contentType,
     Map<String, List<String>>? headers,
+    String? filePath,
+    List<int>? value,
```
AlexV525, line 54: These fields seem unnecessary if the `restoreMultipartFile` is the method of `MultipartFile`, they might bring extra difficulties for users to understand why the class has such fields since we already have related. In order to avoid outside reference, you can use the constructor in `restoreMultipartFile` rather than `multipartFileFromPathSync`.

gabrielaraujoz: Amazing! I just made this change, WDYT?

dio/lib/src/multipart_file.dart
```diff
@@ -140,4 +140,14 @@ class MultipartFile {
     _isFinalized = true;
     return _stream;
   }
+
+  /// Restore MultipartFile, returning a new instance of the same object. This is useful if your request failed and you wish
+  /// to sistematically retry it, for example a 401 error that can be solved by refreshing the token.
+  MultipartFile restoreMultipartFile() => MultipartFile(
```
AlexV525, line 146: I'm not sure `restore` is a good word here because it represents manipulating the original object but here we produce a new one. Any other ideas? Other words like `copy` or `duplicate` might fit more in this case.

dio/lib/src/multipart_file.dart
```diff
@@ -140,4 +140,14 @@ class MultipartFile {
     _isFinalized = true;
     return _stream;
   }
+
+  /// Restore MultipartFile, returning a new instance of the same object. This is useful if your request failed and you wish
+  /// to sistematically retry it, for example a 401 error that can be solved by refreshing the token.
```
AlexV525, line 145: Please update codes to follow the 80-word length format or semantic line breaks.
```suggestion
  /// Restore MultipartFile, returning a new instance of the same object.
  /// This is useful if your request failed and you wish to systematically retry it,
  /// such as an unauthorized exception can be solved by refreshing the token.
```

dio/test/formdata_test.dart
AlexV525, line 1: The new method seems not tested though.

dio/lib/src/multipart_file.dart
gabrielaraujoz: humm, how about `reset`?

dio/lib/src/multipart_file.dart
gabrielaraujoz: I applied the line length rules to all the changes I made, thanks

dio/test/formdata_test.dart
gabrielaraujoz: I changed the tests to comply to this, could you please check it out? https://github.com/cfug/dio/pull/1889/commits/a22208d0d7ae0542ebb4bb0b47db06f0091cc5a8

dio/lib/src/multipart_file.dart
gabrielaraujoz: I renamed it as duplicateMultipartFile, but let me know if you think reset would fit better :)

gabrielaraujoz:
Hey @AlexV525, hope you're doing well. Could you please take a look and let me know if the changes are ok? We're currently dealing with this bug in production and would love to fix it with this PR.

Thanks a lot!

kuhnroyal:
Can we not just call the method `clone()`?

gabrielaraujoz:
@kuhnroyal awesome, done! @AlexV525 fyi

dio/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 ## Unreleased
 
 - Remove `http` from `dev_dependencies`.
+- Add support for restoring MultipartFile from `FormData`.
```
kuhnroyal, line 9:
```suggestion
- Add support for cloning `MultipartFile` from `FormData`.
```

kuhnroyal (approved):

gabrielaraujoz:
@kuhnroyal Thanks a lot for the approval! @AlexV525 do we have any perspective on when we could release this change?

AlexV525 (approved):
Thanks for the contribution.

yuriymalyts:
Hi guys, when you plan to release this updates?

dipu0:
@AlexV525 
why MultipartFile.clone() will Deprecated? i faced this when i use fromStrem or fromFile in the time of retry. Unhandled Exception: Bad state: The MultipartFile has already been finalized. This typically means you are using the same MultipartFile in repeated requests.

but MultipartFile.fromFileSync(value.path).clone(); works great!

AlexV525:
> @AlexV525 why MultipartFile.clone() will Deprecated? i faced this when i use fromStrem or fromFile in the time of retry. Unhandled Exception: Bad state: The MultipartFile has already been finalized. This typically means you are using the same MultipartFile in repeated requests.

I don't see we've deprecated the `MultipartFile.clone()`, only we deprecated the unnamed constructor, and its deprecated message means: *MultipartFile() is not cloneable when the stream is consumed, use the MultipartFile.fromStream instead.*

dipu0:
Then your deprecated message is not clear to me at least! 
But thanks for the quick reply, that clone method is a life saver for me. 

But this dio but why use clones is a nice addition but dio should not throw this error message in the first place. 

Thanks again for clearing my confusion!

PR #1896, opened 2023-07-12
fix(http2_adapter):request by http2 without network once,the same domain request will never success.

sunhapper:

`_ConnectionManager._connect` may throw SocketException when call `await initFuture`,if it occurs, even if the network is restored, subsequent requests with the same domain will still obtain the cached initFuture, and it will still fail

### DIFF
--- a/plugins/http2_adapter/lib/src/connection_manager_imp.dart
+++ b/plugins/http2_adapter/lib/src/connection_manager_imp.dart
@@ -45,12 +45,13 @@ class _ConnectionManager implements ConnectionManager {
       if (initFuture == null) {
         _connectFutures[domain] = initFuture = _connect(options);
       }
-      transportState = await initFuture;
+      transportState = await initFuture.whenComplete(() {
+         _connectFutures.remove(domain);
+      });
       if (_forceClosed) {
         transportState.dispose();
       } else {
         _transportsMap[domain] = transportState;
-        final _ = _connectFutures.remove(domain);
       }
     } else {
       // Check whether the connection is terminated, if it is, reconnecting.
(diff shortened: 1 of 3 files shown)


plugins/http2_adapter/lib/src/connection_manager_imp.dart
```diff
@@ -45,12 +45,13 @@ class _ConnectionManager implements ConnectionManager {
       if (initFuture == null) {
         _connectFutures[domain] = initFuture = _connect(options);
       }
-      transportState = await initFuture;
+      transportState = await initFuture.whenComplete(() {
+         _connectFutures.remove(domain);
```
AlexV525, line 49: Can we remove this only when the future failed?

sunhapper: I changed code only removed cached Future on fail and restored the logic of success to its original state.

AlexV525 (approved):

plugins/http2_adapter/lib/src/connection_manager_imp.dart
```diff
@@ -45,7 +45,12 @@ class _ConnectionManager implements ConnectionManager {
       if (initFuture == null) {
         _connectFutures[domain] = initFuture = _connect(options);
       }
-      transportState = await initFuture;
+      try {
+        transportState = await initFuture;
+      } catch (e) {
```
AlexV525, line 50:
```suggestion
      } catch (_) {
```

sunhapper: Thanks  a lot for the update

Issue #1911, opened 2023-07-31
Error: SocketException type unknown instead of connectionError


zezo357:
### Package

dio

### Version

^5.3.1

### Output of `flutter doctor -v`

```shell
[✓] Flutter (Channel stable, 3.10.6, on macOS 13.4.1 22F770820d darwin-arm64, locale en-EG)
    • Flutter version 3.10.6 on channel stable at /Users/AbdelazizMahdy/Developer/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision f468f3366c (3 weeks ago), 2023-07-12 15:19:05 -0700
    • Engine revision cdbeda788a
    • Dart version 3.0.6
    • DevTools version 2.23.1

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.2)
    • Android SDK at /Users/AbdelazizMahdy/Library/Android/sdk
    • Platform android-33, build-tools 33.0.2
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b802.4-9586694)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14E222b
    • CocoaPods version 1.12.1

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2022.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b802.4-9586694)

[✓] VS Code (version 1.80.1)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.68.0

[✓] Connected device (3 available)
    • iPhone 14 Pro (mobile) • 7E7D7194-D699-40A5-9281-8E5D860CA41F • ios            • com.apple.CoreSimulator.SimRuntime.iOS-16-4 (simulator)
[...]

PR #1912, opened 2023-07-31
Fixing connection error

zezo357:

fixing https://github.com/cfug/dio/issues/1911


dio/test/basic_test.dart
```diff
@@ -44,11 +44,7 @@ void main() {
       throwsA(
         allOf([
           isA<DioException>(),
-          (DioException e) =>
-              e.type ==
-              (isWeb
-                  ? DioExceptionType.connectionError
-                  : DioExceptionType.unknown),
+          (DioException e) => e.type == (DioExceptionType.connectionError),
```
kuhnroyal, line 47: This is good, same behaviour for both platforms

plugins/native_dio_adapter/.flutter-plugins
kuhnroyal, line 1: This should not be here

plugins/native_dio_adapter/.flutter-plugins-dependencies
kuhnroyal, line 1: This should not be here

dio/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: dio
-version: 5.3.1
+version: 5.3.2
```
kuhnroyal, line 2: You can revert the version bump here and in the changelog, we do this automatically when releasing.

dio/lib/src/dio_exception.dart
```diff
@@ -151,14 +151,14 @@ class DioException implements Exception {
...
-        message: 'The connection errored: $reason',
-        requestOptions: requestOptions,
-        response: null,
-        error: null,
-      );
+          type: DioExceptionType.connectionError,
+          message: 'The connection errored: $reason',
+          requestOptions: requestOptions,
+          response: null,
+          error: error);
```
kuhnroyal, line 161: Missing trailing comma

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -116,7 +116,11 @@ class IOHttpClientAdapter implements HttpClientAdapter {
       });
     } on SocketException catch (e) {
       if (!e.message.contains('timed out')) {
```
kuhnroyal, line 118: Can you swap the condition here? That would make it a little bit easier to read.

kuhnroyal (changes requested):
Thanks for your contribution, I left a couple comments.

kuhnroyal:
@cfug/devs In generell I am not sure if this should be considered a breaking change. I know I have code that relies on the current behaviour.

dio/pubspec.yaml
zezo357: So I just set it as unreleased correct?

kuhnroyal: Yes, in the changelog.

plugins/native_dio_adapter/.flutter-plugins
zezo357: > This should not be here

to resolve this i think i should revert the commit right?

kuhnroyal: Or you can push a new commit where the files get removed, we squash everything in the end.

zezo357: > Or you can push a new commit where the files get removed, we squash everything in the end.

done

dio/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 ## Unreleased
 
 *None.*
```
kuhnroyal, line 8: This can be removed since you added a change

zezo357: thank you, i forgot it

kuhnroyal (commented):
LGTM, just need to figure out if this will break peoples code.

zezo357:
> LGTM, just need to figure out if this will break peoples code.

thank you for the fast and awesome review.

dio/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 
 ## Unreleased
 
-*None.*
+- Fixing requests failing with unknown instead of connection error on SocketException
```
AlexV525, line 8:
```suggestion
- Fix failing requests throw `DioException`s with `.unknown` instead of `.connectionError` on `SocketException`.
```

dio/lib/src/dio_exception.dart
```diff
@@ -151,13 +151,14 @@ class DioException implements Exception {
   factory DioException.connectionError({
     required RequestOptions requestOptions,
     required String reason,
+    Object? error,
```
AlexV525, line 154: cc @ueman to check if adding the field would affect something.

AlexV525 (approved):
RSLGTM

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -115,14 +115,18 @@ class IOHttpClientAdapter implements HttpClientAdapter {
         if (v != null) request.headers.set(k, v);
       });
     } on SocketException catch (e) {
-      if (!e.message.contains('timed out')) {
-        rethrow;
+      if (e.message.contains('timed out')) {
+        throw DioException.connectionTimeout(
```
ueman, line 119: We're now loosing the original stacktrace here. So it should either `Error.throwWithStackTrace()` or the original stacktrace should be passed to the DioException

dio/lib/src/dio_exception.dart
ueman: This is safe to do, because it's not a required field

dio/lib/src/adapters/io_adapter.dart
kuhnroyal: Here we are again 😆  I think the original stacktrace doesn't help the user in figuring out where the problem originated and which Dio call caused it. The stacktrace for the `SocketException` always points to some `dart:io`

AlexV525: Let's just pass it anyway since we'll make a huge refactor using `Error.throwWithStackTrace` in future versions. 🤔

kuhnroyal: But then we lose the one that is important.

kuhnroyal: We don't do it for the others (timeout) etc. either.
I created https://github.com/cfug/dio/discussions/1914 to discuss this problem further.

zezo357:
Is something required from me other than the change log text review?

AlexV525:
@kuhnroyal 's review is still requesting changes. PTAL

zezo357:
added the stacktrace


but this fails in the test 

```
                (e) => e.stackTrace
                     .toString()
                    .contains('test/stacktrace_test.dart'),
```
so i commented it, should i remove it? or leave it so it gets tested when a better logic for stacktrace exists.

kuhnroyal:
> @cfug/devs In generell I am not sure if this should be considered a breaking change. I know I have code that relies on the current behaviour.

Any feedback on this?

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -115,15 +115,20 @@ class IOHttpClientAdapter implements HttpClientAdapter {
...
+          requestOptions: options,
+          timeout: options.connectTimeout ??
+              httpClient.connectionTimeout ??
+              Duration.zero,
+          error: e,
```
kuhnroyal, line 124: We don't pass the original stacktrace here neither.

AlexV525:
> > @cfug/devs In generell I am not sure if this should be considered a breaking change. I know I have code that relies on the current behaviour.
> 
> Any feedback on this?

I would not consider this as a BC.

dio/lib/src/adapters/io_adapter.dart
kuhnroyal: This was not meant a s a TODO. Please revert, sorry.

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -115,15 +115,21 @@ class IOHttpClientAdapter implements HttpClientAdapter {
...
       }
-      throw DioException.connectionTimeout(
+      throw DioException.connectionError(
         requestOptions: options,
-        timeout: options.connectTimeout ??
-            httpClient.connectionTimeout ??
-            Duration.zero,
+        reason: e.message,
         error: e,
+        stackTrace: StackTrace.current,
```
kuhnroyal, line 132: `StackTrace.current` is wrong, either we take the original stacktrace by catching it. Or none which will lead to a usable default.

kuhnroyal: Yea, I understand, sorry.
If the others persist in having the original stacktrace, then using the catched stacktrace is correct.

dio/test/stacktrace_test.dart
kuhnroyal, line 1: @cfug/devs The test changes show that we are changing the stacktrace behavior (again). I suggest to keep the current stacktrace behavior and only change the `DioExceptionType` in the scope of this PR. The rest can be discussed separately.

zezo357: just to be sure, so i should remove the stackTrace added in the connection error? correct?

zezo357:
any idea when can this get merged?

kuhnroyal (approved):
We are not passing the original stacktrace for now, same as before this PR.
Further discussion for this in https://github.com/cfug/dio/discussions/1914

kuhnroyal:
Thanks @zezo357 for your contribution and patience!

PR #1915, opened 2023-08-01
Fix for Dio.download not cleaning the file on data handling error

hgraceb:

### DIFF
--- a/dio/lib/src/dio/dio_for_native.dart
+++ b/dio/lib/src/dio/dio_for_native.dart
@@ -117,9 +117,9 @@ class DioForNative with DioMixin implements Dio {
       if (!closed) {
         closed = true;
         await asyncWrite;
-        await raf.close();
+        await raf.close().catchError((_) => raf);
         if (deleteOnError && file.existsSync()) {
-          await file.delete();
+          await file.delete().catchError((_) => file);
         }
       }
     }
@@ -141,9 +141,9 @@ class DioForNative with DioMixin implements Dio {
           try {
             await subscription.cancel();
             closed = true;
-            await raf.close();
+            await raf.close().catchError((_) => raf);
             if (deleteOnError && file.existsSync()) {
-              await file.delete();
+              await file.delete().catchError((_) => file);
             }
           } finally {
             completer.completeError(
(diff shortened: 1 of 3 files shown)


dio/lib/src/dio/dio_for_native.dart
```diff
@@ -140,6 +140,11 @@ class DioForNative with DioMixin implements Dio {
         }).catchError((Object e) async {
           try {
             await subscription.cancel();
+            closed = true;
+            await raf.close();
```
AlexV525, line 144: We might consider closing the RAF without throwing another exception. And the `closed` might be better after the file is actually closed. So:
```suggestion
            await raf.close().catchError((_) => raf);
            closed = true;
```

dio/lib/src/dio/dio_for_native.dart
```diff
@@ -140,6 +140,11 @@ class DioForNative with DioMixin implements Dio {
...
+            if (deleteOnError && file.existsSync()) {
+              await file.delete();
```
AlexV525, line 146: This operation should be the same:
```suggestion
              await file.delete().catchError((_) => file);
```

dio/lib/src/dio/dio_for_native.dart
hgraceb: 1. I didn't handle the error here out of respect for the previous code. If error catching is necessary here, I made a uniform modification to the previous code as well.

```dart
bool closed = false;
Future<void> closeAndDelete() async {
  if (!closed) {
    closed = true;
    await asyncWrite;
    await raf.close().catchError((_) => raf);
    if (deleteOnError && file.existsSync()) {
      await file.delete().catchError((_) => file);
    }
  }
}
```

2. I think `closed = true` is before `raf.close()` because `closed` is just a flag, mainly to avoid duplicate close and delete operations on the file.

3. In the previous unit test, I simulated a file write failure using the `lock` method. However, as the interface says, the file lock API has different implementations on different platforms, making it impossible to ensure the correctness of the unit test. So I changed it to simulate errors using `onReceiveProgress`.

> NOTE file locking does have slight differences in behavior across platforms:
On Linux and OS X this uses advisory locks, which have the surprising semantics that all locks associated with a given file are removed when any file descriptor for that file is closed by the process. Note that this does not actually lock the file for access. Also note that advisory locks are on a process level. This means that several isolates in the same process can obtain an exclusive lock on the same file.
On Windows the regions used for lock and unlock needs to match. If that is not the case unlocking will result in the OS error "The segment is already unlocked".

AlexV525 (approved):
Great catch, thanks!

PR #1929, opened 2023-08-09
Enable request tests on all platforms

kuhnroyal:


dio/lib/src/adapters/browser_adapter.dart
```diff
@@ -71,7 +71,9 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
           xhr.status!,
           headers: xhr.responseHeaders.map((k, v) => MapEntry(k, v.split(','))),
           statusMessage: xhr.statusText,
-          isRedirect: xhr.status == 302 || xhr.status == 301,
+          isRedirect: xhr.status == 302 ||
+              xhr.status == 301 ||
+              options.uri.toString() != xhr.responseUrl,
```
kuhnroyal, line 76: Judging from my tests, the XHR will never return 301/302 here as the browser or JS-engine does always follow redirects.

dio/test/request_integration_test.dart
```diff
@@ -0,0 +1,336 @@
...
+          queryParameters: {'id': '12', 'name': 'wendu'},
+        );
+        expect(response.statusCode, 200);
+        expect(response.isRedirect, isFalse);
+        expect(response.data['method'], 'GET');
+        expect(response.data['args'], {'id': '12', 'name': 'wendu'});
+      });
+
+      // TODO This is not supported on web, should we warn?
+      test('GET with content', () async {
```
kuhnroyal, line 40: GET with a body is not supported in web. What should we do here? Exception, warning?

dio/test/request_integration_test.dart
```diff
@@ -0,0 +1,336 @@
...
+            'x-multi-value-request-header': ['value1', 'value2'],
+          },
+        ),
+      );
+      expect(response.statusCode, 200);
+      expect(response.isRedirect, isFalse);
+      expect(
+        response.data['headers']['X-Multi-Value-Request-Header'],
+        // TODO we have a diff here between browser and non-browser
+        equals('value1, value2'),
```
kuhnroyal, line 302: Here we have different results on io & web, need to check what is going on.

dio/test/request_integration_test.dart
AlexV525: We should do nothing. It's under a policy as I remember, and XHR does not ring any bells when the requesting is sending a payload.

kuhnroyal: > It's under a policy as I remember

Not sure what you mean by that.

> XHR does not ring any bells when the requesting is sending a payload

That is correct, the body data is just ignored. The dio API allows this but there is no feedback for the developer why it is not working. My concern is that people open bug reports for this :)

AlexV525: > My concern is that people open bug reports for this :)

Adding warning logs like the `ImplyContentTypeInterceptor` might work then.

kuhnroyal: I added a warning

dio/test/request_integration_test.dart
kuhnroyal: I adjusted the browser behaviour here, we were actually sending `List.toString()` which resulted in `[foo, bar]` formatting. Running the web request tests now also against `httpbun` detected this.

AlexV525 (approved):

PR #1947, opened 2023-08-30
fix: in the stream mode, request cannot be canceled

lh900519:

### DIFF
--- a/dio/lib/src/adapters/io_adapter.dart
+++ b/dio/lib/src/adapters/io_adapter.dart
@@ -86,7 +86,10 @@ class IOHttpClientAdapter implements HttpClientAdapter {
     Stream<Uint8List>? requestStream,
     Future<void>? cancelFuture,
   ) async {
-    final httpClient = _configHttpClient(options.connectTimeout);
+    final httpClient = _configHttpClient(
+      options.connectTimeout,
+      newClient: options.responseType == ResponseType.stream,
+    );
     final reqFuture = httpClient.openUrl(options.method, options.uri);
 
     late HttpClientRequest request;
@@ -107,7 +110,17 @@ class IOHttpClientAdapter implements HttpClientAdapter {
       }
 
       if (cancelFuture != null) {
-        cancelFuture.whenComplete(() => request.abort());
+        cancelFuture.whenComplete(() async {
+          // Abort request
+          request.abort();
+
+          if (options.responseType == ResponseType.stream) {
+            // Close the HttpClient, in the stream mode, there is a problem that cannot be canceled
+            httpClient.close(force: true);
+            // Reset HttpClient cache
+            _cachedHttpClient = null;
+          }
+        });
       }
 
       // Set Headers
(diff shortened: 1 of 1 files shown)


kuhnroyal:
Thanks for your PR!

Do you have any additional information for the root problem?
The fix currently won't work when thinking about parallel requests, the closing of the HTTP client may kill other requests.
Please provide a test that reproduces the problem.

lh900519:
```dart
import 'dart:async';
import 'dart:convert';
import 'dart:typed_data';

void main() async {
  final dio = Dio();
  clientStream(dio, '1');
  clientStream(dio, '2');
}

Future clientStream(Dio dio, String name) async {
  try {
    final token = CancelToken();
    Timer(const Duration(seconds: 3), () {
      token.cancel("----Canceled");
    });
    const url = "https://hacker-news.firebaseio.com/v0/topstories.json";
    Response<ResponseBody> rs = await dio.get<ResponseBody>(
      url,
      options: Options(
        headers: {
          "Authorization": 'vhdrjb token"',
          "Accept": "text/event-stream",
          "Cache-Control": "no-cache",
        },
        responseType: ResponseType.stream,
      ), // set responseType to `stream`
      cancelToken: token,
    );

    rs.data?.stream
        .transform(unit8Transformer)
        .transform(const Utf8Decoder())
        .transform(const LineSplitter())
        .listen((event) {
      if (event.startsWith("event:")) {
        print("$name $event isCancelled:${token.isCancelled}");
      }
      if (event.startsWith("data:")) {
        print(
            "$name ${event.substring(0, 5)} isCancelled:${token.isCancelled}");
      }
    });
  } catch (e) {
    print('error $e');
  }
}

StreamTransformer<Uint8List, List<int>> unit8Transformer =
    StreamTransformer.fromHandlers(
  handleData: (data, sink) {
    sink.add(List<int>.from(data));
  },
);

```

Run the code with `dart test.dart`
The request was canceled using CancelToken after 3 seconds, but the terminal did not stop and continued to output.

This PR does not reuse _cachedHttpClient for stream type requests, but creates a new httpclient, and closes the newly created httpclient when using CancelToken. The processing of stream types here is the same as in 5.1.2.

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -227,7 +238,13 @@ class IOHttpClientAdapter implements HttpClientAdapter {
     );
   }
 
-  HttpClient _configHttpClient(Duration? connectionTimeout) {
+  HttpClient _configHttpClient(Duration? connectionTimeout,
+      {bool newClient = false}) {
```
AlexV525, line 242: These are not well-formatted (with trailing commas).

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -227,7 +238,13 @@ class IOHttpClientAdapter implements HttpClientAdapter {
...
+    if (newClient) {
+      final client = _createHttpClient();
```
AlexV525, line 244: If we create new clients here, what about the closing after the client has been used?

dio/lib/src/adapters/io_adapter.dart
AlexV525, line 242: I'd prefer `forceNotCachedClient` for better understanding, or someone else can have better namings.

AlexV525 (changes requested):
If you have a reproducible example, please make it a test, we need tests to avoid further breakages.

kuhnroyal:
I am also wondering if there is an upstream issue for this problem. We `abort()` the request correctly, why does this not work?

kuhnroyal:
Maybe this is the source of the problem? https://github.com/dart-lang/sdk/issues/51267

AlexV525:
Converting this to a draft until we get progress with the implementation and tests.

AlexV525:
Should be covered by #2032.

PR #1948, opened 2023-08-31
[cookie_manager] Fix `FileSystemException` when saving redirect cookies without a proper `host`

s0nerik:
Upon relative redirect, cookies were not saved because the `host` was missing. `resolveUri` provides the proper host info for a redirect Uri in such cases and doesn't affect others.

### DIFF
--- a/plugins/cookie_manager/CHANGELOG.md
+++ b/plugins/cookie_manager/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-*None.*
+- Fix `FileSystemException: Cannot open file ... (OS Error: Is a directory, errno = 21)`
 
 ## 3.1.0+1
 
--- a/plugins/cookie_manager/CHANGELOG.md
+++ b/plugins/cookie_manager/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Fix `FileSystemException: Cannot open file ... (OS Error: Is a directory, errno = 21)`
+- Fix `FileSystemException` when saving redirect cookies without a proper `host`.
 
 ## 3.1.0+1
 
--- a/plugins/cookie_manager/lib/src/cookie_mgr.dart
+++ b/plugins/cookie_manager/lib/src/cookie_mgr.dart
@@ -123,6 +123,7 @@ class CookieManager extends Interceptor {
     // users will be available to handle cookies themselves.
     final isRedirectRequest = statusCode >= 300 && statusCode < 400;
     // Saving cookies for the original site.
+    // Spec: https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2.
     final originalUri = response.requestOptions.uri;
     final realUri = originalUri.resolveUri(response.realUri);
     await cookieJar.saveFromResponse(realUri, cookies);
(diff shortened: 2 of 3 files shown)


ueman (changes requested):
Please add a test for this change

s0nerik:
@ueman I've added a test.

plugins/cookie_manager/test/cookies_persistance_test.dart
```diff
@@ -0,0 +1,115 @@
...
+  Future<void> saveFromResponse(Uri uri, List<Cookie> cookies) async {
+    _saveCalls.add(_SaveCall(
+      uri.toString(),
+      cookies.join('; '),
+    ));
+  }
+}
+
+void main() {
+  test(
```
ueman, line 81: Can you add a test that makes sure non relative redirects aren't modified?
I.e.: a redirect from `example.org` to `sample.com` should not write a cookie to `example.org`.

@kuhnroyal or @AlexV525 do you happen to know whether this is or can be a problem? I'm not really knowledgable in this area.

s0nerik: Done.

AlexV525:
I was wondering if this should be resolved from the `cookie_jar`. @s0nerik

s0nerik:
> I was wondering if this should be resolved from the `cookie_jar`. @s0nerik 

It doesn't look like a problem with the `cookie_jar` itself to me. `cookie_jar` lays responsibility of providing a proper host for saving cookies on a caller. The error it gives upon providing it with an incorrect host could've been better, though.

The `CookieManager` just wasn't providing it with a proper host info upon a relative redirect.

plugins/cookie_manager/lib/src/cookie_mgr.dart
```diff
@@ -123,7 +123,9 @@ class CookieManager extends Interceptor {
     // users will be available to handle cookies themselves.
     final isRedirectRequest = statusCode >= 300 && statusCode < 400;
     // Saving cookies for the original site.
-    await cookieJar.saveFromResponse(response.realUri, cookies);
+    final originalUri = response.requestOptions.uri;
+    final realUri = originalUri.resolveUri(response.realUri);
```
AlexV525, line 127: Do we have any standards that indicate we should resolve the real dest based on the original one?

AlexV525 (commented):
Sry for the late review

plugins/cookie_manager/lib/src/cookie_mgr.dart
s0nerik: https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2

> The field value consists of a single URI-reference.  When it has the
   form of a relative reference ([RFC3986], Section 4.2), the final
   value is computed by resolving it against the effective request URI
   ([RFC3986], Section 5).

AlexV525: Great. Could you add it as a comment, so we can track it back somedays if we forgot it.

plugins/cookie_manager/CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-*None.*
+- Fix `FileSystemException: Cannot open file ... (OS Error: Is a directory, errno = 21)`
```
AlexV525, line 5:
```suggestion
- Fix `FileSystemException` when saving redirect cookies without a proper `host`.
```

plugins/cookie_manager/lib/src/cookie_mgr.dart
AlexV525: @s0nerik

AlexV525 (approved):
LGTM after the above comments are resolved

PR #1974, opened 2023-09-20
Make use of new dart api for compute

ueman:
Get rid of a lot of code by using a newer Dart API. Also fix some unrelated issues on the v6 branch.

This targets the v6 branch


kuhnroyal:
Pretty neat to get rid of the custom handling but can we actually raise to Dart 3.0.0? I am all for it :)

ueman:
> Pretty neat to get rid of the custom handling but can we actually raise to Dart 3.0.0? I am all for it :)

Technically, this change only needs Dart 2.19. But given, the ecosystem is quite good at being up to date, I would say it's okay.

AlexV525:
I would be excited to see records and patterns in our packages if we use 3.0 😮. However, if we do so, we might need to consider maintaining separate branches given the fact that thousands (on GitHub) of projects are using Dio 4.0.

ueman:
I reduced the Dart version down to 2.19. That's still higher than the current, but it should be acceptable.

dio/lib/src/dio/dio_for_browser.dart
```diff
@@ -10,7 +10,7 @@ import '../response.dart';
 Dio createDio([BaseOptions? options]) => DioForBrowser(options);
 
 /// Implements features for [Dio] on Web platforms.
-class DioForBrowser with DioMixin implements Dio {
+class DioForBrowser extends DioMixin implements Dio {
```
AlexV525, line 13: Should we make `DioMixin` as mixin instead?

ueman: Yeah, I guess that makes sense. Not in this PR though.

AlexV525 (approved):
RSLGTM

PR #1981, opened 2023-09-23
🔥 Clean

AlexV525:
Here is some cleaning again. :)

- Extract logging calls to `debugLog`.
- Promote types for `encodeMap` (which I've already done in part of #1794 but even more).
- Raise warning for `Map`s other than `Map<String, dynamic>` for encoding.

### DIFF
--- a/dio/lib/src/utils.dart
+++ b/dio/lib/src/utils.dart
@@ -1,9 +1,17 @@
 import 'dart:async';
 import 'dart:collection';
 import 'dart:convert';
+import 'dart:developer' as dev;
 
 import 'options.dart';
 import 'parameter.dart';
+
+// For the web platform, an inline `bool.fromEnvironment` translates to
+// `core.bool.fromEnvironment` instead of correctly being replaced by the
+// constant value found in the environment at build time.
+//
+// See https://github.com/flutter/flutter/issues/51186.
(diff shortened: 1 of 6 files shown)


dio/lib/src/utils.dart
```diff
@@ -133,3 +143,14 @@ Map<String, V> caseInsensitiveKeyMap<V>([Map<String, V>? value]) {
   if (value != null && value.isNotEmpty) map.addAll(value);
   return map;
 }
+
+void debugLog(String message, StackTrace stackTrace) {
```
ueman, line 147: Can we make this a static set-able property on the Dio class? That way this could be configured. On Flutter a user could use `debugPrint` or a custom logging solution for example.

AlexV525: I need to rethink about it as a public API.

AlexV525: Started a new discussion https://github.com/cfug/dio/discussions/1982. May I add a TODO to the method first so we can continue rolling this?

ueman: Sounds good 👍

kuhnroyal (approved):
LGTM, probably should add a line in the changelog or migration guide about the new map encoding warning.

PR #1998, opened 2023-10-16
Improve exception message for DioException.badResponse

ueman:
Since we receive a lot of issues around the topic of status code, I think we should improve message of it. So I've made it more descriptinve and actionable.

### DIFF
--- a/dio/lib/src/dio_exception.dart
+++ b/dio/lib/src/dio_exception.dart
@@ -86,8 +86,7 @@ class DioException implements Exception {
   }) =>
       DioException(
         type: DioExceptionType.badResponse,
-        message: 'The request returned an '
-            'invalid status code of $statusCode.',
+        message: _badResponseExceptionMessage(statusCode),
         requestOptions: requestOptions,
         response: response,
         error: null,
@@ -210,5 +209,39 @@ class DioException implements Exception {
       msg += '\nError: $error';
     }
     return msg;
+  }
+
+  /// Because of [ValidateStatus] we need to consider all status codes when
+  /// creating a [DioException.badResponse].
+  static String _badResponseExceptionMessage(int statusCode) {
+    var message = '';
+    if (statusCode >= 100 && statusCode < 200) {
+      message =
+          'This is an informational response - the request was received, continuing processing';
+    } else if (statusCode >= 200 && statusCode < 300) {
+      message =
+          'The request was successfully received, understood, and accepted';
+    } else if (statusCode >= 300 && statusCode < 400) {
+      message =
+          'Redirection: further action needs to be taken in order to complete the request';
+    } else if (statusCode >= 400 && statusCode < 500) {
+      message =
+          'Client error - the request contains bad syntax or cannot be fulfilled';
+    } else if (statusCode >= 500 && statusCode < 600) {
+      message =
+          'Server error - the server failed to fulfil an apparently valid request';
+    } else {
+      message =
+          'A response with a status code that is not within the range of inclusive 100 to exclusive 600'
+          'is a non-standard response, possibly due to the server\'s software.';
+    }
+
+    return ''
+        'This exception was thrown because the response has a status code of $statusCode '
+        'and RequestOptions.validateStatus was configured to throw for this status code.\n'
+        'The status code of $statusCode has the following meaning: "$message"\n.'
+        'Read more about status codes at https://developer.mozilla.org/en-US/docs/Web/HTTP/Status\n'
+        'In order to resolve this exception you typically have either to verify '
+        'and fix your request code or you have to fix the server code.';
   }
 }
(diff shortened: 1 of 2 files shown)


kuhnroyal:
Nice, just think we should still print the actual status code somewhere.
Something like `Server error (500) - ...`

ueman:
> Nice, just think we should still print the actual status code somewhere. Something like `Server error (500) - ...`

It does right in the first line of the exception message:
`'This exception was thrown because the response has a status code of $statusCode '`

kuhnroyal:
My bad...

kuhnroyal (approved):

dio/lib/src/dio_exception.dart
```diff
@@ -211,4 +210,38 @@ class DioException implements Exception {
...
+    } else if (statusCode >= 400 && statusCode < 500) {
+      message =
+          'Client error - the request contains bad syntax or cannot be fulfilled';
+    } else if (statusCode >= 500 && statusCode < 600) {
+      message =
+          'Server error - the server failed to fulfil an apparently valid request';
+    } else {
+      message =
+          'A response with a status code that is not within the range of inclusive 100 to exclusive 600'
+          'is a non-standard response, possibly due to the server\'s software.';
```
AlexV525, line 236: Seems to be a duplicated dot.

dio/lib/src/dio_exception.dart
```diff
@@ -211,4 +210,38 @@ class DioException implements Exception {
     }
     return msg;
   }
+
+  /// Because of [ValidateStatus] we need to consider all status codes when
+  /// creating a [DioException.badResponse].
+  static String _badResponseExceptionMessage(int statusCode) {
+    var message = '';
```
AlexV525, line 217: I used to write `final String message` if it's declared in every single flow.

dio/lib/src/dio_exception.dart
```diff
@@ -211,4 +210,38 @@ class DioException implements Exception {
...
+    } else if (statusCode >= 500 && statusCode < 600) {
+      message =
+          'Server error - the server failed to fulfil an apparently valid request';
+    } else {
+      message =
+          'A response with a status code that is not within the range of inclusive 100 to exclusive 600'
+          'is a non-standard response, possibly due to the server\'s software.';
+    }
+
+    return ''
```
AlexV525, line 239: Maybe `StringBuffer` would be a good choice at here.

AlexV525 (commented):
Late review comments

dio/lib/src/dio_exception.dart
ueman: I'm not really aware of the benefits of a `StringBuffer`. Can you explain why it would improve this code?

dio/lib/src/dio_exception.dart
ueman: Yeah, but I rather want the message to be helpful than concise.

dio/lib/src/dio_exception.dart
AlexV525: > I'm not really aware of the benefits of a `StringBuffer`. Can you explain why it would improve this code?

You just don't have to constantly write `\n` then.

dio/lib/src/dio_exception.dart
AlexV525: > Yeah, but I rather want the message to be helpful than concise.

Hmm, what I mean is the last dot of `possibly due to the server\'s software.` might be redundant.

ueman: Ah I see, yeah, that might be redundant, but this message is aimed at people who are not yet that knowledgeable of the HTTP spec and so on. So I think it's a rather helpful addition, even if it might seem superfluous for us knowledgeable people.

ueman:
@AlexV525 I plan to do another round of exception message improvements for the other exceptions. I'll incorporate your feedback in that PR, okay?

AlexV525:
> @AlexV525 I plan to do another round of exception message improvements for the other exceptions. I'll incorporate your feedback in that PR, okay?

Yeah sure. I'm available now

dio/lib/src/dio_exception.dart
AlexV525:
```console
DioException [bad response]: This exception was thrown because the response has a status code of 0 and RequestOptions.validateStatus was configured to throw for this status code.
The status code of 0 has the following meaning: "A response with a status code that is not within the range of inclusive 100 to exclusive 600is a non-standard response, possibly due to the server's software."
.Read more about status codes at https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
```

Here it should be `"server's software.".\nRead more`, right? It's `"server's software."\n.Read more` currently.

AlexV525: which made that dot redundant

Issue #1102, opened 2021-03-24
Dio force header to lowercase


zhangxiongbiao:
### Issue Info

| Info                           | Value                                                 |      |
| ------------------------------ | ----------------------------------------------------- | ---- |
| Platform Name                  | flutter                          |      |
| Platform Version               |                              |      |
| Dio Version                    | 3.0.10                                  |      |
| Android Studio / Xcode Version |  Android Studio 4.1.2              |      |
| Repro rate                     | all the time (100%)   |      |
| Repro with our demo prj        |    |      |
| Demo project link              | |      |
| Dart SDK|2.12.0 |      |

### Issue Description and Steps

When I set Dio header like this

```
dio.options.headers['Authorization'] = 'dGVzdDp0ZXN0';
```

I expect request header to look like this

```
Authorization: Basic dGVzdDp0ZXN0
```

Not like this (notice that capital "A" was transformed to "a")

```
authorization: Basic dGVzdDp0ZXN0
```

I found problem in `io_adpater.dart`

![image](https://user-images.githubusercontent.com/10623853/112324131-697efd80-8ced-11eb-893b-e3207dd5cd44.png)

cirediew:
I think this is related to [this](https://github.com/dart-lang/http2/issues/49)

amirulabu:
I also having this issue, this is mainly due to our flutter app is connecting to a service that is using AWS API Gateway and for the service, header field name is case-sensitive. 

> Header names and query parameters are processed in a case-sensitive way.
https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html

Issue #788, opened 2020-05-15
options headers key lowercase


LiangLuDev:
```
    Options options = Options();
    options.headers['AppID'] = '2000004950';
    var response = await http.post<Map>('/api/checkcfg/2000004950',
        data: {
          "keywords": firmwareVersion,
          "cmpmethod": 1,
          "version": version,
          "imei": imei,
        },
        options: options
    );
```

print 

```
headers : {content-type: application/json; charset=utf-8, appid: 2000004950, Timestamp: 1589550316, Sign: , Connection: close}
```

but can use InterceptorsWrapper 

```
class HeaderInterceptor extends InterceptorsWrapper {
  @override
  onRequest(RequestOptions options) async {
    options.connectTimeout = 1000 * 45;
    options.receiveTimeout = 1000 * 45;
    options.headers['AppID'] = '2000005054';
    debugPrint('HeaderInterceptor : ${options.headers}');
    return options;
  }
}
```

log print
```
headers : {content-type: application/json; charset=utf-8, appid: 2000004950, AppID: 2000005054, Timestamp: 1589551032, Sign: , Connection: close}
```

shaxxx:
Possible duplicate of https://github.com/flutterchina/dio/issues/641

Issue #641, opened 2020-01-10
Dio is forcing headers to lowercase


shaxxx:
**Dio version:** `dio: ^3.0.8`

When I set Dio header like this

`dio.options.headers['Authorization'] = 'dGVzdDp0ZXN0';`

I expect request header to look like this

`Authorization: Basic dGVzdDp0ZXN0`

Not like this (notice that capital "A" was transformed to "a")

`authorization: Basic dGVzdDp0ZXN0`

Reason for this is [HERE](https://github.com/flutterchina/dio/blob/5e3eb797a4202b9aca0acc74d325c10523b79584/dio/lib/src/options.dart#L354). All headers are transformed to lower case.
I can only guess what your motivation for this was, but this is wrong.

Please read about it [HERE](https://github.com/dart-lang/sdk/issues/25120) and [HERE](https://github.com/dart-lang/sdk/issues/33501)

Even Dart team is currently reverting this 
https://dart-review.googlesource.com/c/sdk/+/119100

Dio 2.1.5 did not have this behavior. I was able to use [AltHttpClient](https://pub.dev/packages/alt_http) with Dio and headers where left AS IS.

shaxxx:
After some inspection I can confirm that version Dio 3.0.7 works as expected, it doesn't change headers casing. Problem is [commit ](https://github.com/flutterchina/dio/commit/f77d6db80bfdbd3a32877aedd2273e2116666340#diff-d8b422c1405a3050f793a93dbd587b4d) by @wendux 12 days ago.
I'm even more sure now that this is bug since entire commit was marked as "[code style improvement](dio/lib/src/options.dart)". Please revert changes.

shaxxx:
Yes, this issues is very much still open

yangyangyyluck:
same problem. 
Dio version 3.0.8
the bug came to assail me.
I need headers like CamelCase.
user-agent => User-Agent 
referer => Referer
content-type => Content-Type
and so on...
as @shaxxx  say, please revert changes.

shaxxx:
FYI: Dart team has enabled support for case sensitive headers in the new upcoming version.
New Dart (and Flutter) release will support case sensitive headers, so should Dio.

masseelch:
But the RFC clearly states, that headers are case-insensitive. 
This means it should not matter if dio sends "content-type" or "Content-Type" or "cOntEnt-TyPE". 
The server has the bug here. Not this package.

shaxxx:
[...]

PR #2008, opened 2023-10-23
Case-sensitive headers

### LINKED ISSUE #2002: The request headers key changes to lowercase
### Request Statement

When I made a network request, Authorization was added to the network request headers, but the result was changed to lowercase.
example：
var dio = Dio(); Options options = Options(headers: {"Authorization": "Basic xxx"}); Response response = await dio.get(“xxx” options: options, );

276073621-af600c08-56c8-47ab-9610-e58857286808

### Solution Brainstorm

_No response_

AlexV525:
Fixes #2002 #1102 #788 #641

We have `camelCaseContentDisposition` for `FormData` but this is not related.


dio/lib/src/options.dart
```diff
@@ -352,10 +363,20 @@ class Options {
...
 
+  /// Whether the header keys should be case-insensitive.
+  /// e.g.: `Content-Type` will be sent as `content-type` eventually.
+  ///
+  /// Defaults to true.
+  ///
+  /// This option WILL NOT take effect on these circumstances:
+  /// - XHR ([HttpRequest]) does not support handling this explicitly.
+  /// - The HTTP/2 standard only supports lowercase header keys.
+  bool? caseInsensitiveHeaders;
```
kuhnroyal, line 378: The negation in the name sounds weird, dies it make sense to call it `caseSensitiveHeaders` and default to false?

AlexV525: Updated.

AlexV525:
@kuhnroyal I'm running into mockito issues when testing with headers. Could you help me to solve it? Run "Headers can be case-sensitive" in `options_test.dart` and you'll see this:
```console
Error: MissingStubError: 'headers'
No stub was found which matches the arguments of this method call:
headers
```

AlexV525:
The options tests is blocked by https://github.com/sharat87/httpbun/issues/11.

kuhnroyal:
No idea but you seem to have solved it. Probably re-generate with build_runner.

dio/lib/src/dio_mixin.dart
```diff
@@ -517,7 +517,10 @@ abstract class DioMixin implements Dio {
         stream,
         cancelToken?.whenCancel,
       );
-      final headers = Headers.fromMap(responseBody.headers);
+      final headers = Headers.fromMap(
+        responseBody.headers,
+        caseSensitive: reqOpt.caseSensitiveHeaders,
```
ueman, line 522: I guess it makes sense to stick to one naming-pattern, so it should be always `preserveHeaderCase`.

dio/lib/src/options.dart
```diff
@@ -352,10 +361,19 @@ class Options {
 
   /// HTTP request headers.
   ///
-  /// The keys of the header are case-insensitive,
-  /// e.g.: content-type and Content-Type will be treated as the same key.
+  /// The equality of the header keys is case-insensitive,
```
ueman, line 364: What about the following:

By default header keys are compared without taking the case into consideration, e.g.: `content-type` and `Content-Type` will be treated as the same key. To change this set [caseSensitiveHeaders] to [true]

AlexV525: The fact is here we don't use case-sensitive map even if the flag is true. Otherwise it will be too complicated to handle reading fields like `Headers[Headers.contentTypeHeader]` which takes massive conversion if we do this.

ueman: Ohh, I missed that. Okay, then just disregard this comment

ueman:
Apparently, this is a breaking change since it breaks tests for the native dio adapter:

>     error - lib/src/conversion_layer_adapter.dart:46:21 - The getter 'preserveHeaderCase' isn't defined for the type 'RequestOptions'. Try importing the library that defines 'preserveHeaderCase', correcting the name to the name of an existing getter, or defining a getter or field named 'preserveHeaderCase'. - undefined_getter
>     error - lib/src/conversion_layer_adapter.dart:78:17 - The getter 'preserveHeaderCase' isn't defined for the type 'RequestOptions'. Try importing the library that defines 'preserveHeaderCase', correcting the name to the name of an existing getter, or defining a getter or field named 'preserveHeaderCase'. - undefined_getter

as seen in https://github.com/cfug/dio/actions/runs/6641569223/job/18044420186?pr=2008

AlexV525:
I'll tweak melos configs somedays later before I fixes these tests. Before that, the overall implementation can be reviewed.

kuhnroyal (approved):
LGTM

AlexV525:
Yep. The analysis passed as expected. The last thing would be fixing headers conversion.

AlexV525:
Since we cannot get the compatibility from httpbun, we need to update the tests accordingly.

AlexV525:
Finally, mastering Mockito. :)

PR #2011, opened 2023-10-25
feat(http2_adapter): gzip

Reprevise:

Adds gzip encoding/decoding to the http2 adapter because it doesn't seem like Dart will add gzip to it themselves. I welcome comment as I'm not sure this is the _best_ way to do this _and_ because I don't have a test for encoding gzip and I'd like some help making an adequate test. Do let me know if this is the direction you want to go in, that is, adding gzip to the adapter.

### DIFF
--- a/plugins/http2_adapter/lib/src/http2_adapter.dart
+++ b/plugins/http2_adapter/lib/src/http2_adapter.dart
@@ -59,6 +59,9 @@ class Http2Adapter implements HttpClientAdapter {
       Header.ascii(':authority', uri.host),
     ];
 
+    final requestEncoding =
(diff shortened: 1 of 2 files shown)


AlexV525:
Isn't the GZIP feature already embedded? If not, shall we provide this feature from the http2 package rather than here?

AlexV525:
> It should be added to `http2` instead of here, but it doesn't seem like the Dart team is intent on doing it. I don't think I'd be able to do it myself as I'm not _too_ familiar with the protocol.

From your statement I don't think there is anything we can do. If we gonna support compression like this basis, we should request it from our upstream.

PR #2021, opened 2023-11-02
⚡️ Improve melos and workflow setup

AlexV525:
- Reduce separate workflows and matrices to the minimum (`30+` → `3`).
- Add a script to apply file patches for corresponding SDKs.
- Add a script to produce `.melos_ignored_packages` for steps.
- Annotate test platforms for plugins.
- Fix linter recognition between Dart SDK versions.

Melos has poor support around versions and packages filtering. It has `MELOS_PACKAGES` and `packageFilter` for sessions, but those fields are not working well across its major versions.

The main idea about filtering our packages is to prevent unsatisfied packages from being handled by melos process. For example, if `native_dio_adapter` requires Flutter `3.13.0` and the current Flutter is `2.15.1`, exclude it from packages.

To do this, I've added a script to serialize the melos workspace first, then determine whether the Dart version satisfies the package version. It produces package lists according to demand (satisfied packages or ignored packages).


AlexV525:
`native_dio_adapter` has a different SDK requirement than others, so the `min` workflow fails. I didn't figure out a proper way to handle this with melos at the moment. Leaving this until I finish my exam. Any ideas are welcomed.

scripts/sdk_satisfaction
AlexV525, line 1: Compiled by Dart....Huge one

AlexV525:
Melos. :)

AlexV525:
See the edited top comment for more info.

AlexV525:
Also cc @CaiJingLong to see if you have any idea regarding infrastructure abstraction abilities.

kuhnroyal:
Personally, I don't like the combination of format/analyze/tests as this makes it harder to determine why the build failed, especially for new contributors.
Are there any CI limits reached regarding execution time or matrixes?

AlexV525:
> Personally, I don't like the combination of format/analyze/tests as this makes it harder to determine why the build failed, especially for new contributors.

Good point. If we break it into steps, would it be good enough too?

> Are there any CI limits reached regarding execution time or matrixes?

As far as I know there are no limitations for public repos.

AlexV525:
https://github.com/cfug/dio/actions/runs/6785748920/job/18444745192

Steps were split in the job. Now you're able to identify every verify step in a single workflow.

The core is time trade with speed. If we still use job strategies, it will cost massive redundant time since the workspace needs to be reconfigured each time a job starts. But it can display every action as a workflow.

CaiJingLong (approved):
LGTM

Issue #1739, opened 2023-03-14
receiveTimeout not working on iOS


apaolino:
### Package

dio

### Version

5.0.0

### Output of `flutter doctor -v`

```shell
[✓] Flutter (Channel stable, 3.7.7, on macOS 13.2.1 22D68 darwin-arm64, locale en-IT)
    • Flutter version 3.7.7 on channel stable at /Users/alessandro/development/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 2ad6cd72c0 (6 days ago), 2023-03-08 09:41:59 -0800
    • Engine revision 1837b5be5f
    • Dart version 2.19.4
    • DevTools version 2.20.1

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.1)
    • Android SDK at /Users/alessandro/Library/Android/sdk
    • Platform android-33, build-tools 33.0.1
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 14.2)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 14C18
    • CocoaPods version 1.11.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2022.1)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)

[✓] VS Code (version 1.75.0)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.58.0

[✓] Connected device (3 available)
    • iPhone 14 (mobile) • DF033D85-6295-4A8B-BA01-A8C885178B02 • ios            • com.apple.CoreSimulator.SimRuntime.iOS-16-2 (simulator)
    • macOS (desktop)    • macos                                • darwin-arm64   • macOS 13.2.1 22D68 darwin-arm64
[...]

PR #2029, opened 2023-11-12
⚡ Ignore zero duration timeout

### LINKED ISSUE #2022: Download  receiveTimeout
### Package

dio

### Version

5.3.3

### Operating-System

Android

```shell
┌─────────────────────────────────────────────────────────┐
│ A new version of Flutter is available!                  │
│                                                         │
│ To update to the latest version, run "flutter upgrade". │
└─────────────────────────────────────────────────────────┘
[✓] Flutter (Channel stable, 3.13.6, on macOS 14.0 23A344 darwin-arm64, locale
    zh-Hans-CN)
[✓] Android toolchain - develop for Android devices (Android SDK version 32.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 15.0)
[✓] Android Studio (version 2022.3)
[✓] IntelliJ IDEA Ultimate Edition (version 2023.1.1)
[✓] VS Code (version 1.77.1)
[✓] Connected device (4 available)
```

### Dart Version

_No response_

### Steps to Reproduce

In previous versions, it was only necessary to set the receiveTimeout of the Download Options to 0 to download without timeout. Now, after changing to the Duration type, the time cannot be unlimited

```dart
Future downloadFile(urlPath, savePath,
      {CancelToken? cancelToken, ProgressCallback? onReceiveProgress}) async {
    Response? response;
    response = await dio.download(urlPath, savePath,
        cancelToken: cancelToken,
        options: Options(receiveTimeout: Duration(seconds: 0)),
        onReceiveProgress: onReceiveProgress ??
            (int count, int total) {
              logI("$count $total");
            });
    return response;
  }
```

### Expected Result

download success

### Actual Result

[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: DioException [receive timeout]: The request took longer than 0:00:00.000000 to receive data. It was aborted.

AlexV525:
Resolves #2022.

I've found that the `IOHttpClientAdapter` uses `Stopwatch` to integrate with `receiveTimeout`, which might cause infinity awaits if the stream has no response forever. This might be the root cause of #1739.
https://github.com/cfug/dio/blob/78f3813a8d8948887198ef628e5a2e2039489b03/dio/lib/src/adapters/io_adapter.dart#L194-L200

### DIFF
--- a/dio/lib/src/adapters/browser_adapter.dart
+++ b/dio/lib/src/adapters/browser_adapter.dart
@@ -58,13 +58,11 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
       }
     });
 
-    final connectTimeout = options.connectTimeout;
-    final receiveTimeout = options.receiveTimeout;
-    int xhrTimeout = 0;
-    if (connectTimeout != null &&
-        receiveTimeout != null &&
-        receiveTimeout > Duration.zero) {
-      xhrTimeout = (connectTimeout + receiveTimeout).inMilliseconds;
+    final sendTimeout = options.sendTimeout ?? Duration.zero;
+    final connectTimeout = options.connectTimeout ?? Duration.zero;
+    final receiveTimeout = options.receiveTimeout ?? Duration.zero;
+    final xhrTimeout = (connectTimeout + receiveTimeout).inMilliseconds;
+    if (xhrTimeout > 0) {
       xhr.timeout = xhrTimeout;
     }
 
@@ -86,22 +84,19 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
     });
 
     Timer? connectTimeoutTimer;
-
-    final connectionTimeout = options.connectTimeout;
-    if (connectionTimeout != null) {
+    if (connectTimeout > Duration.zero) {
       connectTimeoutTimer = Timer(
-        connectionTimeout,
+        connectTimeout,
         () {
           if (completer.isCompleted) {
             // connectTimeout is triggered after the fetch has been completed.
             return;
           }
-
           xhr.abort();
           completer.completeError(
             DioException.connectionTimeout(
               requestOptions: options,
-              timeout: connectionTimeout,
+              timeout: connectTimeout,
             ),
             StackTrace.current,
           );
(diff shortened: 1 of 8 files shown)


kuhnroyal:
Can we make the timeouts in the options class non-nullable and initialize them with zero? That would make a couple things more readable.

AlexV525:
> Can we make the timeouts in the options class non-nullable and initialize them with zero? That would make a couple things more readable.

You mean zero-timeout is better than non-timeout or sth? I'm okay with both of them though.

kuhnroyal:
Specifically the `RequestOptions.*Timeout` - that would avoid the `options.*Timeout ?? Duration.zero` in all adapters.

AlexV525:
> Specifically the `RequestOptions.*Timeout` - that would avoid the `options.*Timeout ?? Duration.zero` in all adapters.

Ah, then I think we cannot do this because we compose options into `RequestOptions` like `Options.xxx ?? BaseOptions.xxx`. If we change the nullability, the fallback becomes invalid.

AlexV525:
> The `RequestOptions` are used to communicate between `DioMixin` and adapters right? We should be able to handle this in the `DioMixin` in a single place.

I've tried to do this but configs are extended and mixin with multiple classes. It's relatively unactionable at the moment. You can give it a try if you want to.

dio/lib/src/adapters/browser_adapter.dart
```diff
@@ -58,13 +58,11 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
...
-        receiveTimeout != null &&
-        receiveTimeout > Duration.zero) {
-      xhrTimeout = (connectTimeout + receiveTimeout).inMilliseconds;
+    final sendTimeout = options.sendTimeout ?? Duration.zero;
+    final connectTimeout = options.connectTimeout ?? Duration.zero;
+    final receiveTimeout = options.receiveTimeout ?? Duration.zero;
+    final xhrTimeout = (connectTimeout + receiveTimeout).inMilliseconds;
+    if (xhrTimeout > 0) {
       xhr.timeout = xhrTimeout;
     }
```
CaiJingLong, line 67: Should we not assume a default value here by browser, give it zero when `xhrTimeout == 0`?

CaiJingLong (commented):
No other issues found.

PR #2032, opened 2023-11-13
Cancel streamed downloads & fix progress for downloads

kuhnroyal:

### DIFF
--- a/dio/test/cancel_token_test.dart
+++ b/dio/test/cancel_token_test.dart
@@ -45,6 +45,8 @@ void main() {
           cancelToken.cancel();
         });
 
+        expect(response.statusCode, 200);
+
         await expectLater(
           (response.data as ResponseBody).stream.last,
           throwsA(predicate((DioException e) =>
--- a/dio/test/cancel_token_test.dart
+++ b/dio/test/cancel_token_test.dart
@@ -29,6 +29,33 @@ void main() {
       CancelToken().cancel();
     });
 
+    test(
+      'cancels streamed responses',
+      () async {
+        final dio = Dio()..options.baseUrl = 'https://httpbun.com/';
+
+        final cancelToken = CancelToken();
+        final response = await dio.get(
+          'bytes/${1024 * 1024 * 10}',
+          options: Options(responseType: ResponseType.stream),
+          cancelToken: cancelToken,
+        );
+
+        Future.delayed(const Duration(milliseconds: 750), () {
+          cancelToken.cancel();
+        });
+
+        await expectLater(
+          (response.data as ResponseBody).stream.last,
+          throwsA(predicate((DioException e) =>
+              e.type == DioExceptionType.cancel &&
+              e.message!
+                  .contains('The request was manually cancelled by the user'))),
+        );
+      },
+      testOn: 'vm',
+    );
+
     test('cancels multiple requests', () async {
       final client = MockHttpClient();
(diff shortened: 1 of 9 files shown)


dio/test/cancel_token_test.dart
```diff
@@ -29,6 +29,33 @@ void main() {
...
+        final dio = Dio()..options.baseUrl = 'https://httpbun.com/';
+
+        final cancelToken = CancelToken();
+        final response = await dio.get(
+          'bytes/${1024 * 1024 * 10}',
+          options: Options(responseType: ResponseType.stream),
+          cancelToken: cancelToken,
+        );
+
+        Future.delayed(const Duration(milliseconds: 750), () {
```
kuhnroyal, line 44: This can probably be reduced to 500 or even less.

AlexV525:
Before review: Could we abstract the implementation of this cancellation above the specific adapter? If we're doing this in adapters, then every adapter needs to consider the same functionality.

kuhnroyal:
I will see what I can do, but I already know that this is basically not supported on web.

kuhnroyal:
> Before review: Could we abstract the implementation of this cancellation above the specific adapter? If we're doing this in adapters, then every adapter needs to consider the same functionality.

This is hard without breaking. If we want to do this in the `DioMixin` we have no access to the underlying http implementation and can not close the stream/connection. We would have to pass some new argument to the adapters or extend `ResponseBody` with a new `close` function. But this seems like an issue for dio 6.

kuhnroyal:
This is a lot more complex. We have completely different behavior for different adapters when doing streamed responses/downloads and most of it can not be abstracted away.

For `IOHttpClientAdapter` we have a `ResponseBody` with a stream so the `await dio.get` succeeds after the connection is established with a `Response` and the cancellation exception is thrown on the stream.

For `BrowserHttpClientAdapter` we have a `ResponseBody` with bytes and the adapter has to wait until everything is received, before completing with a response. So the cancellation exception is thrown on the `await dio.get`.
And there is no way to access other information from the XHR when aborting.

kuhnroyal:
> What about the HTTP/2 adapter?

Fixed it there as well. I think we should merge this and brain storm if we can abstract this away separately.

AlexV525 (approved):
LGTM

plugins/http2_adapter/lib/src/http2_adapter.dart
```diff
@@ -158,6 +152,19 @@ class Http2Adapter implements HttpClientAdapter {
       cancelOnError: true,
     );
 
+    /// Cancel any up/download streams and connections
+    /// if the [CancelToken] is cancelled
+    /// and propagate the [DioException] to the response stream.
+    cancelFuture?.whenComplete(() {
```
AlexV525, line 158: We have asynchronous gaps between this call and other awaits. Can we do this earlier or separated?

AlexV525:
Is this still on your track? The last time I review this it only remains a few nits.

AlexV525:
@kuhnroyal Is this still on your track? The last time I review this it only remains a few nits.

kuhnroyal:
> @kuhnroyal Is this still on your track? The last time I review this it only remains a few nits.

Yea

kuhnroyal:
Superseeded by https://github.com/cfug/dio/pull/2068 due to lots of updates on main

PR #2043, opened 2023-11-25
Http2 adatper header support dynamic

Passer-by:


Passer-by:
Dio's header is of type Map<String, dynamic>, so the Http2Adapter cannot directly convert the value to a string.
But there is a question about this PR: Why does Dio perform a default trim operation on the value? Unlike the native Dart HttpClient, which does not trim when setting headers. Should I adhere to Dio's standard or the Dart HttpClient standard?

AlexV525:
> Should I adhere to Dio's standard or the Dart HttpClient standard?

Thanks for details. I think we should stick to RFCs instead of Dio's. We'll manage to get rid of the Dio's implementation too.

plugins/http2_adapter/lib/src/http2_adapter.dart
```diff
@@ -61,14 +61,18 @@ class Http2Adapter implements HttpClientAdapter {
...
-          .map(
-            (key) => Header.ascii(
-              key.toLowerCase(),
-              options.headers[key] as String? ?? '',
-            ),
-          )
-          .toList(),
+      options.headers.keys.map(
+        (key) {
+          final String v;
```
AlexV525, line 66: Use `entries` instead.

plugins/http2_adapter/test/http2_test.dart
```diff
@@ -133,4 +133,28 @@ void main() {
...
+          'NumKey': 2,
+          'BooleanKey': false,
+        },
+      ),
+    );
+    expect(res.data.toString(), contains('TEST'));
+    expect(res.data.toString(), contains('Listkey: 1, 2'));
+    expect(res.data.toString(), contains('Stringkey: 1'));
+    expect(res.data.toString(), contains('Numkey: 2'));
+    expect(res.data.toString(), contains('Booleankey: false'));
```
AlexV525, line 158: These toString() calls can combined into only one variable.

plugins/http2_adapter/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Implement `sendTimeout` and `receiveTimeout` for the adapter.
 - Fix redirect not working when requestStream is null.
 - Ignores `Duration.zero` timeouts.
+- `Http2Adapter` header supports dynamic
```
AlexV525, line 13:
```suggestion
- Add more header value types implicit support.
```

AlexV525, line 13: Also, put this in the *Unreleased* section instead of here.

AlexV525 (approved):

PR #2047, opened 2023-11-27
Improve default parameter naming for Interceptor

hgraceb:
Improve default parameter naming for [Interceptor](https://github.com/cfug/dio/blob/a7a787840b87def9d16c10c009c548ddfcd86337/dio/lib/src/interceptor.dart#L192-L206).

```dart
class Interceptor {
  /// The constructor only helps sub-classes to inherit from.
  /// Do not use it directly.
  const Interceptor();

  /// Called when the request is about to be sent.
  void onRequest(
    RequestOptions options,
  ) {
    handler.next(options);
  }

  /// Called when the response is about to be resolved.
  void onResponse(
    Response response,
    ResponseInterceptorHandler handler,
  ) {
    handler.next(response);
  }

  /// Called when an exception was occurred during the request.
  void onError(
    DioException err,
    ErrorInterceptorHandler handler,
  ) {
    handler.next(err);
  }
}
```

### DIFF
--- a/dio/lib/src/interceptor.dart
+++ b/dio/lib/src/interceptor.dart
@@ -220,7 +220,7 @@ typedef InterceptorSuccessCallback = void Function(
 
 /// The signature of [Interceptor.onError].
 typedef InterceptorErrorCallback = void Function(
-  DioException err,
+  DioException error,
   ErrorInterceptorHandler handler,
 );
 
@@ -371,10 +371,10 @@ class QueuedInterceptor extends Interceptor {
   }
 
   void _handleError(
-    DioException err,
+    DioException error,
     ErrorInterceptorHandler handler,
   ) {
-    _handleQueue(_errorQueue, err, handler, onError);
+    _handleQueue(_errorQueue, error, handler, onError);
   }
 
   void _handleQueue<T, V extends _BaseHandler>(
--- a/dio/lib/src/interceptor.dart
+++ b/dio/lib/src/interceptor.dart
@@ -199,10 +199,10 @@ class Interceptor {
 
   /// Called when an exception was occurred during the request.
   void onError(
-    DioException err,
+    DioException error,
     ErrorInterceptorHandler handler,
   ) {
-    handler.next(err);
+    handler.next(error);
   }
 }
 
@@ -220,7 +220,7 @@ typedef InterceptorSuccessCallback = void Function(
 
 /// The signature of [Interceptor.onError].
 typedef InterceptorErrorCallback = void Function(
-  DioException err,
+  DioException error,
   ErrorInterceptorHandler handler,
 );
 
@@ -255,13 +255,13 @@ mixin _InterceptorWrapperMixin on Interceptor {
 
   @override
   void onError(
-    DioException err,
+    DioException error,
     ErrorInterceptorHandler handler,
   ) {
     if (_onError != null) {
-      _onError!(err, handler);
+      _onError!(error, handler);
     } else {
-      handler.next(err);
+      handler.next(error);
     }
   }
 }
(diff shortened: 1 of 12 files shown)


dio/lib/src/interceptor.dart
```diff
@@ -214,13 +214,13 @@ typedef InterceptorSendCallback = void Function(
...
 typedef InterceptorSuccessCallback = void Function(
-  Response<dynamic> e,
+  Response<dynamic> response,
   ResponseInterceptorHandler handler,
 );
 
 /// The signature of [Interceptor.onError].
 typedef InterceptorErrorCallback = void Function(
-  DioException e,
+  DioException err,
```
AlexV525, line 223: I'd prefer `error`.

AlexV525 (commented):
This would breaks downstream packages' CI since it'll raise warnings about the mismatch naming AFAIK. But it's easy to change the name. Any opinion? @cfug-dev

The PR still lacks of the CHANGELOG update.

AlexV525 (commented):
This would breaks downstream packages' CI since it'll raise warnings about the mismatch naming AFAIK. But it's easy to change the name. Any opinion? @cfug-dev

The PR still lacks of the CHANGELOG update.

dio/lib/src/interceptor.dart
hgraceb: Changing `e` to `err` is done to maintain consistency with the existing [Interceptor.onError](https://github.com/cfug/dio/blob/a7a787840b87def9d16c10c009c548ddfcd86337/dio/lib/src/interceptor.dart#L202) method. If `err` in [Interceptor.onError](https://github.com/cfug/dio/blob/a7a787840b87def9d16c10c009c548ddfcd86337/dio/lib/src/interceptor.dart#L202) were changed to `error`, it would cause complaints from downstream packages' CI or IDE.

```dart
/// Called when an exception was occurred during the request.
void onError(
  DioException err,
  ErrorInterceptorHandler handler,
) {
  handler.next(err);
}
```

hgraceb:
> This would breaks downstream packages' CI since it'll raise warnings about the mismatch naming AFAIK. But it's easy to change the name. Any opinion? @cfug-dev
> 
> The PR still lacks of the CHANGELOG update.

Based on my knowledge, the main purpose of parameter naming in aliases is to facilitate code completion in IDEs. It doesn't have any practical impact on other code. Matching names are only required when overriding methods, so I haven't updated the CHANGELOG. 

Alternatively, should I update my understanding?

dio/lib/src/interceptor.dart
AlexV525: Thanks for the explanation! In this case I think it's better to be `error` and we can maybe consider update the callback in the next breaking version.

AlexV525:
> Based on my knowledge, the main purpose of parameter naming in aliases is to facilitate code completion in IDEs. It doesn't have any practical impact on other code. Matching names are only required when overriding methods, so I haven't updated the CHANGELOG.

This makes sense. Thanks.

dio/lib/src/interceptor.dart
hgraceb: That's great. I'm going to try changing all instances of `err` to `error`. If it looks good, I can recommit the commit based on the 6.0.0 branch.

dio/lib/src/interceptor.dart
```diff
@@ -199,10 +199,10 @@ class Interceptor {
 
   /// Called when an exception was occurred during the request.
   void onError(
-    DioException err,
+    DioException error,
```
AlexV525, line 202: Ah did I say it wrong? I mean we only change the type alias's parameter name (as you already did), but not changing the callback's.

hgraceb: > This makes sense.

Oh, I didn't see that sentence just now and I resubmitted the changes. I thought you still wanted to change `err` to `error`, but only on the 6.0.0 branch. If it's okay to do it the way it was before, I can revert the previous commit.

hgraceb: I redid the commit, hoping it's like this. Otherwise, I might have to speak in Chinese😁.

AlexV525 (approved):

PR #2095, opened 2024-01-11
✨ `Dio.clone`

AlexV525:
Similar to `copyWith`, `Dio.clone` would be a handy method to let users create new instances with local or grouping usages.


kuhnroyal:
Independent of this PR, what do you think of making the Dio instance completely immutable in version 6?

AlexV525:
> Independent of this PR, what do you think of making the Dio instance completely immutable in version 6?

Yeah I like the idea, and I'm also exploring the path. The thing is Dio's structure makes pieces nested, we cannot fluently use Dio in cases of grouping usage. So going immutable might be a better path to improve the overall usability and readability.

PR #2101, opened 2024-01-21
[http2] Fix Http2Adapter not redirecting for relative URLs

kuhnroyal:
Fixes a redirect bug discovered during https://github.com/cfug/dio/pull/2077 where the adapter would fail to properly redirect if the location header contains a relative URL.

### DIFF
--- a/plugins/http2_adapter/lib/src/http2_adapter.dart
+++ b/plugins/http2_adapter/lib/src/http2_adapter.dart
@@ -196,11 +196,34 @@ class Http2Adapter implements HttpClientAdapter {
     // Handle redirection.
     if (needRedirect) {
       final url = responseHeaders.value('location');
+      if (url == null) {
+        throw DioException.connectionError(
+          requestOptions: options,
+          reason: 'Received redirect without location header.',
+        );
+      }
+      final uri = Uri.parse(url);
       redirects.add(
-        RedirectRecord(statusCode, options.method, Uri.parse(url ?? '')),
+        RedirectRecord(
+          statusCode,
+          options.method,
+          uri,
+        ),
       );
+      final String path;
+      if (uri.hasScheme) {
+        /// This is a full URL which has to be redirected to as is.
+        path = uri.toString();
+      } else {
+        /// This is relative with or without leading slash and is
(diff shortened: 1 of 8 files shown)


plugins/http2_adapter/lib/src/http2_adapter.dart
```diff
@@ -196,11 +196,34 @@ class Http2Adapter implements HttpClientAdapter {
     // Handle redirection.
     if (needRedirect) {
       final url = responseHeaders.value('location');
+      if (url == null) {
```
AlexV525, line 199: Do we have a RFC for the behavior? I can come out with another idea that we might hanfle this as a self-redirection. Someone else could thought the same thing.

plugins/http2_adapter/test/http2_test.dart
AlexV525, line 1: Can we also test the empty location?

plugins/http2_adapter/lib/src/http2_adapter.dart
kuhnroyal: No, that just didn't make sense to me. Needs to be changed back.

plugins/http2_adapter/test/http2_test.dart
kuhnroyal: I'll try, probably not against httpbun.

kuhnroyal: No way to test this against httpbun. Might have to extract this into some helper function to make it testable.

AlexV525:
@kuhnroyal Still WIP?

kuhnroyal:
Updated

plugins/http2_adapter/test/http2_test.dart
kuhnroyal: Added a util method and separate tests since empty location header is not supported by httpbun.

plugins/http2_adapter/lib/src/http2_adapter.dart
kuhnroyal: Fixed

AlexV525 (approved):
👍

Issue #1869, opened 2023-06-15
Dio is not catching SocketException: Failed host lookup


busslina:
### Package

dio

### Version

5.2.0+1

### Output of `flutter doctor -v`

```shell
[✓] Flutter (Channel stable, 3.10.0, on Microsoft Windows [Versi¢n 10.0.19045.3086], locale es-ES)
    • Flutter version 3.10.0 on channel stable at C:\xxxxxxxxxxxx\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 84a1e904f4 (5 weeks ago), 2023-05-09 07:41:44 -0700
    • Engine revision d44b5a94c9
    • Dart version 3.0.0
    • DevTools version 2.23.1

[✓] Windows Version (Installed version of Windows is version 10 or higher)

[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.3)
    • Android SDK at C:\Users\xxxxxxxx\AppData\Local\Android\Sdk
    • Platform android-33, build-tools 30.0.3
    • ANDROID_HOME = C:\Users\xxxxxxxxxxx\AppData\Local\Android\Sdk
    • ANDROID_SDK_ROOT = C:\Users\xxxxxxxxxx\AppData\Local\Android\Sdk
    • Java binary at: C:\Program Files\Android\Android Studio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)
    • All Android licenses accepted.

[✓] Chrome - develop for the web
    • Chrome at C:\Program Files (x86)\Google\Chrome\Application\chrome.exe

[✓] Visual Studio - develop for Windows (Visual Studio Community 2022 17.4.4)
    • Visual Studio at C:\Program Files\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.4.33213.308
    • Windows 10 SDK version 10.0.20348.0

[✓] Android Studio (version 2022.1)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 11.0.15+0-b2043.56-8887301)

[✓] IntelliJ IDEA Community Edition (version 2021.1)
[...]

PR #2104, opened 2024-01-24
Escape from uncaught exceptions for IDEs

EricssonXD:
I faced the same issue as #314 and #1869, and I also understand that it is not an error with the dio package but an error with the Flutter SDK / VScode, where VScode does not recognise the exceptions as handled when using Future.catchError(). 

Despite it not being a problem of this package, I believe "fixing" it such that it avoids the above mentioned problem is still the correct approach, the following modification avoids this problem, and allows the example Error Handling code on pub.dev works with VScode.

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -10,7 +10,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 - Fix cancellation for streamed responses and downloads when using `IOHttpClientAdapter`.
 - Fix receive progress for streamed responses and downloads when using `IOHttpClientAdapter`.
 - Support relative `baseUrl` on the Web platform.
-- Fix `Excpetions` getting treated as `Uncaught` during debugging
+- Avoid fake uncaught exceptions during debugging with IDEs.
 
 ## 5.4.0
 
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -10,6 +10,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 - Fix cancellation for streamed responses and downloads when using `IOHttpClientAdapter`.
 - Fix receive progress for streamed responses and downloads when using `IOHttpClientAdapter`.
 - Support relative `baseUrl` on the Web platform.
+- Fix `Excpetions` getting treated as `Uncaught` during debugging
 
 ## 5.4.0
 
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -470,11 +470,12 @@ abstract class DioMixin implements Dio {
         RequestInterceptorHandler handler,
       ) async {
         requestOptions = reqOpt;
-        await _dispatchRequest<T>(reqOpt)
-            .then((value) => handler.resolve(value, true))
-            .catchError((e) {
-          handler.reject(e as DioException, true);
-        });
+        try {
+          final value = await _dispatchRequest<T>(reqOpt);
+          handler.resolve(value, true);
+        } on DioException catch (e) {
+          handler.reject(e, true);
+        }
       }),
     );
 
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -468,9 +468,9 @@ abstract class DioMixin implements Dio {
       requestInterceptorWrapper((
         RequestOptions reqOpt,
         RequestInterceptorHandler handler,
-      ) {
+      ) async {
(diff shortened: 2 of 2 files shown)


AlexV525:
Could also help to identify if this is relevant fix for #1869?

EricssonXD:
> Could also help to identify if this is relevant fix for #1869?

I forgot to mention it in the description, this is also an issue that I came across which is the exact problem I ran into initially

EricssonXD:
> @EricssonXD Hi, thanks for taking the time to investigate the cause. Very much appreciated!
> 
> I was wonder if adding the paired `async/await` could be addressed by simply adding `return` to the `_dispatchRequest`. WDYT?

As the function is a parameter that gets passed to the interceptor, which expects a `void Function(RequestOptions options, RequestInterceptorHandler handler)`, it doesn't work. Unless we also modify the interceptor?

EricssonXD:
> @EricssonXD Hi, thanks for taking the time to investigate the cause. Very much appreciated!
> 
> I was wonder if adding the paired `async/await` could be addressed by simply adding `return` to the `_dispatchRequest`. WDYT?

Is there any reasons why you would like to avoid `async/await`? From what I see in how the code is used, I don't think there should be any difference in terms of performance. But I'm still new to Flutter, so correct me if I'm wrong.

AlexV525:
> Is there any reasons why you would like to avoid `async/await`? From what I see in how the code is used, I don't think there should be any difference in terms of performance. But I'm still new to Flutter, so correct me if I'm wrong.

You may spot that we've already implemented `.then` and `.catchError` for the method. Theoretically, all exceptions should be caught here. Awaiting a future that is not void is not recommended.

Consider writing like this:
image

EricssonXD:
> > Is there any reasons why you would like to avoid `async/await`? From what I see in how the code is used, I don't think there should be any difference in terms of performance. But I'm still new to Flutter, so correct me if I'm wrong.
> 
> You may spot that we've already implemented `.then` and `.catchError` for the method. Theoretically, all exceptions should be caught here. Awaiting a future that is not void is not recommended.
> 
> Consider writing like this: image

That's actually what I wrote initially during debugging but I found out adding only the await keyword somehow made it work which lead to my proposed code. I have updated my code now, but I don't quite understand what do you mean by 
>Awaiting a future that is not void is not recommended.
Isn't this new code also awaiting a future that is not void? Or do you mean like we shouldn't await a value that is not void if we are not using it?

AlexV525:
> > Awaiting a future that is not void is not recommended.
> 
> Isn't this new code also awaiting a future that is not void? Or do you mean like we shouldn't await a value that is not void if we are not using it?

Sorry, that's a typo. :(

Awaiting a void future is not recommended.

AlexV525:
> That's actually what I wrote initially during debugging but I found out adding only the await keyword somehow made it work which lead to my proposed code. I have updated my code now, but I don't quite understand what do you mean by
> 
> > Awaiting a future that is not void is not recommended.
> 
> Isn't this new code also awaiting a future that is not void? Or do you mean like we shouldn't await a value that is not void if we are not using it?

What I mean is we shouldn't wait for a future that returns value but not use it eventually.

In general:
```dart
Future<String> foo() async {
  await Future(() {});
  return 'bar';
}

void main() async {
  // Good.
  final result = await foo();
  print(result);

  // Bad.
  await foo();
}
```

EricssonXD:
Ahhh I see, are there any more problems that I should address in my pull request?

dio/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 - Fix cancellation for streamed responses and downloads when using `IOHttpClientAdapter`.
 - Fix receive progress for streamed responses and downloads when using `IOHttpClientAdapter`.
 - Support relative `baseUrl` on the Web platform.
+- Fix `Excpetions` getting treated as `Uncaught` during debugging
```
AlexV525, line 13:
```suggestion
- Avoid fake uncaught exceptions during debugging with IDEs.
```

AlexV525 (approved):
One nit:

AlexV525:
I'd consider the request as test exempt, as it doesn't affect any actual behavior. Waiting for other inputs.

AlexV525:
@kuhnroyal @ueman Any inputs?

AlexV525:
Going to merge this and apply inputs if any in the future.

kuhnroyal:
I think the fix for the root cause is out in Dart:master now. But this looks cleaner in any case.
We should try to avoid `then` & `catchError` in all new could just for readability.

PR #2127, opened 2024-03-07
🥅 Throw argument error with invalid `baseUrl`

AlexV525:
Provide a readable error when the `baseUrl` is not matching the condition.

### DIFF
--- a/dio/lib/src/options.dart
+++ b/dio/lib/src/options.dart
@@ -99,7 +99,13 @@ mixin OptionsMixin {
   late String _baseUrl;
 
   set baseUrl(String value) {
-    assert(value.isEmpty || kIsWeb || Uri.parse(value).host.isNotEmpty);
+    if (value.isNotEmpty && !kIsWeb && Uri.parse(value).host.isEmpty) {
+      throw ArgumentError.value(
+        value,
+        'baseUrl',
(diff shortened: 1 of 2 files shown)


dio/lib/src/options.dart
```diff
@@ -99,7 +99,13 @@ mixin OptionsMixin {
   late String _baseUrl;
 
   set baseUrl(String value) {
-    assert(value.isEmpty || kIsWeb || Uri.parse(value).host.isNotEmpty);
+    if (value.isNotEmpty && !kIsWeb && Uri.parse(value).host.isEmpty) {
+      throw ArgumentError.value(
+        value,
+        'baseUrl',
+        'Must has a valid host on platforms other than the Web',
```
kuhnroyal, line 106:
```suggestion
        'Must be a valid URL on platforms other than Web.',
```

AlexV525: I took other callers of the error as examples and they didn't seem to have the dot at the end of the message.

kuhnroyal: Ah ok, I don't mind either way. Only remembered that you always added a dot at the end :)

kuhnroyal (approved):

PR #2140, opened 2024-03-15
✅ Fix download stream tests

AlexV525:
The GitHub runner is much faster than our configurations, apparently.

### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [ ] I have added the required tests to prove the fix/feature I'm adding
- [ ] I have updated the documentation (if necessary)
- [x] I have run the tests without failures
- [ ] I have updated the `CHANGELOG.md` in the corresponding package

PR #2142, opened 2024-03-15
⚡️ Improve multiple timeout tests

AlexV525:
- Further try #2140, using `drip` to control the overall response duration.
- Make cancellation calls only after the response stream has been obtained to keep the call sequence.
- Using the non-routable URL (10.0.0.0) to mock connection timeouts.

### DIFF
--- a/dio_test/lib/src/test/timeout_tests.dart
+++ b/dio_test/lib/src/test/timeout_tests.dart
@@ -17,7 +17,7 @@ void timeoutTests(
 
   group('Timeout exception of', () {
     group('connectTimeout', () {
-      test('with response', () async {
+      test('throws', () async {
         dio.options.connectTimeout = Duration(milliseconds: 3);
         await expectLater(
           dio.get(nonRoutableUrl),


kuhnroyal:
Using drip is actually a great idea, didn't think of that!

AlexV525:
I've updated the description, this might significantly help with random test failures.

dio_test/lib/src/test/timeout_tests.dart
```diff
@@ -18,7 +20,7 @@ void timeoutTests(
       test('with response', () async {
         dio.options.connectTimeout = Duration(milliseconds: 3);
         await expectLater(
-          dio.get('/'),
```
kuhnroyal, line 21: Not sure about this, there is not gonna be a response so this actually tests something different now.

dio_test/lib/src/test/timeout_tests.dart
```diff
@@ -98,7 +100,7 @@ void timeoutTests(
     );
     dio.options.connectTimeout = Duration(milliseconds: 10);
     await expectLater(
-      dio.get('/drip-lines?delay=1'),
```
kuhnroyal, line 101: Same here.

kuhnroyal (commented):
Wonder if we should test the connection timeouts with mocks?

AlexV525:
> Wonder if we should test the connection timeouts with mocks?

Okay so it's not a mock, it happens because the IP is non-reachable in a regular LAN network. We should use the same URL (or something like 192.168.0.0) if we want to make sure that the connection timeouts statically happened.
Regarding the name of tests, they might also changed to non-response-related.

dio_test/lib/src/test/timeout_tests.dart
AlexV525: The name got a minor update.

kuhnroyal (approved):

PR #2151, opened 2024-03-19
Run test workflow on pull_request_target event

kuhnroyal:
This should allow us to use actions in that create comments etc.
The workflows are being run on. the code from the main branch.


kuhnroyal:
My understanding is that the checks are no running because the workflow changed from `pull_request` to `pull_request_target` - But they should be running for any PR pointing to main after this is merged.

AlexV525 (approved):

PR #2173, opened 2024-04-03
🔥 Do not detaches sockets / terminates client stream

AlexV525:
Fixes #2170

### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [x] I have added the required tests to prove the fix/feature I'm adding
- [x] I have updated the documentation (if necessary)
- [x] I have run the tests without failures
- [x] I have updated the `CHANGELOG.md` in the corresponding package

PR #2176, opened 2024-04-03
Bump min SDK to Dart 2.18.0 / Flutter 3.3.0 & fix CI

kuhnroyal:
- Bump min SDK to Dart 2.18.0 / Flutter 3.3.0 where we can use a newer melos version (3.4.0) that allows simplified usage and supports `pubspec_overrides.yaml` files
- revert workflow event back to `pull_request` & create coverage comment in a separate workflow
- remove unnecessary `exclusive` parameter
- adapt download test to use httpbun instead of downloading artifacts from Github


kuhnroyal:
Well this doesn't work

kuhnroyal:
Looks like something is cached

kuhnroyal:
Ok, so since we switched to `pull_request_target` a lot of the repository content is actually used from main. Probably when we use path or git dependencies.

AlexV525:
> Ok, so since we switched to `pull_request_target` a lot of the repository content is actually used from main. Probably when we use path or git dependencies.

Ah that sounds fair. Maybe `GITHUB.ref` would help?

kuhnroyal:
So after reverting to `pull_request` we still have 2 issues.
1. The download test is taking too long (longer than 2 minutes)
2. The melos version on min SDK (or the Dart version itself) doesn't seem to support the pubspec overrides

kuhnroyal:
Dart 2.18.0 was released on Aug 30, 2022 - so I think this is fine, we should probably publish this as a minor version.

kuhnroyal:
The build still shows as failure but this is due to the old `pull_request_target` usage.

kuhnroyal:
@AlexV525 I think this is good to go, the coverage comment should start to work again once the new workflow is on the main branch. Hopefully it also has correct coverage values after this. This was likely caused by the same problem.

AlexV525:
I'm on a holiday and will be AFK for 3 days. I'm considering a couple things:
1. I'd like to fix #2173 first so we can get as much fixes as we could before we bump SDK requirements.
2. I'd also like if we split the `exclusive` fix as a single request so we can easily manage the SDK upgrade request.
3. We might also stated somewhere that we can at least support SDKs that are released within 1 year in order to provide a solid service.

kuhnroyal:
Have a nice holiday!

I understand your concerns but as it stands now, the min Dart SDK 2.15.0 does not support `pubspec_overrides.yaml`.
So all the testing setup we did recently does not actually work for the min SDK since we removed the path dependencies to the test project.

All changes in this PR are interwoven and atomically won't solve the problems. 
If we change the `exclusive` problem, there would still be wrong tests pointing to the `main` branch in the next PR. It might show a success while the actual code does not work, just how the `exclusive` problem happened in the first place.

Dart 2.15 ist almost 2.5 years old, we need to get rid of it so we can move forward with the limited time we all have to spare for OS projects. If someone is still using such an old version and having actual problems due to specific bugs, then they can afford to pay someone else to solve the problem in a fork or donate to dio and request a custom handmade release.

AlexV525:
Yes I agree with all changes, those concerns are something I think we might do when applying changes. As our regular principle about the maintaining, we can provide a great guide or message to allow users to have a good chance to stand along with our opinions.

AlexV525:
I've added the compatibility policy doc to indicate our general support and exceptions.

kuhnroyal:
Nice, I would go with ~1.5 years instead of 2 years.
We have Dart 3.0.0 on the `6.0.0` branch which is less than 1 year ago. We might be able to go back down to `2.19.0` and then we could release this around mid 2024.

AlexV525:
> Nice, I would go with ~1.5 years instead of 2 years.

Yeah it's a general idea, so it depends, which means you could always do so if you have a proper reason.

> We have Dart 3.0.0 on the `6.0.0` branch which is less than 1 year ago. We might be able to go back down to `2.19.0` and then we could release this around mid 2024.

Sounds good

AlexV525:
@ueman Do you have any input about this request and relevant policies?

AlexV525 (approved):

PR #2205, opened 2024-04-29
Export MediaType from http_parser

### LINKED ISSUE #2212: I can't use MediaType when I create MultipartFile
### Package

dio

### Version

5.4.3+1

### Operating-System

Android, iOS, Web, MacOS, Linux, Windows

### Adapter

Default Dio

```shell
!] Flutter (Channel stable, 3.19.0, on macOS 14.4.1 23E224 darwin-arm64, locale pt-BR)
    • Flutter version 3.19.0 on channel stable at /Users/usuario/fvm/versions/3.19.0
    ! Warning: `dart` on your path resolves to /opt/homebrew/Cellar/dart/3.2.0/libexec/bin/dart, which is not inside your current Flutter SDK checkout at /Users/usuario/fvm/versions/3.19.0. Consider adding /Users/usuario/fvm/versions/3.19.0/bin to the front of your path.
    • Framework revision bae5e49bc2 (3 months ago), 2024-02-13 17:46:18 -0800
    • Engine revision 04817c99c9
    • Dart version 3.3.0
    • DevTools version 2.31.1
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git" directly to perform update checks and upgrades.

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at /Users/usuario/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0
    • ANDROID_HOME = /Users/usuario/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)

[✓] Xcode - develop for iOS and macOS (Xcode 15.3)
    • Build 15E204a
    • CocoaPods version 1.15.2

[✓] Android Studio (version 2022.3)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-17.0.6b829.9-10027231)

[✓] VS Code (version 1.87.2)
    • VS Code at /Users/usuario/apps/Visual Studio Code.app/Contents
    • Flutter extension version 3.89.20240501
```

### Dart Version

_No response_

### Steps to Reproduce

When using MultipartFile.fromFile does it ask for a contentType parameter that is of type MediaType? but I can't pass the parameter because it is an http_parser parameter but dio does not export the http_parser

### Expected Result

Future<FormData?> _convertForFormDataDio(
      GenericFormData genericFormData) async {
    final FormData formData = FormData();

    for (String mediaPath in genericFormData.filePaths) {
      formData.files.add(
        MapEntry<String, MultipartFile>(
          'files',
          await MultipartFile.fromFile(
            mediaPath,
            filename: path.basename(mediaPath),
            contentType: MediaType.parse('.png') ,
            headers: <String, List<String>>{
              'Content-Type': <String>['multipart/form-data'],
            },
          ),
        ),
      );
    }

    for (String key in genericFormData.fields.keys) {
      formData.fields.add(MapEntry<String, String>(
          key, genericFormData.fields[key].toString()));
    }

    return formData;
  }

### Actual Result

Future<FormData?> _convertForFormDataDio(
      GenericFormData genericFormData) async {
    final FormData formData = FormData();

    for (String mediaPath in genericFormData.filePaths) {
      formData.files.add(
        MapEntry<String, MultipartFile>(
          'files',
          await MultipartFile.fromFile(
            mediaPath,
            filename: path.basename(mediaPath),
            contentType: MediaType.parse('.png') ,//Undefined name 'MediaType'. Try correcting the name to one that is defined, or defining the name.
            headers: <String, List<String>>{
              'Content-Type': <String>['multipart/form-data'],
            },
          ),
        ),
      );
    }

    for (String key in genericFormData.fields.keys) {
      formData.fields.add(MapEntry<String, String>(
          key, genericFormData.fields[key].toString()));
    }

    return formData;
  }

Reprevise:
Resolves #2212

This PR exports `MediaType` from the http_parser package so that way users don't have to import a (probably) transitive dependency.


ueman:
What would be the use case here? Can you give an example? 

I'm tempted to decline this PR, since exporting other libraries is rather bad form. Especially if it's just exporting one class of it. If you need http_parser code, you should depend on it. For dio, it's pretty much an implementation detail, and exporting it makes it no longer an implementation detail, and thus harder for us to evolve related things.

kuhnroyal:
Yea, I don't think we use this class as parameter anywhere. If you want to use this for parsing mime/media types you should import this in your package.

AlexV525:
Well, we accept `MediaType` as content-type in the multipart file.

https://github.com/cfug/dio/blob/87e6b1d4c8f9d3c57dd2291f540b730b510f4b20/dio/lib/src/multipart_file.dart#L45

But I agree that users should import the package explicitly.

Reprevise:
I 100% disagree. Why should users have to add a dependency which is an implementation detail of dio just to use the API? Exporting types from other packages is a very normal thing to do and is common in popular packages.

AlexV525:
I'm trying to find best practices for re-exporting packages from the package but failed to find something. Given that we have previous arguments about "best practice", I'm reopening this for further discussions.

dio/lib/dio.dart
```diff
@@ -5,6 +5,8 @@
 /// {@category Plugins}
 library dio;
 
+export 'package:http_parser/http_parser.dart' show MediaType;
```
AlexV525, line 8: However, exporting the `MediaType` from here seems a bad part for me. I'd prefer to export it from the `multipart_file.dart`.

kuhnroyal:
Well, let's do it.

kuhnroyal (approved):

ueman:
Should we document somewhere that we're re-exporting these classes, in case people will run into name conflicts? Is it even possible to run into problems?

kuhnroyal:
What if we use a type alias with a Dio prefix and re-export that? It will still accept the original MediaType type so it should not break anything.

AlexV525:
> What if we use a type alias with a Dio prefix and re-export that? It will still accept the original MediaType type so it should not break anything.

That would be a great idea. @Reprevise Could you make that attempt?

Reprevise:
I could do:
```dart
typedef DioMediaType = MediaType;
```
Though I prefer the exporting approach so that users would just get an [`unnecessary_import`](https://dart.dev/tools/diagnostic-messages#unnecessary_import) lint and its easier to migrate than having to change `MediaType` to `DioMediaType`.

kuhnroyal:
They should not get any warning, the `MediaType` should still be an accepted argument.

Reprevise:
> They should not get any warning, the `MediaType` should still be an accepted argument.

Not saying they should get a warning, but I just mean if they want to get rid of the http_parser import and the explicit dependency of it in pubspec.yaml.

kuhnroyal:
Those who care about removing the explicit dependency are free to change from `MediaType` to `DioMediaType`.
Everyone else is not effected.
And we prevent compiler warnings for duplicate types/exports.

Reprevise:
I have updated to use the type alias. The reason I had the constructors take `DioMediaType` is to make it less confusing for newer users. Users who see MediaType will try and use it and it won't be available to them without first depending on `http_parser` whereas if they use `DioMediaType` it will instantly be available.

kuhnroyal:
For me it looks good, can you maybe add a test which ensure that we can still pass the `MediaType` as argument? Or is that overkill?

Reprevise:
One thing that I didn't consider is that when hovering over the constructor, it doesn't actually say the alias, it just says the type.
image
So even though I changed the constructor to take a `DioMediaType`, you'd only know that if you go to the actual declaration.

dio/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 
 - Raise the min Dart SDK version to 2.18.0.
 - Add constructor for `DioExceptionType.badCertificate`.
+- Create type alias DioMediaType for http_parser's MediaType
```
AlexV525, line 10: We'd like to keep the code as `code`, so
```suggestion
- Create type alias `DioMediaType` for `http_parser`'s `MediaType`.
```

dio/lib/src/multipart_file.dart
```diff
@@ -7,6 +7,8 @@ import 'multipart_file/io_multipart_file.dart'
     if (dart.library.html) 'multipart_file/browser_multipart_file.dart';
 import 'utils.dart';
 
+typedef DioMediaType = MediaType;
```
AlexV525, line 10: Also, we don't want to reduce our docs coverage during changes, so maybe:
```suggestion
/// The type (alias) for specifying the content-type of the `MultipartFile`. 
typedef DioMediaType = MediaType;
```

AlexV525:
@Reprevise Could you solve the format issue?

AlexV525:
> > @Reprevise Could you solve the format issue?
> 
> What format issue?

https://github.com/cfug/dio/actions/runs/9175832587/job/25229693166?pr=2205

PR #2215, opened 2024-05-15
Adapt dart2wasm

huanghui1998hhh:


dio/pubspec.yaml
```diff
@@ -24,6 +24,7 @@ dependencies:
   http_parser: ^4.0.0
   meta: ^1.5.0
   path: ^1.8.0
+  web: ^0.5.1
```
kuhnroyal, line 27: What is the min SDK requirement to use the web package?

dio/lib/src/adapters/browser_adapter.dart
```diff
@@ -135,16 +140,18 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
...
 
       final onSendProgress = options.onSendProgress;
       if (onSendProgress != null) {
-        xhr.upload.onProgress.listen((event) {
-          if (event.loaded != null && event.total != null) {
-            onSendProgress(event.loaded!, event.total!);
-          }
-        });
+        void onSendProgressWrap(web.ProgressEvent event) {
+          onSendProgress(event.loaded, event.total);
```
kuhnroyal, line 151: The progress loaded and total events are now non-nullable?

dio/lib/src/adapters/browser_adapter.dart
```diff
@@ -111,15 +114,17 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
...
     if (requestStream != null) {
       if (connectTimeoutTimer != null) {
-        xhr.upload.onProgress.listen((event) {
+        void cancelConnectTimeoutTimer(web.ProgressEvent _) {
           connectTimeoutTimer?.cancel();
           connectTimeoutTimer = null;
-        });
+        }
+
+        xhr.upload.onprogress = cancelConnectTimeoutTimer.toJS;
```
kuhnroyal, line 122: Is there a way to pass these functions anonymously?

kuhnroyal (commented):
Thanks!
Left some questions, I am generally not sure about min SDK support for this and if the tests work. We might need some test for the new header function.

kuhnroyal:
Looks like we have to move to Dart 3.1.0 for this.

AlexV525:
What about we split the Web support as a package and publish it first, then make a major version upgrade for wasm?

dio/pubspec.yaml
kuhnroyal: Seems to be 3.1.0

kuhnroyal:
Could work but that would also imply a breaking change for the main package.

AlexV525:
> Could work but that would also imply a breaking change for the main package.

As I imagine it won't, we can set the constraint to allow both previous and the new version

dio/lib/src/adapters/browser_adapter.dart
huanghui1998hhh: Yes, we can write like `xhr.upload.onprogress = (web.ProgressEvent event){}.toJS` this.

dio/lib/src/adapters/browser_adapter.dart
huanghui1998hhh: > The progress loaded and total events are now non-nullable?

Yes, they are non-nullable in `package:web`.

dio/pubspec.yaml
huanghui1998hhh: > Seems to be 3.1.0

Maybe 3.3.0?

kuhnroyal:
We have to re-export some of the web classes in dio.

AlexV525:
Offline discussed with @huanghui1998hhh , he will go ahead to try split the package. As I & he has tested the current package does not affect the `flutter build`, only affects the runtime.

dio/lib/src/adapter.dart
```diff
@@ -4,7 +4,8 @@ import 'dart:typed_data';
 import 'package:meta/meta.dart';
 
 import 'adapters/io_adapter.dart'
-    if (dart.library.html) 'adapters/browser_adapter.dart' as adapter;
+    if (dart.library.js_util) 'adapters/web_adapters/web_adapter.dart'
```
Rexios80, line 7: This should be `dart.library.js_interop` as mentioned here: https://dart.dev/interop/js-interop/package-web#conditional-imports

and everywhere else in the dio package

AlexV525:
Closing in favor of #2223

PR #2218, opened 2024-05-17
Split the web support to package dio_browser_adapter

huanghui1998hhh:
https://github.com/cfug/dio/issues/2215#issuecomment-2113653554
Prepare to adapt WASM.


dio/lib/browser.dart
```diff
@@ -1,2 +1 @@
-export 'src/adapters/browser_adapter.dart' show BrowserHttpClientAdapter;
```
kuhnroyal, line 1: Do we need to re-export the new adapter here, ideally with the same name? Otherwise we have a breaking change.

dio/lib/src/adapter.dart
```diff
@@ -4,7 +4,7 @@ import 'dart:typed_data';
 import 'package:meta/meta.dart';
 
 import 'adapters/io_adapter.dart'
-    if (dart.library.html) 'adapters/browser_adapter.dart' as adapter;
```
kuhnroyal, line 7: Does this require a certain Dart version?

AlexV525: Tested locally (with 3.16) and it seems working

AlexV525:
Oops, `pub get` seems always implicitly run with the `example/` too...

huanghui1998hhh:
A tough problem : (

dio/lib/src/adapter.dart
```diff
@@ -4,6 +4,7 @@ import 'dart:typed_data';
 import 'package:meta/meta.dart';
 
 import 'adapters/io_adapter.dart'
+    if (dart.library.js_util) 'adapters/browser_adapter.dart'
     if (dart.library.html) 'adapters/browser_adapter.dart' as adapter;
```
Rexios80, line 8: This line should be removed. ALL references to `dart:html` and `dart.library.html` should be removed.

dio/lib/src/adapter.dart
```diff
@@ -4,6 +4,7 @@ import 'dart:typed_data';
 import 'package:meta/meta.dart';
 
 import 'adapters/io_adapter.dart'
+    if (dart.library.js_util) 'adapters/browser_adapter.dart'
```
Rexios80, line 7: This should be `dart.library.js_interop` as mentioned here: https://dart.dev/interop/js-interop/package-web#conditional-imports

dio/lib/src/compute/compute.dart
```diff
@@ -25,7 +25,9 @@
 
 import 'dart:async';
 
-import 'compute_io.dart' if (dart.library.html) 'compute_web.dart' as _c;
+import 'compute_io.dart'
+    if (dart.library.js_util) 'compute_web.dart'
```
Rexios80, line 29: Same here

dio/lib/src/dio.dart
```diff
@@ -3,6 +3,7 @@ import 'dart:async';
 import 'adapter.dart';
 import 'cancel_token.dart';
 import 'dio/dio_for_native.dart'
+    if (dart.library.js_util) 'dio/dio_for_browser.dart'
```
Rexios80, line 6: Same here

dio/lib/src/dio_mixin.dart
```diff
@@ -15,6 +15,7 @@ import 'headers.dart';
 import 'interceptors/imply_content_type.dart';
 import 'options.dart';
 import 'progress_stream/io_progress_stream.dart'
+    if (dart.library.js_util) 'progress_stream/browser_progress_stream.dart'
```
Rexios80, line 18: Same here

dio/lib/src/multipart_file.dart
```diff
@@ -4,6 +4,7 @@ import 'dart:convert';
 import 'package:http_parser/http_parser.dart';
 
 import 'multipart_file/io_multipart_file.dart'
+    if (dart.library.js_util) 'multipart_file/browser_multipart_file.dart'
```
Rexios80, line 7: Same

dio/lib/src/adapter.dart
Rexios80: Actually removing these might require upping the min Dart version to 3.3 which isn't ideal

AlexV525: > This line should be removed. ALL references to `dart:html` and `dart.library.html` should be removed.

Will this cause WASM to degrade or something related?

Rexios80: I'm not sure. All I know is the migration docs say to remove them.

Rexios80: Since this is an import condition it shouldn't cause any issues, but the js_util ones definitely need changed to js_interop as js_util is one of the deprecated web packages

kuhnroyal: I think the imports are resolved at compile time, so it should be fine. But definitely should switch to the new `js_interop` package. We need to figure out if it is possible to run our adapter test in WASM, that will tell us that it is working

AlexV525:
Closing in favor of #2223

PR #2230, opened 2024-06-02
Add support for `download` function in web

mbfakourii:

### Additional context and info

In this PR, download support was added to dio with the help of `Blob` files.

Note: There is no need to send `savePath` on the web and it can be empty.


kuhnroyal:
This looks interesting, thanks! I read about `createObjectUrlFromBlob` just the other day and thought the same.

We will need some tests and and likely an example.

dio/lib/src/dio/dio_for_browser.dart
```diff
@@ -29,9 +27,29 @@ class DioForBrowser with DioMixin implements Dio {
     String lengthHeader = Headers.contentLengthHeader,
     Object? data,
     Options? options,
-  }) {
-    throw UnsupportedError(
-      'The download method is not available in the Web environment.',
+  }) async {
+    options ??= DioMixin.checkOptions('GET', options);
+
+    // Set receiveTimeout to 48 hours because `Duration.zero` not work!
```
kuhnroyal, line 33: What happened?

kuhnroyal (commented):
Probably missing some error handling.

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: The download was canceled after 3 seconds. I had to use this method!

dio/lib/src/dio/dio_for_browser.dart
```diff
@@ -1,10 +1,8 @@
+import 'dart:async';
+import 'dart:html';
+
+import '../../dio.dart';
```
AlexV525, line 4: Do not import the export library.

dio/lib/src/dio/dio_for_browser.dart
```diff
@@ -29,9 +27,31 @@ class DioForBrowser with DioMixin implements Dio {
     String lengthHeader = Headers.contentLengthHeader,
     Object? data,
     Options? options,
-  }) {
-    throw UnsupportedError(
-      'The download method is not available in the Web environment.',
+  }) async {
+    options ??= DioMixin.checkOptions('GET', options);
+
+    options = options.copyWith(responseType: ResponseType.bytes);
```
AlexV525, line 33: So the download would wait until all bytes got saved, which will likely cause high memory usage and long waiting, right?

dio/lib/src/dio/dio_for_browser.dart
```diff
@@ -29,9 +27,31 @@ class DioForBrowser with DioMixin implements Dio {
...
+
+    // Set receiveTimeout to 48 hours because `Duration.zero` not work!
+    options = options.copyWith(receiveTimeout: const Duration(hours: 48));
```
AlexV525, line 36: This doesn't look right.

dio/lib/src/dio/dio_for_browser.dart
```diff
@@ -29,9 +27,31 @@ class DioForBrowser with DioMixin implements Dio {
...
+
+    final completer = Completer<Response>();
+
+    // Create blob url from byte data
+    response.data = Url.createObjectUrlFromBlob(Blob([response.data]));
+
+    // Set response in Completer
+    completer.complete(response);
+
+    return DioMixin.listenCancelForAsyncTask(cancelToken, completer.future);
```
AlexV525, line 55: Why do we need extra handling with the CancelToken? Isn't it already handled in the adapter?

AlexV525 (changes requested):
Please write up a general idea of what behavior/standard it follows.

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: Yes, it should wait until it is finished, but it can be checked with onReceiveProgress

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: Every time the download started, it timed out for 3 seconds and the download was canceled. I had to use this method.

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: I did this based on the [`dio_for_native`](https://github.com/cfug/dio/blob/6e5a96b957c302bc1010edb78b0f64d204dea303/dio/lib/src/dio/dio_for_native.dart#L188) file.

dio/lib/src/dio/dio_for_browser.dart
mbfakourii:
```dart
import 'dart:async';
import 'dart:html';
```

These packages are needed to create a blob file !

AlexV525: I mean `../../dio.dart`.

dio/lib/src/dio/dio_for_browser.dart
AlexV525: What is the root cause of the behavior? Does it caused by the implementation of Dart or Flutter?

dio/lib/src/dio/dio_for_browser.dart
AlexV525: It uses the explicitly handling because there are subscriptions that need to be cancelled which seems not related with the browser implementation.

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: @AlexV525 

I am trying to do this with stream, but the data is not returned through stream! Is there a problem with stream on the web side?

dio/lib/src/dio/dio_for_browser.dart
ueman: Please explain why it's not working in a comment. Stating that it doesn't work isn't an actually helpful comment, since the next person working on it will need to figure out why it's not working again.

mbfakourii: I don't know exactly why this happens, but I get the following error

```dart
*** DioException ***:
uri: https://jsoncompare.org/LearningContainer/SampleFiles/Video/MP4/sample-mp4-file.mp4
DioException [receive timeout]: The request took longer than 0:00:03.000000 to receive data. It was aborted. To get rid of this exception, try raising the RequestOptions.receiveTimeout above the duration of 0:00:03.000000 or improve the response time of the server.
```

mbfakourii:
@kuhnroyal @ueman @AlexV525 

By further examining the web request, I realized that we use [`arraybuffer`](https://github.com/cfug/dio/blob/f87a09566c076936be236e3313f8843d6d60564c/dio/lib/src/adapters/browser_adapter.dart#L43) when requesting the web. `XMLHttpRequest` supports [blob](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType#value). I made a change in the implementation

The problem of high memory is solved.

thank you for checking again

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: changed

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: changed

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: What is your suggestion?

dio/lib/src/options.dart
```diff
@@ -630,6 +633,11 @@ class RequestOptions extends _RequestConfig with OptionsMixin {
 
   /// {@macro dio.options.ProgressCallback}
   ProgressCallback? onSendProgress;
+
+  /// Get blob url
+  ///
+  /// Only work in web
+  bool blobUrl;
```
AlexV525, line 640: Is that the only way to enable the Blob feature for the request on the Web? Could we use pattern matching to determine a Blob URL like `Uri.parse(input).scheme == 'blob'`?

dio/lib/src/dio/dio_for_browser.dart
AlexV525: The `completer` in the `dio_for_native.dart` handles data and errors which cover all cases, in which the current implementation does not look the same. So they should be at least the same behavior.

dio/lib/src/adapters/browser_adapter.dart
```diff
@@ -67,18 +67,34 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
...
-          statusMessage: xhr.statusText,
-          isRedirect: xhr.status == 302 ||
-              xhr.status == 301 ||
-              options.uri.toString() != xhr.responseUrl,
-        ),
-      );
+      if (options.blobUrl) {
+        completer.complete(
+          ResponseBody.fromString(
+            Url.createObjectUrl(xhr.response),
```
AlexV525, line 73: Any chance that this could fail? Maybe another `catch` here would be good.

AlexV525 (commented):
Still need some tests and examples here... 🛎️

dio/lib/src/options.dart
mbfakourii: `BaseOptions` does [not support](https://github.com/cfug/dio/blob/4dec2237d14ecd9b27a5718e720c114ffacd81a9/dio/lib/src/options.dart#L143) Uri, it only takes string !

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: What should I use instead of  `return DioMixin.listenCancelForAsyncTask(cancelToken, completer.future);`?

dio/lib/src/options.dart
ueman: It should be safe to parse to a URI though.

On a related note: This would probably a good API change for the next stable version. URLs should be URIs, not Strings

mbfakourii: I tested with `requestUri`, first it adds the address of the word `blob` and the download encounters an error.

Download address instead
`https://example.com/1.mp4`
It changes to this
`blob:https://example.com/1.mp4`

dio/lib/src/adapters/browser_adapter.dart
mbfakourii: I tested and did not encounter exception !

AlexV525:
Could you move on with the new structure of the code and add tests?

mbfakourii:
> Could you move on with the new structure of the code and add tests?

The changes have been made. You may check again

dio/lib/src/options.dart
```diff
@@ -41,6 +41,10 @@ enum ResponseType {
 
   /// Get the original bytes, the [Response.data] will be [List<int>].
   bytes,
+
+  /// Get blob url.
+  /// Only work in web.
```
AlexV525, line 46: The description seems to lack explanation.

plugins/web_adapter/lib/src/html/adapter.dart
```diff
@@ -65,18 +66,35 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
...
-          statusMessage: xhr.statusText,
-          isRedirect: xhr.status == 302 ||
-              xhr.status == 301 ||
-              options.uri.toString() != xhr.responseUrl,
-        ),
-      );
+      if (options.responseType == ResponseType.blobUrl) {
+        completer.complete(
+          ResponseBody.fromString(
+            Url.createObjectUrl(xhr.response),
```
AlexV525, line 72: Could you reuse other fields by constructing the `ResponseBody` with the unnamed constructor?

dio/lib/src/dio/dio_for_browser.dart
AlexV525: It is not about the `listenCancelForAsyncTask`, the completer does not handle any errors which it should.

https://github.com/cfug/dio/blob/47abf18387e89983f010771e19e6d643f5d043b6/dio/lib/src/dio/dio_for_native.dart#L133-L183

AlexV525 (commented):
Tests are still lacking.

plugins/web_adapter/lib/src/html/adapter.dart
mbfakourii: The `ResponseBody` with the unnamed constructor requires steam, we have a url string !

dio/lib/src/options.dart
mbfakourii: I made changes to the description

dio/lib/src/dio/dio_for_browser.dart
mbfakourii: Changes were made

mbfakourii:
Changes were made

AlexV525:
Please add tests.

mbfakourii:
> Please add tests.

What kind of test is needed? And in what form should it be?

mbfakourii:
> Please add tests.

I tried to add the test, but I get the following error

```dart
PS C:\Users\aaa\bbb\dio\plugins\web_adapter> flutter test
Error on line 10, column 3 of dart_test.yaml: Unknown platform "chrome".
   ╷
10 │   chrome:
   │   ^^^^^^
   ╵
```

or

```dart
PS C:\Users\aaa\bbb\dio\plugins\web_adapter\test> flutter test .\browser_test.dart
Changing current working directory to: C:\Users\mbfakourii\Projects\dio\plugins\web_adapter
Error on line 10, column 3 of dart_test.yaml: Unknown platform "chrome".
   ╷
10 │   chrome:
   │   ^^^^^^
   ╵
```

AlexV525:
> I tried to add the test, but I get the following error

Try to set up the test like the GitHub workflow of this library.

mbfakourii:
> Please add tests.

Test added

plugins/web_adapter/lib/src/dio_impl.dart
```diff
@@ -25,9 +27,24 @@ class DioForBrowser with DioMixin implements Dio {
...
     Options? options,
-  }) {
-    throw UnsupportedError(
-      'The download method is not available in the Web environment.',
+  }) async {
+    final Response response = await fetch(
+      RequestOptions(
+        baseUrl: urlPath,
+        data: data,
+        method: 'GET',
```
AlexV525, line 35: Why always `GET`?

plugins/web_adapter/lib/src/dio_impl.dart
AlexV525, line 1: What will the result be after downloaded? Could you demonstrate how it works?

plugins/web_adapter/lib/src/dio_impl.dart
mbfakourii: fixed

plugins/web_adapter/lib/src/dio_impl.dart
```diff
@@ -25,9 +27,24 @@ class DioForBrowser with DioMixin implements Dio {
...
+        onReceiveProgress: onReceiveProgress,
+      ),
     );
+
+    final completer = Completer<Response>();
+
+    // Set response in Completer
+    completer.complete(response);
+
+    return completer.future;
```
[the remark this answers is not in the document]
mbfakourii: removed.

plugins/web_adapter/lib/src/dio_impl.dart
mbfakourii: I added an example in example_flutter_app

mbfakourii: To test on the local web, make sure CORS is [disabled](https://pub.dev/packages/flutter_cors#usage).

and use local `dio` and `dio_web_adapter` in example_flutter_app pubspec
```dart
dependency_overrides:
  dio_web_adapter:
    path: ...\dio\plugins\web_adapter
  dio:
    path: ...\dio\dio
```

plugins/web_adapter/test/browser_test.dart
```diff
@@ -18,4 +18,18 @@ void main() {
     browserAdapter.fetch(opts, testStream, cancelFuture);
     expect(browserAdapter.xhrs.every((e) => e.withCredentials == true), isTrue);
   });
+
+  test('ResponseType in blobUrl', () async {
+    final browserAdapter = BrowserHttpClientAdapter(withCredentials: true);
+    final opts = RequestOptions(responseType : ResponseType.blobUrl);
+    final testStream = Stream<Uint8List>.periodic(
+      const Duration(seconds: 1),
+          (x) => Uint8List(x),
```
AlexV525, line 27: Seems not well formatted.

example_flutter_app/lib/download_web/download_blob.dart
AlexV525, line 1: It might be better to place the example in the Dart example instead of Flutter.

plugins/web_adapter/lib/src/dio_impl.dart
AlexV525: Why CORS is required?

mbfakourii: To download a file from an online server to a local server, the browser gives a CORS error. This has nothing to do with DIO. I said this for testing.

AlexV525: Hmmm, how about a local server to scope with the minimal functionality instead of an outer source?

plugins/web_adapter/test/browser_test.dart
mbfakourii: fixed.

example_flutter_app/lib/download_web/download_blob.dart
mbfakourii: This example is for web support, how do we implement it in pure Dart?

AlexV525: > This example is for web support, how do we implement it in pure Dart?

All underlying packages are pure Dart packages.

https://dart.dev/web/get-started

plugins/web_adapter/lib/src/dio_impl.dart
mbfakourii: Do you mean to put a file locally? If we do this, the download display will be very fast and the download percentage will not be clearly visible.

Are you sure about this? This is just an example to demonstrate a feature

AlexV525: A demo would not require displaying the progress, correct? The file could be as small as possible too.

mbfakourii: I think there are some examples for pure Dart
https://github.com/cfug/dio/blob/0b0f527029bd36b65536e1f61b688284edb2594e/dio_test/lib/src/test/download_tests.dart#L94

Just need to run it on the web.

mbfakourii: @AlexV525

PR #2239, opened 2024-06-08
add Utf8JsonTransformer

### LINKED ISSUE #2238: Opportunity to improve the json decoding performance
### Request Statement

Hi there!

I experimented a bit with the hidden `_JsonUtf8Decoder` which is created when [fusing the `utf8` decoder with the `json` decoder](https://github.com/dart-lang/sdk/blob/5b2ea0c7a227d91c691d2ff8cbbeb5f7f86afdb9/sdk/lib/_internal/vm/lib/convert_patch.dart#L40).

This significantly improves the decoding performance.

I created a benchmark [here](https://github.com/knaeckeKami/dio_benchmark).

It shows an improvement of around ~10x in AOT mode on sufficiently large json responses. (around 64 KiB).
 
Here's the improved Transformer:

```dart
import 'dart:convert';
import 'dart:typed_data';

class UTF8JsonTransformer extends SyncTransformer {
  /// on VM, this creates a special fast decoder.
  /// see https://github.com/dart-lang/sdk/blob/5b2ea0c7a227d91c691d2ff8cbbeb5f7f86afdb9/sdk/lib/_internal/vm/lib/convert_patch.dart#L54
  final decoder = const Utf8Decoder().fuse(const JsonDecoder());

  UTF8JsonTransformer();

  @override
  Future<dynamic> transformResponse(
    RequestOptions options,
    ResponseBody responseBody,
  ) async {
    final responseType = options.responseType;
    // Do not handled the body for streams.
    if (responseType == ResponseType.stream) {
      return responseBody;
    }

    // Return the finalized bytes if the response type is bytes.
    if (responseType == ResponseType.bytes) {
      final chunks = await responseBody.stream.toList();
      final responseBytes =
          Uint8List.fromList(chunks.expand((c) => c).toList());
      return responseBytes;
    }

    final isJsonContent = Transformer.isJsonMimeType(
      responseBody.headers[Headers.contentTypeHeader]?.first,
    );

    if (isJsonContent) {
      final stream = responseBody.stream;
      final decodedStream = decoder.bind(stream);
      final decoded = await decodedStream.toList();
      assert(decoded.length == 1);
      return decoded.first;
    }

    // If the response is not JSON, we should return the response as a string

    final chunks = await responseBody.stream.toList();
    final responseBytes = Uint8List.fromList(chunks.expand((c) => c).toList());

    return utf8.decode(responseBytes, allowMalformed: true);
  }
}
```

Note: you need a recent version of Dart to see these performance improvements.

This would be a nice improvement for Dio. Who doesn't like free improved performance?

A caveat is, that this is not compatible with the default `BackgroundTransformer`,  as the callback in this transformer takes a String, not a Uint8List. 
To get the benefits there, a change in the API would be required (or we could add a new `BackgroundTransformer` class). 

Would you be open to a PR that integrates these changes in the existing SyncTransformer? 

### Solution Brainstorm

_No response_

knaeckeKami:

fixes #2238 

This PR adds the `Utf8JsonTransformer`.

I made sure that it is as close to `BackgroundTransformer` as possible.
The only behavior change is that is checks to Content-Length header for deciding whether to switch to a background isolate for decoding or not.
If the Content-Lenght header is not set or not valid, it will fall back to counting the bytes in the response.

This is because it's slightly faster to feed to the response stream directly into the decoder without consolidating all the response bytes into a single Uint8List first. In case the server does not send a valid Content-Length header, the Transformer will check the actual bytes from the response, as the `BackgroundTransformer` does.

The remaining question is if we should set the Transformer as default. 
IMO it would make sense to change the default transformer to

```dart
  /// The default [Transformer] that transfers requests and responses
  /// into corresponding content to send.
  @override
  Transformer transformer = Utf8JsonTransformer(contentLengthIsolateThreshold: 50 * 1024);
```

so all users get the improved performance by default.


ueman:
https://github.com/knaeckeKami/dio_benchmark

dio/lib/src/transformer.dart
```diff
@@ -3,6 +3,7 @@ import 'package:http_parser/http_parser.dart';
 
 import 'adapter.dart';
 import 'options.dart';
+import 'transformers/sync_transformer.dart';
```
AlexV525, line 6: Is this required?

dio/lib/src/transformers/utf8_json_transformer.dart
```diff
@@ -0,0 +1,184 @@
+import 'dart:convert';
+import 'dart:typed_data';
+
+import 'package:dio/dio.dart';
+import 'package:dio/src/compute/compute.dart';
```
AlexV525, line 5: We'd prefer relatively import here

dio/lib/src/transformers/utf8_json_transformer.dart
```diff
@@ -0,0 +1,184 @@
...
+
+/// A [Transformer] that has a fast path for decoding utf8-encoded JSON.
```
AlexV525, line 7: For sentences I think `utf8` should be `UTF-8`, so utf8-encoded can be `encoded in UTF-8`?

dio/lib/src/transformers/utf8_json_transformer.dart
```diff
@@ -0,0 +1,184 @@
...
+/// If the response is utf8-encoded JSON and no custom decoder for a Request is specified, this transformer
```
AlexV525, line 8: What is the `Request` referencing here?

dio/lib/src/transformers/utf8_json_transformer.dart
```diff
@@ -0,0 +1,184 @@
...
+/// is significantly faster than the default [SyncTransformer] or [BackgroundTransformer].
```
AlexV525, line 9: or -> and the?

dio/lib/src/transformers/utf8_json_transformer.dart
```diff
@@ -0,0 +1,184 @@
...
+/// This improvement is achieved by using a fused [Utf8Decoder] and [JsonDecoder] to decode the response,
+/// which is faster than decoding the utf8-encoded JSON in two separate steps, since
+/// Dart uses a special fast decoder for this case.
+/// See https://github.com/dart-lang/sdk/blob/5b2ea0c7a227d91c691d2ff8cbbeb5f7f86afdb9/sdk/lib/_internal/vm/lib/convert_patch.dart#L40
+///
+/// By default, this transformer will transform responses in the main isolate,
+/// but a custom threshold can be set to switch to an isolate for large responses by passing
+/// [contentLengthIsolateThreshold].
+class Utf8JsonTransformer extends Transformer {
```
AlexV525, line 18: How about `FusedTransformer`?

dio/lib/src/transformers/utf8_json_transformer.dart
```diff
@@ -0,0 +1,184 @@
...
+      responseBytes = await _consolidateStream(responseBody.stream);
+      contentLength = responseBytes.length;
+    } else {
+      contentLength = int.parse(contentLengthHeader.first);
+    }
+
+    final shouldUseIsolate = !(contentLengthIsolateThreshold < 0) &&
+        contentLength >= contentLengthIsolateThreshold;
+    if (shouldUseIsolate) {
+      // we can't send the stream to the isolate, so we need to decode the response bytes first
```
AlexV525, line 128: Would that lead to a memory issue compare to the previous transformer?

dio/lib/src/transformers/utf8_json_transformer.dart
```diff
@@ -0,0 +1,184 @@
...
+      return compute(
+        _decodeUtf8ToJson,
+        responseBytes ?? await _consolidateStream(responseBody.stream),
+      );
+    } else {
+      if (!hasContentLengthHeader || contentLength == 0) {
+        responseBytes ??= await _consolidateStream(responseBody.stream);
+        // if the response is empty, return null, since the decoder would throw
```
AlexV525, line 136: Could you explain more on this case with examples? Also comments should start with capital letter and end with dot too:

```suggestion
        // Returns `null` to let the decoder throws when the response is empty.
```

dio/lib/src/transformer.dart
knaeckeKami: ah, it's for the JsonEncodeCallback for static `defaultTransformRequest` method.
Do you want me to extract the typedefs to another file?

dio/lib/src/transformers/utf8_json_transformer.dart
knaeckeKami: no, the previous transformers always decode into a single UInt8List first.
SyncTransformer:
```dart
    final chunks = await responseBody.stream.toList();
    final responseBytes = Uint8List.fromList(chunks.expand((c) => c).toList());
```

dio/lib/src/transformer.dart
AlexV525: Yeah probably to this file

dio/lib/src/transformers/utf8_json_transformer.dart
knaeckeKami: I added a more detailed explanation as comment.
This branch is for backwards compatibility with the other Transformers.

The old behavior was, that if a server responds with an `Content-Type: application/json` header. but the response body was empty, the transformer would return null. 
But the fused decoder would throw an exception in this case.

So, if we don't have a valid Content-Length header, we first check if the body is empty before converting the bytes to json.
If we do have a Content-Length header which is >0, we can feed the response bytes directly into the decoder, which is the most performant path.

dio/lib/src/transformers/utf8_json_transformer.dart
knaeckeKami: fine by me, changed the name

dio/lib/src/transformers/utf8_json_transformer.dart
knaeckeKami: applied changes

dio/lib/src/transformers/utf8_json_transformer.dart
knaeckeKami: the RequestOptions. If a custom decoder is set in to RequestOptions, we respect that, but in this case we cannot use the fused decoder. 
I changed the comments to clarify that.

dio/lib/src/transformers/utf8_json_transformer.dart
knaeckeKami: yes, applied your suggestion

dio/lib/src/transformers/utf8_json_transformer.dart
knaeckeKami: changed the imports

dio/lib/src/transformer.dart
knaeckeKami: done

Reprevise:
Can you add a `jsonDecodeCallback` parameter to this new transformer? I have a long running isolate who's purpose is just to decode JSON whereas the internal compute function spawns a new isolate per request.

ueman:
> Can you add a `jsonDecodeCallback` parameter to this new transformer? I have a long running isolate who's purpose is just to decode JSON whereas the internal compute function spawns a new isolate per request.

That doesn't really make sense for this specific transformer as far as I can tell. It's an interesting idea though.

knaeckeKami:
We could allow to pass in a jsonDecodeCallback, but it would need to operate an on a Uint8List, not a String like the backgroundtransformer in order to leverage the fused decoder.

However, we should measure if that even makes sense in current versions of Dart.
Spawning a new isolate got a lot cheaper recently, and an advantage of using `compute()` is that is can use `Isolate.exit` to return the decoded json to the calling isolate, which avoids a copy.

see https://api.dart.dev/stable/3.4.4/dart-isolate/Isolate/exit.html
> If the port is a native port -- one provided by [ReceivePort.sendPort](https://api.dart.dev/stable/3.4.4/dart-isolate/ReceivePort/sendPort.html) or [RawReceivePort.sendPort](https://api.dart.dev/stable/3.4.4/dart-isolate/RawReceivePort/sendPort.html) -- the system may be able to send this final message more efficiently than normal port communication between live isolates. In these cases this final message object graph will be reassigned to the receiving isolate without copying. Further, the receiving isolate will in most cases be able to receive the message in constant time.

So the question is, is spawning a new Isolate more expensive than copying the decoded map back into the main isolate?
I don't know, we could benchmark it.

knaeckeKami:
FWIW, I found that a large portion of the performance improvement actually comes from the more efficient consolidating of the `Stream<Uint8List> `to a single `Uint8List`

The `SyncTransformer` and the extending `BackgroundTransformer` do 

`Uint8List.fromList(chunks.expand((c) => c).toList());` which copies every byte sequentially, probably also creating wrapper objects for each byte.

Using a `ByteBuffer` is an optimization that could also be integrated into `SyncTransformer` without breaking changes.

Reprevise:
> So the question is, is spawning a new Isolate more expensive than copying the decoded map back into the main isolate?
I don't know, we could benchmark it.

I'm just interested in the best performance possible. In my project, the size cutoff is 50KB. Any larger than that and it gets sent to the long-running Isolate for parsing. If we find spawning many isolates is better, then I'll concede.

Reprevise:
> FWIW, I found that a large portion of the performance improvement actually comes from the more efficient consolidating of the Stream<Uint8List> to a single Uint8List

Sounds like a free win. Maybe a separate PR is in order? I think this PR should move forward as well though, I think the performance benefits of using the fused decoder is worth it.

knaeckeKami:
> Sounds like a free win. Maybe a separate PR is in order? I think this PR should move forward as well though, I think the performance benefits of using the fused decoder is worth it.

Yeah, when this is merged I'll prepare a new one, that extracts the _consolidateStream function and make SyncTransformer use it as well.

knaeckeKami:
> I'm just interested in the best performance possible. In my project, the size cutoff is 50KB. Any larger than that and it gets sent to the long-running Isolate for parsing. If we find spawning many isolates is better, then I'll concede.

The question is, for which aspect do we want to optimize? 
We get the least amount of CPU usage overall with decoding the JSON in the main isolate because no data needs to be copied between isolates and no new isolates need to be spawned.

The least amount of CPU usage on the main isolate is likely achieved by spawning a new isolate each time, as the spawning of the new isolate happens on another thread, and the decoded JSON does not need to be copied back in the main isolate.

With a long-running Isolate, the question is first consolidating the byte stream on the main isolate, then copying it to another isolate and then copying the decoded JSON map back into the main isolate is even less CPU intensive than just decoding it in the main isolate in the first place.

But taking the https://github.com/knaeckeKami/dio_benchmark repo and adapting it to show how a long running isolate performs might be a good idea.

ueman (approved):
LGTM. This is awesome. Thank you!

AlexV525:
> The remaining question is if we should set the Transformer as default. IMO it would make sense to change the default transformer to
> 
> ```dart
>   /// The default [Transformer] that transfers requests and responses
>   /// into corresponding content to send.
>   @override
>   Transformer transformer = Utf8JsonTransformer(contentLengthIsolateThreshold: 50 * 1024);
> ```
> 
> so all users get the improved performance by default.

We are not doing this currently, right? IMO`Content-Length` might not be specified in regular requests so consolidating bytes will still be its place. If there are no drawbacks then we definitely can move on with the new one.

ueman:
Excactly, that's not yet done

knaeckeKami:
> We are not doing this currently, right? IMOContent-Length might not be specified in regular requests so consolidating bytes will still be its place. If there are no drawbacks then we definitely can move on with the new one.

There are no drawbacks compared to the old Transformer, as the old transformer does the same unconditionally.
Also, I think in practice pretty much any server that returns JSON also sets Content-Length anyway.

PR #2248, opened 2024-06-15
backport improved byte consolidation to SyncTransformer and BackgroundTranformer

knaeckeKami:
Follow up to #2239:

The more efficient Stream<UintList> -> Uint8List conversion can be backported to `SyncTransformer` & the extending `BackgroundTransformer`

### DIFF
--- a/dio/lib/src/transformers/util/consolidate_bytes.dart
+++ b/dio/lib/src/transformers/util/consolidate_bytes.dart
@@ -1,0 +1,12 @@
+import 'dart:typed_data';
+
+/// Consolidates a stream of [Uint8List] into a single [Uint8List]
+Future<Uint8List> consolidateBytes(Stream<Uint8List> stream) async {
+  final builder = BytesBuilder(copy: false);
+
+  await for (final chunk in stream) {
+    builder.add(chunk);
+  }
+
+  return builder.takeBytes();
+}
--- a/dio/lib/src/transformers/fused_transformer.dart
+++ b/dio/lib/src/transformers/fused_transformer.dart
@@ -6,6 +6,7 @@ import '../compute/compute.dart';
 import '../headers.dart';
 import '../options.dart';
 import '../transformer.dart';
+import 'util/consolidate_bytes.dart';
 
 /// A [Transformer] that has a fast path for decoding UTF8-encoded JSON.
 /// If the response is utf8-encoded JSON and no custom decoder is specified in the [RequestOptions], this transformer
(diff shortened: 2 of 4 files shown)


knaeckeKami:
the failing CI is very likely a flake, due to request to https://wrong.host.badssl.com/ timing out

ueman (approved):

dio/lib/src/transformers/util/consolidate_bytes.dart
```diff
@@ -0,0 +1,12 @@
+import 'dart:typed_data';
+
+/// Consolidates a stream of [Uint8List] into a single [Uint8List]
+Future<Uint8List> consolidateBytes(Stream<Uint8List> stream) async {
```
ekuleshov, line 4: isn't this the same as `collectBytes(Stream<List<int> stream)` in the `async` package that dio already uses?

AlexV525: They don't share the same arguments, no?

AlexV525: cc @knaeckeKami

knaeckeKami: An Uint8List implements List<int>, so I guess it would work. 
The implementation also looks very similar. 


```dart

Future<Uint8List> collectBytes(Stream<List<int>> source) {
  return _collectBytes(source, (_, result) => result);
}

/// Generalization over [collectBytes] and [collectBytesCancelable].
///
/// Performs all the same operations, but the final result is created
/// by the [result] function, which has access to the stream subscription
/// so it can cancel the operation.
T _collectBytes<T>(Stream<List<int>> source,
    T Function(StreamSubscription<List<int>>, Future<Uint8List>) result) {
  var bytes = BytesBuilder(copy: false);
  var completer = Completer<Uint8List>.sync();
  var subscription =
      source.listen(bytes.add, onError: completer.completeError, onDone: () {
    completer.complete(bytes.takeBytes());
  }, cancelOnError: true);
  return result(subscription, completer.future);
}
```

vs.

```dart
/// Consolidates a stream of [Uint8List] into a single [Uint8List]
Future<Uint8List> consolidateBytes(Stream<Uint8List> stream) async {
  final builder = BytesBuilder(copy: false);

  await for (final chunk in stream) {
    builder.add(chunk);
  }

  return builder.takeBytes();
}
```

we could probably swap it out for the one in dart:async. I don't have the resources to open a PR and test it right now though.

PR #2250, opened 2024-06-17
Set FusedTransformer as default

knaeckeKami:

Also fix three oversights in the transformer

- Fix parsing content as JSON when Content-Type is like `application/json` but `ResponseType == .plain`

- HEAD requests have an empty body, even when Content-Length is >0. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD . I did not take this into account. This is also fixed now.

- Respect custom decoders returning null

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -11,8 +11,8 @@ See the [Migration Guide][] for the complete breaking changes list.**
 - Fix the type conversion regression when using `MultipartFile.fromBytes`.
 - Split the Web implementation to `package:dio_web_adapter`.
 - Add FusedTransformer for improved performance when decoding JSON.
-- Improves `InterceptorState.toString()`.
 - Set FusedTransformer as the default transformer.
+- Improves `InterceptorState.toString()`.
 
 ## 5.4.3+1
 
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -12,6 +12,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 - Split the Web implementation to `package:dio_web_adapter`.
 - Add FusedTransformer for improved performance when decoding JSON.
 - Improves `InterceptorState.toString()`.
+- Set FusedTransformer as the default transformer.
 
 ## 5.4.3+1
 
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -40,9 +40,13 @@ abstract class DioMixin implements Dio {
 
   /// The default [Transformer] that transfers requests and responses
   /// into corresponding content to send.
+  /// For response bodies greater than 50KB, a new Isolate will be spawned to
+  /// decode the response body to JSON.
+  /// Taken from https://github.com/flutter/flutter/blob/135454af32477f815a7525073027a3ff9eff1bfd/packages/flutter/lib/src/services/asset_bundle.dart#L87-L93
+  /// 50 KB of data should take 2-3 ms to parse on a Moto G4, and about 400 μs
+  /// on a Pixel 4.
   @override
-  Transformer transformer =
-      FusedTransformer(contentLengthIsolateThreshold: 50 * 1024);
+  Transformer transformer = FusedTransformer(contentLengthIsolateThreshold: 50 * 1024);
 
   bool _closed = false;
 
@@ -432,8 +436,7 @@ abstract class DioMixin implements Dio {
 
         // The request has already been cancelled,
         // there is no need to listen for another cancellation.
-        if (state.data is DioException &&
-            state.data.type == DioExceptionType.cancel) {
+        if (state.data is DioException && state.data.type == DioExceptionType.cancel) {
           return handleError();
         }
         if (state.type == InterceptorResultType.next ||
@@ -455,9 +458,8 @@ abstract class DioMixin implements Dio {
 
     // Add request interceptors into the request flow.
     for (final interceptor in interceptors) {
-      final fun = interceptor is QueuedInterceptor
-          ? interceptor._handleRequest
-          : interceptor.onRequest;
+      final fun =
+          interceptor is QueuedInterceptor ? interceptor._handleRequest : interceptor.onRequest;
       future = future.then(requestInterceptorWrapper(fun));
     }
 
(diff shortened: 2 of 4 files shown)


dio/lib/src/dio_mixin.dart
```diff
@@ -41,7 +41,8 @@ abstract class DioMixin implements Dio {
   /// The default [Transformer] that transfers requests and responses
   /// into corresponding content to send.
   @override
-  Transformer transformer = BackgroundTransformer();
+  Transformer transformer =
+      FusedTransformer(contentLengthIsolateThreshold: 50 * 1024);
```
ueman, line 45: Can you add a comment explaining this specific number? Is it chosen rather arbitrarily or an educated guess?

ueman (commented):
Thanks again! I have no objections to changing the default transformer, but I'd like to hear the opinions on it from the others

dio/lib/src/dio_mixin.dart
ueman: Does it make sense to be configurable?

knaeckeKami: 50 * 1024 is the hardcoded default in BackgroundTransformer, I chose the same to keep the behavior as close as possible to the old one. 
I did not measure how long it takes to decode on smartphones.

IMO it makes sense to make it configurable, for example, when running the whole HTTP client in a background isolate, users might not care about blocking the isolate for a few milliseconds, so they could set a higher (or -1 to disable spawning new isolates completely).

dio/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 - Split the Web implementation to `package:dio_web_adapter`.
 - Add FusedTransformer for improved performance when decoding JSON.
 - Improves `InterceptorState.toString()`.
+- Set FusedTransformer as the default transformer.
```
kuhnroyal, line 15: Can you combine this with the previous FusedTransformer entry? This would make it easier to understand.

knaeckeKami: yes, done

dio/lib/src/transformers/fused_transformer.dart
```diff
@@ -88,9 +89,12 @@ class FusedTransformer extends Transformer {
     if (isJsonContent && decodedResponse != null) {
       // slow path decoder, since there was a custom decoder specified
       return jsonDecode(decodedResponse);
-    } else if (decodedResponse != null) {
+    } else if (customResponseDecoder != null) {
```
AlexV525, line 92: Is this making a difference compared to the previous? I'm seeing more readability if it's `decodedResponse != null`.

dio/lib/src/transformers/fused_transformer.dart
```diff
@@ -88,9 +89,12 @@ class FusedTransformer extends Transformer {
...
       return decodedResponse;
     } else {
+      if (responseBytes.isEmpty) {
```
AlexV525, line 95: Any references?

dio/lib/src/transformers/fused_transformer.dart
```diff
@@ -111,6 +118,10 @@ class FusedTransformer extends Transformer {
     // of the response or the length of the eagerly decoded response bytes
     final int contentLength;
 
+    // Successful HEAD requests don't have a response body, even if the content-length header
+    // present.
+    final contentLengthHeaderImpliesContent = options.method != 'HEAD';
```
AlexV525, line 123: - With `CONNECT` too although it's an edge case?
- Also the MDN reference would be helpful to be placed here.
- Could the name be `neverHasResponseBody` and then flip all conditions to the opposite?

```dart
final neverHasResponseBody = options.method == 'CONNECT' || options.method == 'HEAD';

if (neverHasResponseBody || !hasContentLengthHeader) {
  // ...
}
```

dio/test/transformer_test.dart
```diff
@@ -254,6 +269,24 @@ void main() {
...
+          Stream.value(Uint8List(0)),
+          200,
+          headers: {
+            Headers.contentTypeHeader: ['application/json'],
+            Headers.contentLengthHeader: ['123'],
+          },
+        ),
+      );
+      expect(response, null);
+    });
```
AlexV525, line 289: Please put a trailing comma

dio/lib/src/transformers/fused_transformer.dart
knaeckeKami: The issue only occurs when  the `ResponseType` is `.json`, the response body has a Content-Type of `application/json`, and a Content-Length header with a value > 0, but no response body.

A valid `CONNECT` response will have no Content-Type and no Content-Length header, so the transformer won't try to parse it as json.

> A server MUST NOT send any Transfer-Encoding or [Content-Length](https://www.rfc-editor.org/rfc/rfc9110#field.content-length) header fields in a [2xx (Successful)](https://www.rfc-editor.org/rfc/rfc9110#status.2xx) response to CONNECT. A client MUST ignore any Content-Length or Transfer-Encoding header fields received in a successful response to CONNECT.

I mean sure, we could also support invalid responses to CONNECT, but I'd argue if a user want's that, they should just set the `ResponseType` to `.plain, .bytes or .stream.` or use their own decoder / transformer.

dio/lib/src/transformers/fused_transformer.dart
knaeckeKami: Sorry, did not mean to commit that. Is unnecessary, utf8.decode returns an empty string anyway on an empty bytes list.

dio/test/transformer_test.dart
knaeckeKami: done

dio/lib/src/transformers/fused_transformer.dart
AlexV525: Cool, thanks for the explanation. But the flag should be flipped because if you are using its opposite result always you should simply flip it.

knaeckeKami: true. done

dio/lib/src/dio_mixin.dart
```diff
@@ -40,8 +40,14 @@ abstract class DioMixin implements Dio {
...
   /// into corresponding content to send.
+  /// For response bodies greater than 50KB, a new Isolate will be spawned to
+  /// decode the response body to JSON.
+  /// Taken from https://github.com/flutter/flutter/blob/135454af32477f815a7525073027a3ff9eff1bfd/packages/flutter/lib/src/services/asset_bundle.dart#L87-L93
+  /// 50 KB of data should take 2-3 ms to parse on a Moto G4, and about 400 μs
+  /// on a Pixel 4.
   @override
-  Transformer transformer = BackgroundTransformer();
+  Transformer transformer =
+      FusedTransformer(contentLengthIsolateThreshold: 50 * 1024);
```
AlexV525, line 50: We're likely to add the `require_trailing_commas` rule in the near future, so let's follow it as much as possible

```suggestion
  Transformer transformer = FusedTransformer(
    contentLengthIsolateThreshold: 50 * 1024，
  );
```

dio/lib/src/transformers/fused_transformer.dart
knaeckeKami: Yes, tests failed after I changed the default transformer because of that.

Custom decoders can return null.
Since there are tests that rely on this, there also might be users.

dio/lib/src/dio_mixin.dart
knaeckeKami: Sure, I like `require_trailing_commas`! It seems your suggestion contains full-width `，` instead of a `,` (fixed it in my last commit)

AlexV525: Ah it must be the creation of my input method :D

AlexV525 (approved):
LGTM

ueman (approved):

PR #2276, opened 2024-07-30
default-target-platform

brunogabriel:
#COMMENTS
* Remove Platform avoiding Fii import in web

- [ ]  I have added the required tests to prove the fix/feature I'm adding

Using defaultTargetPlatform to avoid fii imports in web projects.

### DIFF
--- a/plugins/native_dio_adapter/lib/src/native_adapter.dart
+++ b/plugins/native_dio_adapter/lib/src/native_adapter.dart
@@ -1,9 +1,7 @@
-import 'dart:io';
-import 'dart:typed_data';
-
 import 'package:cronet_http/cronet_http.dart';
 import 'package:cupertino_http/cupertino_http.dart';
 import 'package:dio/dio.dart';
+import 'package:flutter/foundation.dart';
 import 'cronet_adapter.dart';
 import 'cupertino_adapter.dart';
 
@@ -32,11 +30,12 @@ class NativeAdapter implements HttpClientAdapter {
     )
     URLSessionConfiguration? cupertinoConfiguration,
   }) {
-    if (Platform.isAndroid) {
+    if (defaultTargetPlatform == TargetPlatform.android) {
       _adapter = CronetAdapter(
         createCronetEngine?.call() ?? androidCronetEngine,
       );
-    } else if (Platform.isIOS || Platform.isMacOS) {
+    } else if (defaultTargetPlatform == TargetPlatform.iOS ||
+        defaultTargetPlatform == TargetPlatform.macOS) {
       _adapter = CupertinoAdapter(
         createCupertinoConfiguration?.call() ??
             cupertinoConfiguration ??


kuhnroyal:
@ueman The native adapter already depends on Flutter doesn't it?
@brunogabriel Can you provide a little bit more context? You should never have the native adapter in any web builds.

brunogabriel:
> @ueman The native adapter already depends on Flutter doesn't it? @brunogabriel Can you provide a little bit more context? You should never have the native adapter in any web builds.

Hi, @ueman  thanks for ask, yeah, I believe that NativeAdapter is not used for web apps, but I was testing my implementation ran by accident on flutter web and application suddenly broken by imports (fii). My idea here is to avoid this scenario.

ueman:
> @ueman The native adapter already depends on Flutter doesn't it?

True, I missed that.

> defaultTargetPlatform is a get (final), right?

Kinda, it can be overriden via `debugDefaultTargetPlatformOverride`.

> My idea here is to avoid this scenario.

`defaultTargetPlatform` defaults to `TargetPlatform.android`, as stated in the docs:

> In a test environment, the platform returned is [TargetPlatform.android](https://api.flutter.dev/flutter/foundation/TargetPlatform.html) regardless of the host platform.

kuhnroyal:
Thanks for the context!
From my understanding, `TargetPlatform` is mostly used for theming etc. and we should not rely on it for implementation details. I think you should be able to fix this by adjusting your adapter instantiation.

Alternatively we could think about using https://pub.dev/packages/platform - not sure how established this is.

brunogabriel:
> > @ueman The native adapter already depends on Flutter doesn't it?
> 
> True, I missed that.
> 
> > defaultTargetPlatform is a get (final), right?
> 
> Kinda, it can be overriden via `debugDefaultTargetPlatformOverride`.
> 
> > My idea here is to avoid this scenario.
> 
> `defaultTargetPlatform` defaults to `TargetPlatform.android`, as stated in the docs:
> 
> > In a test environment, the platform returned is [TargetPlatform.android](https://api.flutter.dev/flutter/foundation/TargetPlatform.html) regardless of the host platform.

do you suggest another approach? i know that this code isn't for web in this adapter, but the goal is to avoid this situation.

brunogabriel:
> Thanks for the context!
> From my understanding, `TargetPlatform` is mostly used for theming etc. and we should not rely on it for implementation details. I think you should be able to fix this by adjusting your adapter instantiation.
> 
> Alternatively we could think about using https://pub.dev/packages/platform - not sure how established this is.

yeah, understood, i got this behavior by accident in multiOS app, in my DI i can provide using simple chec. I'll read this lib tks

brunogabriel:
> Thanks for the context!
> From my understanding, `TargetPlatform` is mostly used for theming etc. and we should not rely on it for implementation details. I think you should be able to fix this by adjusting your adapter instantiation.
> 
> Alternatively we could think about using https://pub.dev/packages/platform - not sure how established this is.

another approach, i thought it's create a wrapper.

ueman:
Honestly, I wouldn't change anything and reject this PR. This package is not intended to be run on web and it also states so on pub.dev. The decision to use the native adapter should be done in the code of the consumer of this package.

kuhnroyal:
I agree.
@brunogabriel You should be able to extract your Dio configuration to 2 separate files based on the platform and then use a conditional import similar to https://github.com/cfug/dio/blob/main/dio/lib/src/adapter.dart#L6

PR #2281, opened 2024-08-14
Add ability to change [FileMode] of the download  by introducing [Dio…

shehabmohamed0:
…FileMode]

- [Yes] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [Yes] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [Yes] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [Yes] I have added the required tests to prove the fix/feature I'm adding
- [Yes] I have updated the documentation (if necessary)
- [Yes] I have run the tests without failures
- [No] I have updated the `CHANGELOG.md` in the corresponding package

Add the ability to change the [FileMode] of the download without altering the default value, which is [FileMode.write], to make resuming downloads easier to implement by introducing [DioFileMode] and mapping it to FileMode in [DioForNative]
without breaking web_adapter


shehabmohamed0:
> Can you please explain how this change enables resuming downloads? Please also add tests for both cases, so it will not get broken in the future.

When downloading a file a stream is opened and a Random Access File are opened in the client side to store bytes 
the current implementation to open a Random Access File is [FileMode.write] this mode creates a new file and overwrite the existing one if it already exists

changing the file access mode to [FileMode.append] or [FileMode.writeOnlyAppend] with change the behavior to write to the end of the file 


So if you had already started downloading a file and for some reason the download stopped let's say a network problem and you set [deleteOnError = false], the file had already downloaded 50% and it's not deleted it exists but is not complete 

You check if the file exists and get it's size then you request to download the file again by requesting a partial content setting the range header to file size, that means skip downloading the 50% you already downloaded then you append the remaining 50% to the end of the file and here it's you resumed your download with minimum ease.

shehabmohamed0:
> Can you please explain how this change enables resuming downloads? Please also add tests for both cases, so it will not get broken in the future.

Ok, I will add tests.

ueman:
I wonder whether it makes sense to automatically add range header in the case of a resuming download, that way it would work out of the box, without the user needing to implement anything. The server needs to support them though. (This is not something which necessarily needs to be added to this PR though, I guess)

shehabmohamed0:
> I wonder whether it makes sense to automatically add range header in the case of a resuming download, that way it would work out of the box, without the user needing to implement anything. The server needs to support them though. (This is not something which necessarily needs to be added to this PR though, I guess)

Yes, it's a good idea. I think it's rare to find a server that doesn't support range requests.
Should I change the implementation to  just a boolean parameter in [Dio.download] called resume ?
If the server doesn't support range requests don't use it

shehabmohamed0:
> I wonder whether it makes sense to automatically add range header in the case of a resuming download, that way it would work out of the box, without the user needing to implement anything. The server needs to support them though. (This is not something which necessarily needs to be added to this PR though, I guess)
Please, review the PR again I added download test to insure appending the bytes to the downloaded file

shehabmohamed0:
> I wonder whether it makes sense to automatically add range header in the case of a resuming download, that way it would work out of the box, without the user needing to implement anything. The server needs to support them though. (This is not something which necessarily needs to be added to this PR though, I guess)

I added test case for the file mode append, now can you review the code again

dio_test/lib/src/test/download_tests.dart
```diff
@@ -380,6 +380,54 @@ void downloadTests(
...
+        final path = p.join(tmp.path, 'download_3.txt');
+        final requestedBytes = 1024 * 1024 * 10;
+        var recievedBytes1 = 0;
+        await expectLater(
+          dio.download(
+            '/bytes/$requestedBytes',
+            path,
+            cancelToken: cancelToken,
+            onReceiveProgress: (c, t) {
+              if (c > 5000) {
```
AlexV525, line 394: The test seems fake when it involves the real world's use case. Predicting the length of bytes is meanless since:
1. You've set the condition to cancel when bytes reach 5000.
2. No actual content has been validated.

dio/lib/src/dio_file_mode.dart
```diff
@@ -0,0 +1,24 @@
+/// {@template dio.DioFileMode}
+///  The file access mode when downloading the file.
+/// - [DioFileMode.write]: Mode for opening a file for reading and writing.
+///    The file is overwritten if it already exists. The file is created if it
+///    does not already exist
+/// - [DioFileMode.append]: Mode for opening a file for reading and writing
+///    to the end of it. The file is created if it does not already exist.
+/// {@endtemplate}
+enum DioFileMode {
```
AlexV525, line 9: The name is a bit overwhelming. The type is only used for the download process.

Also the enhanced enum is not working on our minimum SDK support.

dio/lib/src/dio/dio_for_native.dart
```diff
@@ -95,7 +97,12 @@ class DioForNative with DioMixin implements Dio {
     // Shouldn't call file.writeAsBytesSync(list, flush: flush),
     // because it can write all bytes by once. Consider that the file is
     // a very big size (up to 1 Gigabytes), it will be expensive in memory.
-    RandomAccessFile raf = file.openSync(mode: FileMode.write);
+    RandomAccessFile raf = file.openSync(
+      mode: fileMode.map(
+        write: () => FileMode.write,
+        append: () => FileMode.append,
```
AlexV525, line 103: What will happen if the request is not range-supported?

dio_test/lib/src/test/download_tests.dart
shehabmohamed0: canceling the request is mandatory for this test to work because I am cancelling the first download request after loading 5000 bytes and make the second request to verify adding to existing file we are requesting 1024*1024*10 
we are loading 5000 first then 5000 as appending in the check we are checking if the summation of the two first requests equals file size and the test confirms this

dio/lib/src/dio_file_mode.dart
shehabmohamed0: I will convert it to extension

dio/lib/src/dio/dio_for_native.dart
shehabmohamed0: it's a thing developer will validate before using this feature. in the package we are only enabling developers to append data to an existing file.

the feature will be useless if there is no range-support but it's a rare case

dio/lib/src/dio_file_mode.dart
shehabmohamed0: any suggestion for naming

dio/lib/src/dio/dio_for_native.dart
AlexV525: > it's a thing developer will validate before using this feature. in the package we are only enabling developers to append data to an existing file.

Will they? When such features are provided without proper warning or case coverage, developers will have like 50 percent chance to be aware of the edges of the feature, but the user that uses their products will likely have zero chance to know whether their developers have implemented this correctly, so that means we break users experiences.

> the feature will be useless if there is no range-support but it's a rare case

It is common that we live in old systems, and that is what happened with Dio as you may find them in our repo issues.

dio_test/lib/src/test/download_tests.dart
AlexV525: What I mean is validating the length cannot prove the content is being written correctly. Say you have a content sequence `abcdefg`, if you wrote `efgabcd` which is the same length, the length comparison is then fake.

shehabmohamed0: validating the sequence of content is not our job we just need to validate that the bytes are added to the existed file, We just enabled developer to change file access mode.

We need to validate bytes order if we are making a complete feature of resuming the download including checking the existing file size and requesting a range size.

dio/lib/src/dio/dio_for_native.dart
shehabmohamed0: > Will they? When such features are provided without proper warning or case coverage, developers will have like 50 percent chance to be aware of the edges of the feature, but the user that uses their products will likely have zero chance to know whether their developers have implemented this correctly, so that means we break users experiences.

as I said It's not our responsibility to check for range support, we just enabled users to change file access mode

AlexV525:
@ueman Could you follow above reviews and stamp this if applicable?

dio/lib/src/file_access_mode.dart
```diff
@@ -0,0 +1,26 @@
...
+/// - [FileAccessMode.write]: Mode for opening a file for reading and writing.
+///    The file is overwritten if it already exists. The file is created if it
+///    does not already exist
+/// - [FileAccessMode.append]: Mode for opening a file for reading and writing
+///    to the end of it. The file is created if it does not already exist.
+/// {@endtemplate}
+enum FileAccessMode {
+  write,
+  append,
+}
```
AlexV525, line 12: Can we put the enumeration in `src/options.dart` and split the document?

```suggestion
/// {@template dio.options.FileAccessMode}
/// The file access mode when downloading a file, corresponds to a subset of 
/// dart:io::[FileMode].
/// {@endtemplate}
enum FileAccessMode {
  /// Mode for opening a file for reading and writing. The file is overwritten
  /// if it already exists. The file is created if it does not already exist.
  write,

  /// Mode for opening a file for reading and writing to the end of it.
  /// The file is created if it does not already exist.
  append,
}
```

dio/lib/src/file_access_mode.dart
```diff
@@ -0,0 +1,26 @@
...
+
+extension FileAccessModeExtension on FileAccessMode {
+  T map<T>({
+    required T Function() write,
+    required T Function() append,
+  }) {
```
AlexV525, line 18: This seems unnecessary since there is only one caller.

dio_test/lib/src/test/download_tests.dart
```diff
@@ -380,6 +380,54 @@ void downloadTests(
           completes,
         );
       });
+      test('append bytes to previous download', () async {
```
AlexV525, line 383: Missing prepend empty line. Please do `dart format`.

AlexV525:
The CI is complaining about the v1 of `dio_web_adapter`, which should be updated after the PR has landed (I'll do that).

dio/lib/src/dio.dart
```diff
@@ -209,6 +209,9 @@ abstract class Dio {
   /// [deleteOnError] whether delete the file when error occurs.
   /// The default value is [true].
   ///
+  /// [fileAccessMode]
+  /// {@macro dio.FileAccessMode}
```
AlexV525, line 213:
```suggestion
  /// {@macro dio.options.FileAccessMode}
```

dio_test/lib/src/test/download_tests.dart
AlexV525, line 383:
```suggestion

      test('append bytes to previous download', () async {
```

dio/lib/src/dio.dart
```diff
@@ -254,6 +258,7 @@ abstract class Dio {
     ProgressCallback? onReceiveProgress,
     CancelToken? cancelToken,
     bool deleteOnError = true,
+    FileAccessMode mode = FileAccessMode.write,
```
AlexV525, line 261: Oh this should be `fileAccessMode`.

dio/lib/src/dio_mixin.dart
```diff
@@ -286,6 +286,7 @@ abstract class DioMixin implements Dio {
     ProgressCallback? onReceiveProgress,
     CancelToken? cancelToken,
     bool deleteOnError = true,
+    FileAccessMode mode = FileAccessMode.write,
```
AlexV525, line 289: This should be `fileAccessMode` too.

dio_test/lib/src/test/download_tests.dart
```diff
@@ -380,6 +380,55 @@ void downloadTests(
...
+            deleteOnError: false,
+          ),
+          throwsDioException(
+            DioExceptionType.cancel,
+            stackTraceContains: 'test/download_tests.dart',
+          ),
+        );
+
+        final cancelToken2 = CancelToken();
+        var recievedBytes2 = 0;
```
AlexV525, line 409:
```suggestion
        int recievedBytes2 = 0;
```

AlexV525 (commented):
RSLGTM. Also you need to update CHANGELOGs for the main package and the web adapter, just see how other PRs do this.

shehabmohamed0:
@AlexV525  Workflow needs approval

shehabmohamed0:
@AlexV525 What is the next step is there is a problem or you will merge it when possible ?

AlexV525:
> I'll merge this once I have cleared my recent works to gain some bandwidth.

↑

bailyzheng:
Excellent work. Hope release as soon as possible.

PR #2293, opened 2024-09-12
docs: add Japanese README

eltociear:
I created Japanese translated README.


AlexV525:
I'm not sure about whether we should add more languages for the repo. We have English version and Chinese version because we have ability to maintain them, adding another languages might result in lack of updates.

AlexV525:
I'm going to close the request because of the above concern. But thanks for your contribution anyway!

PR #2330, opened 2024-11-21
logging only specific methods

Bahrom2101:
#### These changes help us to improve our logging by reducing unnecessary logs when needed.

I've just added a list field for the LogInterceptor class that can define the options method to print logs.


AlexV525:
The interceptor is working for generally all requests and you can always build your own on top of that. Thanks for the contributions anyway!

PR #2344, opened 2024-12-22
Update dio library doc

mcmah309:
Not sure Why these exist, they make the generated docs look squirrely https://pub.dev/documentation/dio/latest/dio/

### DIFF
--- a/dio/lib/dio.dart
+++ b/dio/lib/dio.dart
@@ -1,8 +1,6 @@
 /// A powerful HTTP client for Dart and Flutter, which supports global settings,
 /// [Interceptors], [FormData], aborting and canceling a request,
 /// files uploading and downloading, requests timeout, custom adapters, etc.
-/// {@category Migration Guide}
-/// {@category Plugins}
 library dio;
 
 export 'src/adapter.dart';


AlexV525:
Could you elaborate "squirrely"?

mcmah309:
"Migration Guide" and "Plugins" point to the same location and are just the regular docs for the dio library. It does not make sense to have these as categories. 

The side view currently looks like this -

Libraries
[browser](https://pub.dev/documentation/dio/latest/browser)
[io](https://pub.dev/documentation/dio/latest/io)
&nbsp;&nbsp;Migration Guide
&nbsp;&nbsp;[dio](https://pub.dev/documentation/dio/latest/dio)
&nbsp;&nbsp;Plugins
&nbsp;&nbsp;[dio](https://pub.dev/documentation/dio/latest/dio)

While it should look like

Libraries
[browser](https://pub.dev/documentation/dio/latest/browser)
[dio](https://pub.dev/documentation/dio/latest/dio)
[io](https://pub.dev/documentation/dio/latest/io)

AlexV525:
Maybe the problem is `library dio` should be `library` only.

mcmah309:
I don't see how that would remove the unnecessary categories

AlexV525:
These are necessary topics IMO. Is there a proper way to keep them while removing the extra library entries?

mcmah309:
I don't think the changes in this PR will remove the topics. Just the sub-categories under "Libraries". "Topics" should remain the same.

AlexV525:
Unfortunately, the topics are gone after removing these categories. Maybe this should be tracked with a Dart doc issue?

image

AlexV525:
Converting to draft since we received no recent response from the upstream. We are still interested in resolving this. :)

PR #2363, opened 2025-01-24
👷 Check bootstrappable package as a prehook of melos and log checked packages

AlexV525:
Picked and updated from https://github.com/cfug/dio/pull/2362

image
image

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -16,6 +16,10 @@ ide:
 command:
   bootstrap:
     runPubGetInParallel: false
+    hooks:
+      pre: |
+        # Check bootstrappable packages.
+        dart ./scripts/melos_packages.dart
   clean:
     hooks:
       post: |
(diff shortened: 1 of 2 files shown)


kuhnroyal:
Just saw this an noticed that it sadly does not work for the bootstrapping itself.

AlexV525:
> Just saw this an noticed that it sadly does not work for the bootstrapping itself.

Could you elaborate? I didn't understand the sentence. 😢

kuhnroyal:
> > Just saw this an noticed that it sadly does not work for the bootstrapping itself.
> 
> Could you elaborate? I didn't understand the sentence. 😢

Sorry for being unclear :)
The script correctly marks the packages but it does not limit the packages being bootstrapped to the same set. As far as I can tell all packages are being bootstrapped. This still only works by using the ugly marker files.
Not much we can do atm, this was more of a mental note.

PR #2375, opened 2025-02-06
Make `LogInterceptor` constant

bambinoua:
Just add the `const` keyword for `LogInterceptor` constructor and make all fields `final` for using it as default parameter. 

- [X] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [X] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [X] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)

### DIFF
--- a/dio/lib/src/interceptors/log.dart
+++ b/dio/lib/src/interceptors/log.dart
@@ -20,7 +20,7 @@ import '../response.dart';
 /// );
 /// ```
 class LogInterceptor extends Interceptor {
-  LogInterceptor({
+  const LogInterceptor({
     this.request = true,
     this.requestHeader = true,
     this.requestBody = false,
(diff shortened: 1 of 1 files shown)


kuhnroyal:
While I agree with this and immutability in general, this is a breaking change.

AlexV525:
I personally disagree with the changes because:
1. The interceptor was designed to be mutable because you can trigger these switches in the runtime for flexibility such as developing a log panel.
2. Changes are breaking change.

bambinoua:
>     1. The interceptor was designed to be mutable because you can trigger these switches in the runtime for flexibility such as developing a log panel.

Erm.. I hard imaging how the developer will mutate flags of `LogInterceptor` on runtime..
```dart
final dio = Dio()..interceptors.add(LogInterceptor());
```

```dart
// Somewhere on runtime
if (doesUserWantToChangeRequestHeaderVisibility) {
  final logInterceptors = dio.interceptors.whereType<LogInterceptor>();
  if (logInterceptors.length == 1) {
    logInterceptors[0].requestHeader = true;
  }
}
```

I can imagine this code:
```dart
final dio = Dio()..interceptors.add(LogInterceptor(
  requestHeader: kIsWeb && kDebugMode,
));
```


As I understand this is a _simple_ logger which intends just to trace requests/responses on development. It just a similar to `print`. If user need to format the message with data, color or whatever he just creates _another_ `print`er.

P.S. But it's just a suggestion, all is up to you. :) I just noticed that the `Interceptor` class has a `const` constructor and decided why do not apply it for `LogInterceptor`.

AlexV525:
> Erm.. I hard imaging how the developer will mutate flags of `LogInterceptor` on runtime..

https://github.com/odinsam/OdinPlugin/blob/747389200711d85df89f82b178c3e60d5cd0961e/odindio/lib/odindio.dart#L13

AlexV525:
Closing based on the above discussion. But we do appreciate your contribution. Thanks!

PR #2389, opened 2025-03-19
docs: fix the return value type.

StanleyCocos:
Because none of the examples below explicitly declare a type, I believe the intended type is actually the Response response; at the top. However, it is not returning a Response; it is an array type.

It is an array type. When I copy and run it, it actually throws an error.

Although it’s not a big issue, I believe explicitly declaring the type would be better.

https://github.com/cfug/dio/blob/d6888cc4b1b085e6cf94c930ff9cb858c271d144/dio/README.md?plain=1#L115

### DIFF
--- a/dio/README.md
+++ b/dio/README.md
@@ -133,7 +133,7 @@ response = await dio.post('/test', data: {'id': 12, 'name': 'dio'});
 ### Performing multiple concurrent requests
 
 ```dart
-List<Response> responses = await Future.wait<Response>([dio.post('/info'), dio.get('/token')]);
+List<Response> responses = await Future.wait([dio.post('/info'), dio.get('/token')]);
 ```
 
 ### Downloading a file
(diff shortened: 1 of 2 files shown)


dio/README.md
```diff
@@ -133,7 +133,7 @@ response = await dio.post('/test', data: {'id': 12, 'name': 'dio'});
 ### Performing multiple concurrent requests
 
 ```dart
-response = await Future.wait([dio.post('/info'), dio.get('/token')]);
+List<Response> responses = await Future.wait<Response>([dio.post('/info'), dio.get('/token')]);
```
AlexV525, line 136: `wait<T>` T seems implicitly promoted to `Response` by your definition, so we can write:
```suggestion
List<Response> responses = await Future.wait([dio.post('/info'), dio.get('/token')]);
```

StanleyCocos: Yes, you said it better.

AlexV525 (approved):

PR #2427, opened 2025-08-02
⚡️ Allow passing an initial interceptors list to the constructor of `Interceptors`

AlexV525:
Introduce another way to fill a list of `Interceptor`s by `Interceptors(initialInterceptors: list)`. This does not change the position of the `ImplyContentTypeInterceptor`.


CaiJingLong (approved):
LGTM

PR #2435, opened 2025-08-09
feat: add handshakeTimeout to Http2Adapter's ConnectionManager

vanelizarov:
In some situations, like when ISP infrastructure is down or misconfigured, user may end up long-waiting for handshake to complete. In my case it was around 1 minute which is huge. This PR adds an optional param `handshakeTimeout` to `ConnectionManager` that after all returns chains with [this future](https://github.com/dart-lang/sdk/blob/9f2218e4509210576abecc79d7184c94a3ad0e26/sdk/lib/io/secure_socket.dart#L633) in `_RawSecureSocket`

### DIFF
--- a/plugins/http2_adapter/lib/src/connection_manager_imp.dart
+++ b/plugins/http2_adapter/lib/src/connection_manager_imp.dart
@@ -4,9 +4,11 @@ part of 'http2_adapter.dart';
 class _ConnectionManager implements ConnectionManager {
   _ConnectionManager({
     Duration? idleTimeout,
+    Duration? handshakeTimeout,
     this.onClientCreate,
     this.proxyConnectedPredicate = defaultProxyConnectedPredicate,
-  }) : _idleTimeout = idleTimeout ?? const Duration(seconds: 1);
+  })  : _idleTimeout = idleTimeout ?? const Duration(seconds: 1),
+        _handshakeTimout = handshakeTimeout ?? const Duration(seconds: 1);
 
   /// Callback when socket created.
   ///
@@ -21,6 +23,9 @@ class _ConnectionManager implements ConnectionManager {
   /// connections. For the sake of socket reuse feature with http/2,
   /// the value should not be less than 1 second.
   final Duration _idleTimeout;
+
+  /// Sets the handshake timeout(milliseconds) for secure socket connections
+  final Duration _handshakeTimout;
 
   /// Saving the reusable connections
   final _transportsMap = <String, _ClientTransportConnectionState>{};
@@ -175,7 +180,7 @@ class _ConnectionManager implements ConnectionManager {
         context: clientConfig.context,
         onBadCertificate: clientConfig.onBadCertificate,
         supportedProtocols: ['h2'],
-      );
+      ).timeout(_handshakeTimout);
       _throwIfH2NotSelected(target, socket);
       return socket;
     }
@@ -252,7 +257,7 @@ class _ConnectionManager implements ConnectionManager {
       context: clientConfig.context,
       onBadCertificate: clientConfig.onBadCertificate,
       supportedProtocols: ['h2'],
-    );
+    ).timeout(_handshakeTimout);
     _throwIfH2NotSelected(target, socket);
 
     proxySubscription.cancel();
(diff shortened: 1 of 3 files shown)


vanelizarov:
any updates?

plugins/http2_adapter/lib/src/connection_manager_imp.dart
```diff
@@ -4,9 +4,11 @@ part of 'http2_adapter.dart';
 class _ConnectionManager implements ConnectionManager {
   _ConnectionManager({
     Duration? idleTimeout,
+    Duration? handshakeTimeout,
     this.onClientCreate,
     this.proxyConnectedPredicate = defaultProxyConnectedPredicate,
-  }) : _idleTimeout = idleTimeout ?? const Duration(seconds: 1);
+  })  : _idleTimeout = idleTimeout ?? const Duration(seconds: 1),
+        _handshakeTimout = handshakeTimeout ?? const Duration(seconds: 1);
```
AlexV525, line 11: Is this a proper timeout? How do we evaluate? Do you have any data sources that back this value up?

plugins/http2_adapter/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ See the [Migration Guide][] for the complete breaking changes list.**
 
 *None.*
 
+## 2.7.0
```
AlexV525, line 10: We don't update the version header in changes. Please refer to other PRs and see how they handle this.

plugins/http2_adapter/lib/src/connection_manager_imp.dart
```diff
@@ -175,7 +180,7 @@ class _ConnectionManager implements ConnectionManager {
         context: clientConfig.context,
         onBadCertificate: clientConfig.onBadCertificate,
         supportedProtocols: ['h2'],
-      );
+      ).timeout(_handshakeTimout);
```
AlexV525, line 183: Do we have other libraries that have a similar approach? I'm not sure how this is related to *handshake* only. Could it be related to other processes?

plugins/http2_adapter/lib/src/connection_manager_imp.dart
vanelizarov: Changed it to be 15 seconds by default. As far as I could tell, this is the average value in all libraries with a small margin; they usually use exactly 10-15 seconds.

plugins/http2_adapter/lib/src/connection_manager_imp.dart
vanelizarov: The only example I have in mind that allows to explicitly set handhshake timeout is `net/http.Transport` package in Go that has a property `TLSHandshakeTimeout`

In Java there's a [`Socket.connect`](https://developer.android.com/reference/java/net/Socket#connect(java.net.SocketAddress,%20int)) method that allows to set TCP-connection timeout which is also applied to handshake because it can only begin after connection is established

Technically in this case this property is applied only to [`handshake future`](https://github.com/dart-lang/sdk/blob/9f2218e4509210576abecc79d7184c94a3ad0e26/sdk/lib/io/secure_socket.dart#L633)

plugins/http2_adapter/lib/src/http2_adapter.dart
```diff
@@ -140,7 +140,9 @@ class Http2Adapter implements HttpClientAdapter {
 
     if (hasRequestData) {
       Future<dynamic> requestStreamFuture = requestStream!.listen((data) {
-        stream.outgoingMessages.add(DataStreamMessage(data));
+        try {
+          stream.outgoingMessages.add(DataStreamMessage(data));
+        } catch (_) {}
```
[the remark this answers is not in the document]
vanelizarov: Eventually, I got a `StateError: Bad state: Cannot add event after closing` in Crashlytics. I can’t reproduce it locally — maybe it’s caused by some race condition, I’m not sure.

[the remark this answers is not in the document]
AlexV525: Let's revert until we have a clear path on how to reproduce and what we are fixing.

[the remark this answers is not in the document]
vanelizarov: Ok, sure
Should I add a TODO?

AlexV525 (approved):
LGTM. Sorry for the long wait.

Issue #2138, opened 2024-03-13
A request with an interceptor, which produces an error, makes the code stuck.


yehorh:
### Package

dio

### Version

5.4.1

### Operating-System

Android

### Output of `flutter doctor -v`

```shell
flutter doctor -v
[✓] Flutter (Channel stable, 3.19.3, on macOS 14.4 23E214 darwin-arm64, locale en-UA)
    • Flutter version 3.19.3 on channel stable at /Users/yehorh/opt/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision ba39319843 (6 days ago), 2024-03-07 15:22:21 -0600
    • Engine revision 2e4ba9c6fb
    • Dart version 3.3.1
    • DevTools version 2.31.1

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0-rc1)
    • Android SDK at /Users/yehorh/Library/Android/sdk
    • Platform android-34, build-tools 34.0.0-rc1
    • Java binary at: /Users/yehorh/Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 15.3)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15E204a
    • CocoaPods version 1.14.3

[✓] Chrome - develop for the web
    • Chrome at /Applications/Google Chrome.app/Contents/MacOS/Google Chrome

[✓] Android Studio (version 2023.2)
    • Android Studio at /Users/yehorh/Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)

[✓] IntelliJ IDEA Ultimate Edition (version 2023.3.5)
    • IntelliJ at /Users/yehorh/Applications/IntelliJ IDEA Ultimate.app
    • Flutter plugin version 78.3.1
    • Dart plugin version 233.14888

[✓] VS Code (version 1.87.2)
    • VS Code at /Applications/Visual Studio Code.app/Contents
    • Flutter extension version 3.84.0

[...]

PR #2139, opened 2024-03-14
fix: Catch sync/async* exceptions in interceptor's handlers

CaiJingLong:
Resolves #2138

### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [x] I have added the required tests to prove the fix/feature I'm adding
- [ ] I have updated the documentation (if necessary)
- [x] I have run the tests without failures
- [x] I have updated the `CHANGELOG.md` in the corresponding package

PR #2478, opened 2026-01-05
fix: handle synchronous exceptions in QueuedInterceptor

Danilospano00:
Fixes a bug where QueuedInterceptor would cause requests to hang indefinitely when an interceptor callback threw a synchronous exception.

The _handleQueue method now accepts an onError callback that properly completes the handler when a synchronous exception is caught, ensuring requests fail gracefully instead of hanging.

## Description

Fixes a bug where `QueuedInterceptor` would cause requests to hang indefinitely when an interceptor callback threw a synchronous exception.

## Problem

When using `QueuedInterceptor`, if an interceptor callback (`onRequest`, `onResponse`, or `onError`) threw a synchronous exception (not wrapped in a `Future`), the request would hang until timeout. This occurred because the `_handleQueue` method's `try-catch` block caught the exception but only called `_processNextInQueue()` without completing the handler's `Completer`, leaving the request waiting forever.

## Changes

Modified `_handleQueue` to accept an `onError` callback parameter that properly handles synchronous exceptions:

- **onRequest/onResponse**: Wraps the error in `DioException` and calls `handler.reject()` to fail the request gracefully
- **onError**: Wraps the error in `DioException` and calls `handler.next()` to pass the error to the next interceptor in the chain

### DIFF
--- a/.github/workflows/coverage_comment.yml
+++ b/.github/workflows/coverage_comment.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - name: Download artifact
         id: download-artifact
-        uses: dawidd6/action-download-artifact@v11
+        uses: dawidd6/action-download-artifact@v12
         with:
           workflow: tests.yml
           workflow_conclusion: success


AlexV525:
Thanks for the fix. I was wondering:
1. Can the code add more comments to tell people what happens at this place?
2. Does it also fix https://github.com/cfug/dio/issues/2138, which https://github.com/cfug/dio/pull/2139 didn't make it?

Issue #2167, opened 2024-03-27
Interceptor onError handling broken


fischermario:
### Package

dio

### Version

5.4.2

### Operating-System

Android

### Adapter

Default Dio

### Output of `flutter doctor -v`

```shell
[✓] Flutter (Channel stable, 3.19.4, on Ubuntu **************, locale en_US.UTF-8)
    • Flutter version 3.19.4 on channel stable at **************
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 68bfaea224 (**************), 2024-03-20 15:36:31 -0700
    • Engine revision a5c24f538d
    • Dart version 3.3.2
    • DevTools version 2.31.1

[✓] Android toolchain - develop for Android devices (Android SDK version 34.0.0)
    • Android SDK at **************
    • Platform android-34, build-tools 34.0.0
    • Java binary at: /snap/android-studio/151/jbr/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)
    • All Android licenses accepted.

[✓] Chrome - develop for the web
    • Chrome at google-chrome

[✓] Linux toolchain - develop for Linux desktop
    • clang version 10.0.0-4ubuntu1
    • cmake version 3.16.3
    • ninja version 1.10.0
    • pkg-config version 0.29.1

[✓] Android Studio (version 2023.2)
    • Android Studio at /snap/android-studio/151
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)

[✓] VS Code (version 1.85.0)
    • VS Code at /usr/share/code
    • Flutter extension version 3.84.0

[✓] Connected device (2 available)
    • sdk gphone64 x86 64 (mobile) • emulator-5554 • android-x64    • Android 13 (API 33) (emulator)
    • Chrome (web)                 • chrome        • web-javascript • Google Chrome 122.0.6261.128

[✓] Network resources
    • All expected network resources are available.

• No issues found!
```

### Dart Version

3.3.2

### Steps to Reproduce

The onError handler of a new class that extends the "Interceptor" class will not allow subsequent interceptors to successfully catch exceptions.

[...]

PR #2499, opened 2026-03-11
fix(dio): prevent request hang when async interceptor throws

### LINKED ISSUE #2138: A request with an interceptor, which produces an error, makes the code stuck.

ersanKolay:
## Summary

Fixes #2138.

When an interceptor overrides `onRequest`/`onResponse`/`onError` with an `async` implementation and throws after an `await` without calling the handler, the request hangs indefinitely. The handler's `Completer` is never completed because:

1. The callback typedef is `void`, so the returned `Future` is discarded
2. The unhandled async error goes to the zone's uncaught error handler
3. Nobody ever calls `handler.next()`/`reject()`/`resolve()`

```dart
// This hangs forever:
class MyInterceptor extends Interceptor {
  @override
  void onRequest(RequestOptions options, RequestInterceptorHandler handler) async {
    await refreshToken(); // throws
    handler.next(options); // never reached
  }
}
```

### Approach

Each interceptor callback is now invoked inside a forked `Zone` with a custom `handleUncaughtError`. When an uncaught async error occurs and the handler hasn't been completed yet, the zone handler calls `handler.reject()` (or `handler.next()` for error interceptors) to unblock the request.

This approach differs from the reverted PR #2139 which changed the callback typedefs to `FutureOr<void>` and awaited the return value. That changed the microtask ordering and broke multi-interceptor chains (#2167). `Zone.fork()` does **not** await the callback -- it only catches async errors that would otherwise be lost, preserving the original execution order.

### Why this works

| Scenario | Before | After |
|----------|--------|-------|
| Sync interceptor, calls handler | Works | Works (no change) |
| Sync interceptor, throws | Works (propagates via Future) | Works (sync errors pass through Zone.run) |
| Async interceptor, calls handler | Works | Works (zone has no effect) |
| Async interceptor, throws | **Hangs forever** | Zone catches error, rejects handler |
| Duplicate handler.next() calls | Throws StateError | Throws StateError (sync, passes through) |

## Test plan

- Added 3 new tests for async throw scenarios (onRequest, onResponse, onError)
- All 218 existing tests pass
- All 32 interceptor-specific tests pass (including "duplicate handler calls" and the existing "Caught exceptions before handler called" test)
- `dart format` and `dart analyze` clean

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -5,7 +5,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 
 ## Unreleased
 
-*None.*
+- Fix request hanging indefinitely when async interceptor callbacks throw without calling the handler.
 
 ## 5.9.2
 
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -420,7 +420,9 @@ abstract class DioMixin implements Dio {
               createInterceptorZone(
                 handler,
                 (error) => handler.reject(
-                    assureDioException(error, requestOptions), true),
+                  assureDioException(error, requestOptions),
+                  true,
+                ),
               ).run(() => cb(state.data as RequestOptions, handler));
               return handler.future;
             }),
@@ -446,7 +448,9 @@ abstract class DioMixin implements Dio {
               createInterceptorZone(
                 handler,
                 (error) => handler.reject(
-                    assureDioException(error, requestOptions), true),
+                  assureDioException(error, requestOptions),
+                  true,
+                ),
               ).run(() => cb(state.data as Response, handler));
               return handler.future;
             }),
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -386,6 +386,25 @@ abstract class DioMixin implements Dio {
       }
     }
 
+    // Create an error zone that catches uncaught async errors from
+    // interceptor callbacks. This prevents requests from hanging when
+    // an async interceptor throws without calling the handler.
+    // Synchronous exceptions are not caught here and propagate normally.
+    Zone createInterceptorZone(
+      _BaseHandler handler,
+      void Function(Object error) onUncaughtError,
+    ) {
+      return Zone.current.fork(
+        specification: ZoneSpecification(
+          handleUncaughtError: (self, parent, zone, error, stackTrace) {
+            if (!handler.isCompleted) {
+              onUncaughtError(error);
+            }
+          },
+        ),
+      );
+    }
+
     // Convert the request interceptor to a functional callback in which
     // we can handle the return value of interceptor callback.
     FutureOr Function(dynamic) requestInterceptorWrapper(
@@ -398,7 +417,11 @@ abstract class DioMixin implements Dio {
             requestOptions.cancelToken,
             Future(() async {
               final handler = RequestInterceptorHandler();
-              cb(state.data as RequestOptions, handler);
+              createInterceptorZone(
+                handler,
+                (error) => handler.reject(
+                    assureDioException(error, requestOptions), true),
+              ).run(() => cb(state.data as RequestOptions, handler));
               return handler.future;
             }),
           );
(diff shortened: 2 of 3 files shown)


AlexV525:
@ersanKolay Hi, are you still going to work on this?

ersanKolay:
Yep, just pushed an update covering the review notes — `stackTrace` is threaded into the `DioException` now so you get the real throw site, errors get forwarded to `parent.handleUncaughtError` when the handler is already completed, and the 10ms delays in the tests are swapped for `Future<void>.value()`. Analyzer clean, interceptor suite green.

AlexV525 (approved):
LGTM. Thanks for the fix!

kuhnroyal:
I remember the zones having quite an overhead performance wise so we removed some usage in the past. Shouldn't we try to type the interceptors as FutureOr instead and then catch the error?

AlexV525:
> I remember the zones having quite an overhead performance wise so we removed some usage in the past. Shouldn't we try to type the interceptors as FutureOr instead and then catch the error?

@kuhnroyal From my experience this should have zero impact to the performance (although we don't have corresponding benchmark), unless we heavily use `ZoneSpecification` to override queues like `scheduleMicrotask`.

PR #2510, opened 2026-04-22
fix: parse single-string JSON response (#2509)

### LINKED ISSUE #2509: When single string json response received it doesn't parse it.
### Package

dio

### Version

5.9.1

### Operating-System

Android

### Adapter

Default Dio

```shell
When server responds "adc89802-9c92-418f-b2ce-8f89308571f9" and Content-Type: application/json which is valid json. Dio should output adc89802-9c92-418f-b2ce-8f89308571f9 /double quote removed/. But it didn't.
```

### Dart Version

3.11.4

### Steps to Reproduce

`final _result = await _dio.fetch<String>(_options);`

### Expected Result

When server send: "adc89802-9c92-418f-b2ce-8f89308571f9"
Dio should parse it into adc89802-9c92-418f-b2ce-8f89308571f9

Like this:
  final json = "\"adc89802-9c92-418f-b2ce-8f89308571f9\"";
  final parsed = jsonDecode(json);
  print(parsed);

### Actual Result

double quoted raw response

IKBASYSTEMS:
Fixes #2509.

`ResponseType.json` with a single JSON string value (e.g. `"ok"`) was not parsed correctly.
This updates parsing in `dio_mixin` and adds regression tests in `options_test.dart` (with adapter support in `test/mock/adapters.dart`).

Checklist:
✅ tests added
✅ branch up to date (you are)
✅ run tests first, then mark done

### DIFF
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -563,6 +563,22 @@ abstract class DioMixin implements Dio {
           reqOpt,
           responseBody,
         );
+        if (data is String &&
+            data.isNotEmpty &&
+            T == String &&
+            reqOpt.responseType == ResponseType.plain &&
+            Transformer.isJsonMimeType(
+              responseBody.headers[Headers.contentTypeHeader]?.first,
+            )) {
+          try {
+            final decoded = jsonDecode(data);
+            if (decoded is String) {
+              data = decoded;
+            }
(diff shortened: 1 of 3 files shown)


AlexV525 (changes requested):
The issue is missing a specific repro path thus the fix is also invalid at this moment. We have multiple options and checks that guard JSON parses.

PR #2512, opened 2026-05-10
🐛 Fix `validateCertificate` to fire pre-emission for TLS pinning

### LINKED ISSUE #2418: validateCertificate callback fires after request body is sent, so true TLS-pinning is impossible
### Request Statement

`IOHttpClientAdapter.validateCertificate` runs only after the HTTP
request has already been flushed to the socket.
This means an attacker who can present a publicly trusted certificate for
`example.com` will receive the full request body (and often the
headers) before the app’s pinning code gets a chance to abort.

So while the callback can block the response from being delivered to the
caller, it cannot stop leakage of sensitive outbound data.

The behavior breaks the primary use-case for validateCertificate:
certificate / public-key pinning.

### Solution Brainstorm

`validateCertificate` (or a new early-handshake hook) should be invoked
before any body bytes or HTTP headers are written:
- Open TCP socket
- Start TLS handshake
- Call validateCertificate with the server’s leaf cert
- Only if it returns true finish handshake and write request
- That is how Android’s <pin-set> and iOS ATS / TrustKit hooks work, and
it’s how users expect pinning to behave.

realmeylisdev:
Fixes #2418.

### Summary

`IOHttpClientAdapter.validateCertificate` runs only **after** `HttpClientRequest.close()` has flushed the request method, headers, and body to the socket. The callback can prevent the response from being delivered to the caller, but it cannot prevent leakage of outbound request data — defeating the canonical purpose of certificate or public-key pinning.

This PR makes `validateCertificate` fire **between the TLS handshake and the first HTTP byte** for direct HTTPS connections by installing a custom `HttpClient.connectionFactory`. When the callback returns `false`, the request body never leaves the client.

### Reproducing the bug (on `main` today)

The bug is structural — no exploit setup needed.

1. Read `_fetch` in `dio/lib/src/adapters/io_adapter.dart` on `main`:
   ```dart
   Future<HttpClientResponse> future = request.close();  // sends body
   ...
   final responseStream = await future;                  // response head arrives

   if (validateCertificate != null) {                    // <-- runs HERE, too late
     final bool isCertApproved = validateCertificate!(
       responseStream.certificate, host, port,
     );
     if (!isCertApproved) {
       throw DioException.badCertificate(...);
     }
   }
   ```
2. By the time the callback runs, the request has already been transmitted and the server has begun replying. The library can block the response from being returned to the app, but the credentials/tokens/body are already on the wire to an attacker who presented any publicly trusted certificate for the wrong host.

Demonstrable end-to-end:

1. Start a local `HttpServer.bindSecure('localhost', 0, ctx)` with a self-signed cert. Record bytes received per request.
2. Configure dio:
   ```dart
   dio.httpClientAdapter = IOHttpClientAdapter(
     validateCertificate: (cert, host, port) => false,
   );
   ```
3. `await dio.post('https://localhost:\$port/leak', data: 'SENSITIVE');`
4. **On `main`:** server records `bytesReceived > 0` (the payload reached it).
5. **With this PR:** server records `bytesReceived == 0`, the request handler is never invoked, and the caller gets `DioException.badCertificate` from inside `openUrl` — before any byte is written.

This is encoded as the load-bearing regression test in `dio/test/pinning_test.dart`:

```dart
test('rejection blocks request body emission', () async {
  ...
  expect(error.type, DioExceptionType.badCertificate);
  expect(bytesReceived, 0);
  expect(requestsHandled, 0);
});
```

### Approach

Install `HttpClient.connectionFactory` in `_createHttpClient` whenever `validateCertificate != null` and `createHttpClient` is not supplied. The factory:

1. **Plain HTTP** (`url.scheme != 'https'`) — short-circuits to `Socket.startConnect`. Pre-emission validation is skipped (no TLS handshake to gate); the post-response block still fires with a null cert.
2. **HTTPS direct** (no proxy) — `SecureSocket.startConnect(...)` with `supportedProtocols: ['http/1.1']`, awaits the socket once to run the user's validator, then returns that same stock `ConnectionTask` (no `ConnectionTask.fromSocket` needed, so no Dart 3.5+ requirement). On rejection, throws a private `_BadCertificateException` (a `HandshakeException` subtype carrying the cert) — `_fetch` catches it and converts to `DioException.badCertificate(error: cert)` so callers see the same exception type as before.
3. **HTTPS via proxy** — returns a plain socket to the proxy and lets `HttpClient` perform its own `CONNECT` tunnel and TLS handshake (we cannot pre-emption-validate proxy connections because `HttpClient` writes its own `CONNECT` over whatever socket we return). Validation runs only via the post-response block in this case — documented as a known limitation.

The post-response validation block in `_fetch` is retained and runs **unconditionally** when `validateCertificate != null`, providing defense in depth:

- On the **direct-HTTPS pre-emission path** it is redundant (same cert, idempotent for pinning) — callbacks with side effects will see one extra call per request, documented.
- It is the **only line of defense** for paths where pre-emission did not run: custom `createHttpClient`, late-mutation of `validateCertificate` after the cached client was built, and HTTPS through a proxy.

On the pre-emission path, `validateCertificate` is the sole gate for certificate trust — system / CA validation is bypassed (equivalent to `HttpClient.badCertificateCallback: (_, _, _) => true` in the existing example). This lets self-signed and pinned-CA setups work without supplying `createHttpClient`. Documented on the doc-comment.

### Behavioral changes

- For direct HTTPS pinning, validation happens before request emission. **The byte-leak documented in #2418 is fixed for this case.**
- The pre-emission path is **per TCP connection** (matches `dio_http2_adapter`); the post-response path is **per request**.
- HTTPS through a proxy still uses post-response validation only (HttpClient owns the `CONNECT` + TLS path).
- Min Dart SDK is unchanged (`>=2.18.0`). The pre-emission hook uses the stock `SecureSocket.startConnect` — available well before 2.18 — so it needs neither `ConnectionTask.fromSocket` (Dart 3.5+) nor any other SDK-version-gated API.
- `example_dart/lib/certificate_pinning.dart` is simplified — the `SecurityContext(withTrustedRoots: false)` + `badCertificateCallback: (_, _, _) => true` workaround is no longer needed.
- The `DioException.badCertificate` stack-trace test in `dio/test/stacktrace_test.dart` is updated to use the `createHttpClient` escape hatch, since the new factory path can't be exercised against the existing `MockHttpClient` (no stub for `connectionFactory=`).
- The local `HttpServer.bindSecure` tests generate a throwaway self-signed `localhost` keypair at runtime via `openssl` (in `setUpAll`, into a temp dir, cleaned up in `tearDownAll`). No private key is committed to the repository.

### Edge cases acknowledged

| Case | Behavior |
|---|---|
| `createHttpClient` supplied | Factory not installed; validation runs post-response only (legacy 5.x behavior). |
| Late mutation of `validateCertificate` (after first request) | Factory was gated at construction; pre-emission does not retro-apply. Post-response validation runs normally. |
| HTTPS through a proxy | Factory returns plain socket to proxy; `HttpClient` does `CONNECT` + TLS itself. Pre-emission is bypassed; post-response block validates. Documented. |
| Authenticated HTTP proxy (`Proxy-Authorization`) | `HttpClient.findProxy` / `HttpClient.addProxyCredentials` continue to work because we delegate `CONNECT` to `HttpClient`. |
| `SecurityContext` with mTLS + `validateCertificate` without `createHttpClient` | mTLS context is not used on the pre-emission path. Recommend `createHttpClient` for mTLS + pinning. |
| HTTPS → HTTP redirect | Factory invoked with `http://`; short-circuits. Same as today. |
| Anonymous-DH cipher (no peer cert) | `validateCertificate(null, host, port)` invoked. Same as 5.9.x. |
| Connection pool reuse | Pre-emission runs once per pooled connection; post-response runs once per request. |
| Plain HTTP | Pre-emission skipped; post-response fires with null cert (matches 5.9.x). |

- [x] I have updated this branch with the latest \`main\` branch to avoid conflicts (via merge from master or rebase)
- [x] I have updated the \`CHANGELOG.md\` in the corresponding package

### Test plan

- [x] \`cd dio && dart test\` — full suite passes; the \`tls\`-tagged network tests are skipped by default (run them with \`-P all\`).
- [x] \`cd dio && dart analyze lib/\` — no issues. (Two pre-existing \`unreachable_switch_default\` warnings in \`test/transformer_test.dart\` surfaced by the SDK bump; they are unrelated to this fix.)
- [x] \`'rejection blocks request body emission'\` — load-bearing regression test for #2418. Fails on \`main\`, passes here. Asserts \`bytesReceived == 0\` and the request handler is never invoked.
- [x] \`'approval lets the request through'\` — happy path.
- [x] \`'createHttpClient escape hatch keeps post-response validation'\` — locks in the fallback path.
- [x] \`'plain http skips pre-emission validation'\` — confirms HTTP short-circuit; post-response still fires with null cert.
- [x] \`'validateCertificate set after construction still fires (legacy post-response path)'\` — regression guard against the H2 issue surfaced in the code review.
- [x] \`'through a CONNECT proxy (post-response validation)'\` group — two tests exercising a local stub \`CONNECT\` proxy fixture, confirming proxy + pinning works (via post-response) and rejection raises \`badCertificate\`.
- [x] \`'two pooled requests: 1 pre-emission + 2 post-response validations'\` — deterministic local-server test: two back-to-back requests share one pooled connection, so the validator fires 3 times (1 pre-emission + 2 post-response). Replaces the former network-dependent \`'2 requests'\` test.
- [x] Manual: \`example_dart/lib/certificate_pinning.dart\` against badssl.com after refreshing fingerprint via \`scripts/prepare_pinning_certs.sh\` — trusted host returns a response, bad-host throws \`badCertificate\`.

### Known limitations / follow-ups

- HTTPS through a proxy still leaks request bytes (post-response only). Pre-emission behind a proxy would require either intercepting at a lower layer than \`HttpClient.connectionFactory\` (not exposed) or rebuilding the HTTP/1.1 framing layer (out of scope for this PR — \`dio_http2_adapter\` already does this for h2). Users needing pinning through a proxy today should use \`createHttpClient\` and install a custom \`HttpClient.connectionFactory\`.
- Cancellation during the pre-emission TLS handshake is not propagated to abort the in-flight \`SecureSocket.startConnect\` — the eventual socket is dropped on the floor. Pre-existing semantics weren't great either; not a regression. Follow-up issue welcome.


dio/lib/src/dio_mixin.dart
AlexV525, line 1: Please revert since we don't match the requirement here.

dio/lib/src/adapters/io_adapter.dart
AlexV525, line 1: What about other adapters?

dio/pubspec.yaml
AlexV525, line 1: Please revert.

dio/lib/src/adapters/io_adapter.dart
```diff
@@ -244,7 +324,63 @@ class IOHttpClientAdapter implements HttpClientAdapter {
...
+      if (validator != null) {
+        final cert = ss.peerCertificate;
+        if (!validator(cert, url.host, url.port)) {
+          ss.destroy();
+          throw _BadCertificateException(url.host, url.port, cert);
+        }
+      }
+      return ss;
+    }();
+    return ConnectionTask.fromSocket(socketFuture, () {});
```
AlexV525: The implementation of `ConnectionTask` is simple, try to copy it into the code base to avoid bumping the SDK version too high.

AlexV525 (changes requested):

realmeylisdev:
Pushed review fixes ([23aa0c0](https://github.com/realmeylisdev/dio/commit/23aa0c0)):

- **H2 (late mutation):** the legacy post-response validation block now runs unconditionally when `validateCertificate` is non-null. Setting the field after the first request now fires the callback (was silently no-oping). For direct HTTPS the callback fires twice (pre-emission + post-response, idempotent for fingerprint pinning) — documented.
- **H1 + M1 (proxy):** the manual `CONNECT` tunnel in the factory was broken — `HttpClient` writes its own `CONNECT` over the socket we return, so we ended up issuing two `CONNECT` requests and the TLS handshake corrupted. Discovered by the new proxy test failing. Dropped the manual tunnel; HTTPS through a proxy now uses post-response validation only (HttpClient handles `CONNECT` + TLS). Added two proxy tests via a local stub `CONNECT` fixture.
- **Fixture README:** added `dio/test/_pinning/README.md` with the `openssl` command to regenerate the 10-year self-signed cert.
- **Cleanups:** dropped redundant `as Socket` cast; consolidated dangling `## Unreleased *None.*` with the new entry; removed `dart:convert` import that's no longer used.

PR description updated to reflect the corrected scope (proxy = known limitation, post-response only).

AlexV525:
Still not covering previous comments

realmeylisdev:
@AlexV525 thanks for the review. Pushed [8a66e1e](https://github.com/realmeylisdev/dio/commit/8a66e1e) addressing all four points:

**1. `dio/pubspec.yaml` (revert)** ✅ Restored `sdk: '>=2.18.0 <4.0.0'`. Same for `dio_test/pubspec.yaml`.

**2. `dio/lib/src/dio_mixin.dart` (revert)** ✅ Back to `abstract class DioMixin implements Dio` with the `// TODO(EVERYONE): Use 'mixin class' when the lower bound of SDK is raised to 3.0.0.` comment intact.

**3. `dio/lib/src/adapters/io_adapter.dart` line 384 (vendor `ConnectionTask`)** ✅ Achieved the underlying goal (keep min SDK at 2.18) without literal vendoring. Quick note on the constraint and the workaround:

> Literally copying `ConnectionTask` into the codebase isn't viable across the Dart 3.0–3.4 window — Dart 3.0 made `ConnectionTask` a `final class`, so any vendored `class _ConnectionTask implements ConnectionTask` would fail to compile on Dart 3.0+. The public factory `ConnectionTask.fromSocket` that I'd been using is itself Dart 3.5+ only.

The fix is to use `SecureSocket.startConnect(...)` — a stable `dart:io` API that predates dio's 2.18 floor — which returns a real SDK-built `ConnectionTask<SecureSocket>`. We await `task.socket` once to validate the leaf certificate, then return the same `task`; `HttpClient` re-awaits `task.socket` and gets the already-resolved socket immediately (`Future` awaits are idempotent). Covariance does the upcast from `ConnectionTask<SecureSocket>` to `ConnectionTask<Socket>` at the return position.

Diff for the affected branch is small:
- before: `await SecureSocket.connect(...)` + `ConnectionTask.fromSocket(socketFuture, () {})`
- after: `await SecureSocket.startConnect(...)` + `await task.socket` (for validation) + `return task`

**4. \"What about other adapters?\"** ✅ The fix is intentionally scoped to `IOHttpClientAdapter` because that's where the bug exists. For completeness:

- **`plugins/http2_adapter`** — already does pre-emission validation correctly at `connection_manager_imp.dart:133-150`. No change needed; this PR brings `IOHttpClientAdapter` to parity.
- **`plugins/web_adapter`** — no `validateCertificate` field on the web adapter (it's only on `IOHttpClientAdapter`). TLS is handled by the browser and is not introspectable from Dart. Not applicable.
- **`plugins/native_dio_adapter`** — delegates to `cronet_http` / `cupertino_http`, where TLS pinning is platform-native (Android NSC, iOS ATS/TrustKit). Not exposed via a dio-level callback. Not applicable.

Happy to add a one-line cross-reference comment in `io_adapter.dart` pointing to this clarification if you'd like that captured in source.

All 222 tests pass; `dart analyze lib/` is clean.

dio/pubspec.yaml
realmeylisdev: Reverted in [8a66e1e](https://github.com/cfug/dio/pull/2512/commits/8a66e1ebd3b907cd9569a396e4d9efaa5ef8c8db) — back to `sdk: '>=2.18.0 <4.0.0'`. Same for `dio_test/pubspec.yaml`. The pre-emission factory no longer needs Dart 3.5+.

dio/lib/src/dio_mixin.dart
realmeylisdev: Reverted in [8a66e1e](https://github.com/cfug/dio/pull/2512/commits/8a66e1ebd3b907cd9569a396e4d9efaa5ef8c8db) — back to `abstract class DioMixin implements Dio` with the original `TODO(EVERYONE): Use \`mixin class\` when the lower bound of SDK is raised to 3.0.0.` comment.

dio/lib/src/adapters/io_adapter.dart
realmeylisdev: Addressed differently in [8a66e1e](https://github.com/cfug/dio/pull/2512/commits/8a66e1ebd3b907cd9569a396e4d9efaa5ef8c8db) — literal vendoring is blocked because Dart 3.0 made `ConnectionTask` a `final class` (any vendored `class _ConnectionTask implements ConnectionTask` fails to compile from 3.0 onward), and `ConnectionTask.fromSocket` is Dart 3.5+ only.

Switched to `SecureSocket.startConnect(...)` (stable pre-2.18 API) which returns a real SDK `ConnectionTask<SecureSocket>`. We `await task.socket` once to validate the leaf cert, then `return task`; HttpClient re-awaits the same Future and gets the resolved socket immediately. Covariance handles the upcast to `ConnectionTask<Socket>`. No vendoring, no SDK floor change — same observable behavior.

dio/lib/src/adapters/io_adapter.dart
realmeylisdev: The fix is intentionally scoped to `IOHttpClientAdapter` since that's the adapter with the bug. Quick rundown of the others:

- **`plugins/http2_adapter`** — already does pre-emission validation correctly at `connection_manager_imp.dart:133-150`. This PR brings `IOHttpClientAdapter` to parity.
- **`plugins/web_adapter`** — no `validateCertificate` field exists on the web adapter; TLS is browser-controlled and not introspectable from Dart. Not applicable.
- **`plugins/native_dio_adapter`** — delegates to `cronet_http` / `cupertino_http`; TLS pinning is platform-native (Android NSC, iOS ATS/TrustKit), not a dio-level callback. Not applicable.

Happy to add this as a one-line cross-reference comment in `io_adapter.dart` if you'd like that captured in source.

AlexV525:
Tests didn't seem passing.
Additionally, I'm considering ignoring these pinning test certs for Git for security concerns.

AlexV525:
@realmeylisdev Would you like to bring this up again?

realmeylisdev:
@AlexV525 thanks — picked this back up. Both points are addressed in a7372d3.

**1. CI / failing tests.** Two distinct causes against the previous HEAD:
- `dart format` wasn't clean on `io_adapter.dart`, `pinning_test.dart`, and `stacktrace_test.dart`, so the `stable` job failed at the format step before it ever reached the tests. Reformatted.
- The `'2 requests share connection (1 approval)'` test asserted `approvalCount == 1`, but the post-response validation block still fires once per request — so a 2-request pooled flow invokes the validator **3 times** (1 pre-emission + 2 post-response), not 1. I replaced that network-dependent test with a deterministic local-server test, `'two pooled requests: 1 pre-emission + 2 post-response validations'`, asserting 3.

**2. Test certs.** Agreed — no private key should live in git. The committed `test/_pinning/` keypair and README are removed; the local `HttpServer.bindSecure` tests now mint a throwaway self-signed `localhost` keypair at runtime via `openssl` in `setUpAll` (temp dir, cleaned up in `tearDownAll`). `openssl` is already required by `scripts/prepare_pinning_certs.sh`, so there's no new dependency and `dart test` works with no extra steps.

Verified locally: `dart test test/pinning_test.dart` and `test/stacktrace_test.dart` pass, `dart format --set-exit-if-changed` is clean, and `dart analyze` reports no issues. I also refreshed the PR description, which had gone stale after the earlier min-SDK and `DioMixin` reverts.

PR #2514, opened 2026-05-14
feat(dio): propagate custom exception types from interceptors (#1950)

### LINKED ISSUE #1950: Developer should throw custom exception in interceptor
### Request Statement

When i write a interceptor class extends Interceptor, in some case, i wish throw some exception to tell the developer some thing went wrong. so i throw a custom exception in override function 'onResponse', but this exception was catched by DioException, and this exception can not tell me more infomation about this error. 

image
image

So, i wish this will be supported

### Solution Brainstorm

_No response_

realmeylisdev:
## Summary

Adds an opt-in mechanism for developers to propagate custom exception types from interceptors back to the caller of `dio.get/post/...`, so that `try { … } on MyException catch (e)` matches at the await boundary instead of always seeing `DioException`.

Resolves #1950.

Continuation of prior work in #2484 (closed because the contributor's fork was deleted, not for design reasons). API shape follows @kuhnroyal's suggestion in https://github.com/cfug/dio/issues/1950#issuecomment-2685591982. Original issue filed by @WutangNailao.

## Reproducing the original problem (without this PR)

```dart

class MyApiException implements Exception {
  MyApiException(this.code);
  final String code;
}

Future<void> main() async {
  final dio = Dio()
    ..interceptors.add(InterceptorsWrapper(
      onResponse: (response, handler) {
        // The user wants to throw their own exception type here.
        throw MyApiException('unauthorized');
      },
    ));

  try {
    await dio.get('https://example.com/');
  } on MyApiException catch (e) {
    print('caught MyApiException: \${e.code}');     // <-- never reached
  } on DioException catch (e) {
    print('caught DioException, inner: \${e.error}'); // <-- always reached
  }
}
```

**Before this PR:** `on MyApiException` never matches. The exception is always wrapped as `DioException`; the original is accessible only via `e.error`.

**After this PR (opt-in):**

```dart
dio.interceptors.add(InterceptorsWrapper(
  onResponse: (response, handler) {
    handler.rejectCustomError(
      MyApiException('unauthorized'),
      response.requestOptions,
    );
  },
));

try {
  await dio.get('https://example.com/');
} on MyApiException catch (e) {
  print('caught MyApiException: \${e.code}');     // <-- now matches
}
```

The equivalent throw-based form also works:

```dart
throw DioException.custom(
  MyApiException('unauthorized'),
  requestOptions: response.requestOptions,
);
```

## Changes

- `DioException.custom(...)` factory and instance flag `propagateInnerError` (default `false`).
- `rejectCustomError(...)` helper on `RequestInterceptorHandler`, `ResponseInterceptorHandler`, and `ErrorInterceptorHandler`.
- Single unwrap point in `DioMixin.fetch()`'s final catch using `Error.throwWithStackTrace` to preserve the throw-site stack.
- `copyWith` carries the flag (so downstream `onError` interceptors can enrich without losing the marker).
- `dio/README.md`: new "Throwing custom exceptions from interceptors" subsection + `propagateInnerError` field listing.
- `dio/CHANGELOG.md`: Unreleased entry.

## Backwards compatibility

**Default behavior unchanged.** Every existing `on DioException catch (e)` keeps working — the new behavior is strictly opt-in via the new factory or helper. A regression sentinel test (`unmarked DioException(error: customEx) still wraps`) verifies this.

Compatibility Policy unaffected (no SDK bump). `Error.throwWithStackTrace` is in Dart 2.16; dio's floor is 2.18.

## Why this design

- **One funnel.** All interceptor errors converge at `DioMixin.fetch()`'s final `catch`; unwrap there is sufficient *and* safe (transport errors created by `_dispatchRequest` never set `propagateInnerError`).
- **Marker survives by virtue of `assureDioException`'s short-circuit** (`if (error is DioException) return error;`). No need to teach `errorInterceptorWrapper`, `QueuedInterceptor._handleRequest/Response/Error`, or `_dispatchRequest` about the marker — they already pass through DioException unchanged.
- **No interaction with #2499** (the in-flight async-zone fix). Disjoint code regions; either order of merge is fine.

## Test plan

Added 9 tests in `dio/test/interceptor_test.dart` under group `Custom exception unwrapping (#1950)`:

1. `handler.rejectCustomError` from `onResponse` propagates inner type
2. `handler.rejectCustomError` from `onRequest` propagates inner type
3. `handler.rejectCustomError` from `onError` propagates inner type
4. `throw DioException.custom(...)` directly from interceptor propagates inner type
5. **Regression sentinel**: unmarked `DioException(error: customEx)` still wraps as `DioException`
6. `copyWith` preserves `propagateInnerError` through an `onError` chain (with `callFollowingErrorInterceptor: true`)
7. `rejectCustomError` works inside `QueuedInterceptor`
8. `DioException.custom` factory sets `propagateInnerError: true`
9. `DioException` default constructor leaves `propagateInnerError: false`

Verified locally:

- `melos run format` (CI-strict): clean
- `dart analyze --fatal-infos` on `dio` package: no issues
- Full `dio` VM test suite: all 38 interceptor tests + all 4 queued-interceptor tests pass; broader suite green except for `test/test_suite_test.dart: download delete on cancel` which is a pre-existing flake against external `httpbun.com` (passes on isolated re-run; unrelated to this change).

## Files changed

- `dio/lib/src/dio_exception.dart` — flag, factory, copyWith
- `dio/lib/src/interceptor.dart` — three `rejectCustomError` helpers
- `dio/lib/src/dio_mixin.dart` — funnel unwrap
- `dio/test/interceptor_test.dart` — 9 new tests + helper class
- `dio/CHANGELOG.md` — Unreleased entry
- `dio/README.md` — docs subsection + `propagateInnerError` field

### DIFF
--- a/dio/lib/src/dio_exception.dart
+++ b/dio/lib/src/dio_exception.dart
@@ -74,6 +74,7 @@ class DioException implements Exception {
     this.error,
     StackTrace? stackTrace,
     this.message,
+    this.propagateInnerError = false,
   }) : stackTrace = identical(stackTrace, StackTrace.empty)
             ? requestOptions.sourceStackTrace ?? StackTrace.current
             : stackTrace ??
@@ -183,6 +184,58 @@ class DioException implements Exception {
         error: error,
       );
 
+  /// Wraps a custom [error] and signals the request pipeline to rethrow
+  /// [error] verbatim (preserving its runtime type) to the caller of the
+  /// request method, instead of throwing this `DioException` wrapper.
+  ///
+  /// Use this from inside an [Interceptor] when you want callers to be
+  /// able to `try { await dio.get(...) } on MyException catch (e)` and
+  /// match by the original exception type.
+  ///
+  /// Example:
+  /// ```dart
+  /// dio.interceptors.add(InterceptorsWrapper(
+  ///   onResponse: (response, handler) {
+  ///     if (response.statusCode == 401) {
+  ///       handler.reject(DioException.custom(
+  ///         UnauthorizedException(),
+  ///         requestOptions: response.requestOptions,
+  ///       ));
+  ///       return;
+  ///     }
+  ///     handler.next(response);
+  ///   },
+  /// ));
+  ///
+  /// try {
+  ///   await dio.get('/protected');
+  /// } on UnauthorizedException catch (_) {
+  ///   // Matches.
+  /// }
+  /// ```
+  ///
+  /// Internally, the marker survives traversal through subsequent
+  /// `onError` interceptors and through `QueuedInterceptor`s. The
+  /// inner [error] is only thrown at the final boundary of `Dio.fetch`.
+  ///
+  /// Resolves https://github.com/cfug/dio/issues/1950.
+  factory DioException.custom(
+    Object error, {
+    required RequestOptions requestOptions,
+    StackTrace? stackTrace,
+    Response? response,
+    String? message,
+  }) =>
+      DioException(
+        type: DioExceptionType.unknown,
+        requestOptions: requestOptions,
+        response: response,
+        error: error,
+        stackTrace: stackTrace,
+        message: message,
+        propagateInnerError: true,
+      );
+
   /// The request info for the request that throws exception.
   ///
   /// The info can be empty (e.g. `uri` equals to "")
@@ -206,6 +259,18 @@ class DioException implements Exception {
   /// The error message that throws a [DioException].
   final String? message;
 
+  /// When `true`, signals that [error] should be rethrown verbatim by
+  /// the request pipeline at the final boundary of `Dio.fetch`,
+  /// preserving its runtime type for the caller.
+  ///
+  /// Set automatically by [DioException.custom] and by the
+  /// `rejectCustomError(...)` helpers on interceptor handlers. Defaults
+  /// to `false` for all other constructors and factories — existing
+  /// behavior is unchanged.
+  ///
+  /// Resolves https://github.com/cfug/dio/issues/1950.
+  final bool propagateInnerError;
+
   /// Users can customize the content of [toString] when thrown.
   static DioExceptionReadableStringBuilder readableStringBuilder =
       defaultDioExceptionReadableStringBuilder;
@@ -222,6 +287,7 @@ class DioException implements Exception {
     Object? error,
     StackTrace? stackTrace,
     String? message,
+    bool? propagateInnerError,
   }) {
     return DioException(
       requestOptions: requestOptions ?? this.requestOptions,
@@ -230,6 +296,7 @@ class DioException implements Exception {
       error: error ?? this.error,
       stackTrace: stackTrace ?? this.stackTrace,
       message: message ?? this.message,
+      propagateInnerError: propagateInnerError ?? this.propagateInnerError,
     );
   }
 
--- a/dio/lib/src/dio_exception.dart
+++ b/dio/lib/src/dio_exception.dart
@@ -79,7 +79,7 @@ class DioException implements Exception {
     this.error,
     StackTrace? stackTrace,
     this.message,
-    this.propagateInnerError = false,
+    this.isCustom = false,
   }) : stackTrace = identical(stackTrace, StackTrace.empty)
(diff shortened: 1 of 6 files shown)


kuhnroyal:
Thanks! Wondered what happened to the last PR. I only have a phone for the next 10 days, will try to review afterwards.

realmeylisdev:
Old fork got deleted on my end, which auto-closed it — no design issues, just sat with only a Copilot review. This PR is the same `DioException.custom(...)` shape you suggested in [#1950](https://github.com/cfug/dio/issues/1950#issuecomment-2685591982), with a few refinements.

dio/lib/src/interceptor.dart
```diff
@@ -97,6 +97,26 @@ class RequestInterceptorHandler extends _BaseHandler {
...
+  /// Rejects the request with a [DioException.custom] wrapping [error] and
+  /// signals the pipeline to rethrow [error] verbatim to the caller of the
+  /// request method, preserving its runtime type.
+  ///
+  /// Use when you want callers to `on MyException catch (e)` directly,
+  /// rather than `on DioException catch (e) { final inner = e.error; ... }`.
+  /// See [DioException.custom] for details.
+  ///
+  /// Resolves https://github.com/cfug/dio/issues/1950.
+  void rejectCustomError(
```
kuhnroyal: I would probably go with just `rejectCustom` everywhere

dio/lib/src/dio_exception.dart
```diff
@@ -206,6 +259,18 @@ class DioException implements Exception {
...
+  /// the request pipeline at the final boundary of `Dio.fetch`,
+  /// preserving its runtime type for the caller.
+  ///
+  /// Set automatically by [DioException.custom] and by the
+  /// `rejectCustomError(...)` helpers on interceptor handlers. Defaults
+  /// to `false` for all other constructors and factories — existing
+  /// behavior is unchanged.
+  ///
+  /// Resolves https://github.com/cfug/dio/issues/1950.
+  final bool propagateInnerError;
```
kuhnroyal: `isCustom`?

kuhnroyal (commented):
Looking good, just 2 ideas for some consistency in naming but I would be fine either way.

AlexV525:
(To resolve the current CI failure, update the branch to the latest for recent fixes)

realmeylisdev:
Thanks @AlexV525 — merged latest `main` into the branch (now at the `http2_adapter v2.7.1` tip, via #2522 / #2499). That pulls in the HTTP/2 authority-header fix behind the `Verify packages abilities` failure (`HTTP/2 error: Connection is being forcefully terminated`).

The merge is disjoint from the #1950 custom-exception funnel — `assureDioException`'s `is DioException` short-circuit preserves the `propagateInnerError` marker, so it survives #2499's new async error-zone cleanly. As a bonus that interaction now also propagates the inner type for an async `throw DioException.custom(...)`; I added a regression test (`74b0cc2`) locking that in.

Locally green: `dart analyze --fatal-infos`, `dart format`, and the full `dio` VM suite (including all #1950 tests).

The new workflow runs are showing `action_required` — could you approve them when you get a chance? 🙏

AlexV525:
@realmeylisdev Could you respond to the review comments from @kuhnroyal ?

dio/lib/src/interceptor.dart
realmeylisdev: Done — renamed to `rejectCustom` on all three handlers, with the doc refs, README, and tests updated to match. 👍

dio/lib/src/dio_exception.dart
realmeylisdev: Done — renamed the field (plus the constructor and `copyWith` params) to `isCustom`, matching the `DioException.custom` factory. Tests + README updated.

realmeylisdev:
Thanks @AlexV525 — addressed both of @kuhnroyal's naming suggestions: `rejectCustomError` → `rejectCustom` and `propagateInnerError` → `isCustom`, aligned with the `DioException.custom` factory. Also restored the #1950 CHANGELOG entry that got dropped in an earlier `main` merge. `dart analyze --fatal-infos`, `dart format`, and the interceptor/queued/exception VM suites are green locally.

PR #2518, opened 2026-05-17
fix(dio): start receiveTimeout timer immediately after headers, not on first body byte

zdmitrynsk:
## Problem

`receiveTimeout` is not triggered when a server sends response headers but does not send any body bytes.

In `handleResponseStream`, `watchReceiveTimeout()` is only called inside the `onData` callback. If the server sends headers but then goes silent and does not send any body bytes, `onData` is never called, the timer never starts, and the request hangs indefinitely despite `receiveTimeout` being configured.

This is especially common in networks with Deep Packet Inspection (DPI) and blocking systems — they may allow the TCP connection and response headers through, but block or indefinitely delay the response body.

## Fix

Call `watchReceiveTimeout()` once before `source.listen()` so the timer starts immediately after response headers are received.

A regression test is included that verifies `receiveTimeout` fires even when no body bytes arrive.

### DIFF
--- a/dio/lib/src/response/response_stream_handler.dart
+++ b/dio/lib/src/response/response_stream_handler.dart
@@ -67,6 +67,10 @@ Stream<Uint8List> handleResponseStream(
     });
   }
 
+  // Also start watching before the first chunk to cover the gap between
+  // headers and body (see onData below for per-chunk reset).
+  watchReceiveTimeout();
+
   responseSubscription = source.listen(
     (data) {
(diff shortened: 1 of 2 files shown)


AlexV525:
Please also update the branch with our latest CI fixes, thx

zdmitrynsk:
Done, rebased onto latest main. Looks like CI is waiting for workflow approval when you get a chance. Thanks!

AlexV525 (approved):
👍

PR #2526, opened 2026-05-25
🐛 Fix QueuedInterceptor queue stalling when the active request is cancelled

ultramcu:
## Summary

When a request is cancelled (via `CancelToken`) while it is the **active** task in a `QueuedInterceptor`'s internal queue — i.e. its `onRequest`/`onResponse`/`onError` callback is still running an async step and has **not yet** called `handler.next`/`resolve`/`reject` — the queue stalls permanently. Every subsequent request routed through the same `QueuedInterceptor`, **including requests with no cancel token**, then hangs forever.

This hits a very common, real-world setup: a `QueuedInterceptor` that refreshes an auth token in an async `onRequest` (the canonical reason to use a *queued* interceptor), where the user cancels the in-flight request.

## Root cause

`QueuedInterceptor._handleQueue` only advances the queue from inside the handler's `next`/`resolve`/`reject` (via `_processNextInQueue`). Cancellation completes the **request** future (through `listenCancelForAsyncTask`) without ever completing the queued **handler**, so the active task is never released, `processing` stays `true`, and no further task is dequeued.

## Fix

Each task now arms a one-shot cancel hook — registered **only while it is the active task**, so a queued task is never advanced out of turn — that releases the queue slot if the request is cancelled before the handler completes. Queue advancement is made **idempotent** (an `advanced` guard) so a normal completion and a cancellation can't double-advance. No public API change; behaviour for non-cancelled requests is unchanged.

## Tests

Added two regression tests in `queued_interceptor_test.dart` (request queue + response queue): they cancel the active task mid-callback and assert that an independent, **un-cancelled** request still completes promptly. Both **fail (time out / "queue stalled") on `main`** and pass with this change.

- Full `dart test` suite: **221 passed, 6 network-gated skips, 0 failures**
- `dart format` and `dart analyze`: clean
- No public API change; `CHANGELOG.md` updated.

### DIFF
--- a/dio/lib/src/interceptor.dart
+++ b/dio/lib/src/interceptor.dart
@@ -443,20 +443,67 @@ class QueuedInterceptor extends Interceptor {
     void Function(T, V) callback,
     void Function(Object, V) onError,
   ) {
-    final task = _InterceptorParams<T, V>(data, handler);
-    task.handler._processNextInQueue = () {
-      if (taskQueue.queue.isNotEmpty) {
-        final next = taskQueue.queue.removeFirst();
-        assert(next.handler._processNextInQueue != null);
-        callback(next.data, next.handler);
-      } else {
-        taskQueue.processing = false;
+    // Runs [task] as the active task and wires up how the queue advances to
+    // the next task once this one is done.
+    void runTask(_InterceptorParams<T, V> task) {
+      // Hold the task in a clearable reference so that the `whenCancel`
+      // closure registered below — which cannot be detached from the
+      // [CancelToken]'s internal completer — stops pinning the request
+      // payload ([RequestOptions]/[Response]/[DioException] and its handler)
+      // once the queue has advanced. The closure observes `taskRef == null`
+      // and then holds only a null reference, so for a [CancelToken] shared
+      // across many requests (a supported usage pattern documented on
+      // [CancelToken]) per-task allocations are released as each task
+      // completes, instead of accumulating until the token is cancelled.
+      //
+      // A `WeakReference` is intentionally NOT used here: while a task is the
+      // active one, this cell is its only strong reference (the queue has
+      // already removed it and the pending callback captured `task.data`/
+      // `task.handler`, not the wrapper). A weak reference could be collected
+      // mid-flight, so a later cancellation would see a null target, skip
+      // `advance()`, and never release the queue slot — reintroducing the stall
+      // this guards against. The clearable cell keeps the task reachable for
+      // exactly as long as it is active and releases it deterministically when
+      // `advance()` runs.
+      _InterceptorParams<T, V>? taskRef = task;
+
+      // Advancing the queue must happen exactly once per task: either when the
+      // handler is completed (next/resolve/reject) or when the request is
+      // cancelled before the handler is ever called. `advanced` guards against
+      // both paths firing for the same task.
+      bool advanced = false;
+      void advance() {
+        if (advanced) {
+          return;
+        }
+        advanced = true;
+        taskRef = null;
+        if (taskQueue.queue.isNotEmpty) {
+          runTask(taskQueue.queue.removeFirst());
+        } else {
+          taskQueue.processing = false;
+        }
       }
-    };
-    taskQueue.queue.add(task);
-    if (!taskQueue.processing) {
-      taskQueue.processing = true;
-      final task = taskQueue.queue.removeFirst();
+
+      task.handler._processNextInQueue = advance;
+
+      // If the request is cancelled while the interceptor callback is still
+      // pending — i.e. it never calls next/resolve/reject, e.g. an async
+      // `onRequest` awaiting a token refresh that gets cancelled — the
+      // handler's completer would never complete. Since the queue only
+      // advances through the handler, the active slot would never be released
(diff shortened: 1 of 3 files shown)


AlexV525:
Thanks for the PR! Requesting our agent to review...

ultramcu:
Thanks for the review @AlexV525 + Copilot — both findings were valid and have been addressed in [`1cb445f`](https://github.com/cfug/dio/pull/2526/commits/1cb445f) (amended onto the existing commit so the PR stays one commit).

### 1. Memory leak (interceptor.dart) — fixed

The `whenCancel.then(...)` listener can't be detached from the `CancelToken`'s internal completer, so the closure was indeed retaining the per-task payload (`RequestOptions`/`Response`/`DioException` + handler) until the token was cancelled or GC'd — exactly as Copilot described, and impactful for the long-lived shared-token usage pattern `CancelToken` documents.

Fix: a clearable nullable cell `_InterceptorParams<T, V>? taskRef = task;` is captured by the cancel closure instead of `task` directly. `advance()` nulls the cell, so once the queue moves on, the closure pins only `null` (Dart closures capture variables, not initial values).

Verified empirically with a Finalizer-instrumented harness (5 tasks × 16 MB payload each, with a long-lived `whenCancel`-like Future holding the listener):
- Before fix: 0/5 payloads collected
- After fix: 5/5 collected

The cancel-before-completion path still triggers `advance()` correctly; `advanced` + `taskRef == null` together guarantee single-fire.

### 2. `expect(futureA, throwsA(...))` without await (queued_interceptor_test.dart) — fixed

Confirmed: reverting one test to the previous form reproduced the unhandled-async-error leak (`[E]` marker on a passing test).

Both tests now use the "capture and await separately" form Copilot suggested:

```dart
final aRejected = expectLater(
  futureA,
  throwsA(isA<DioException>()
      .having((e) => e.type, 'type', DioExceptionType.cancel)),
);
// ... cancellation + futureB assertions ...
await aRejected;
```

I went with the up-front registration (rather than `await expectLater(...)` after `tokenA.cancel(...)`) so the error listener provably attaches before the cancel propagates — no microtask-timing window.

### CI
- `dart analyze` — clean on changed files
- `dart test test/queued_interceptor_test.dart` — 6/6 pass
- `dart test` — 221 pass, 6 skipped, 1 flake (`test_suite_test.dart`: `send progress` → HTTP 502 from `httpbun.com`; passes in isolation, unrelated to this PR)

Ready for another look whenever you have time. 🙏

dio/lib/src/interceptor.dart
```diff
@@ -443,20 +443,57 @@ class QueuedInterceptor extends Interceptor {
...
+      // Hold the task in a clearable reference so that the `whenCancel`
+      // closure registered below — which cannot be detached from the
+      // [CancelToken]'s internal completer — stops pinning the request
+      // payload ([RequestOptions]/[Response]/[DioException] and its handler)
+      // once the queue has advanced. The closure observes `taskRef == null`
+      // and then holds only a null reference, so for a [CancelToken] shared
+      // across many requests (a supported usage pattern documented on
+      // [CancelToken]) per-task allocations are released as each task
+      // completes, instead of accumulating until the token is cancelled.
+      _InterceptorParams<T, V>? taskRef = task;
```
AlexV525, line 468: I realized Dart introduced `WeakReference` since 2.17, and we are supporting Dart 2.18+. Is it possible that we use the `WeakReference` at this place?

ultramcu: Good one to weigh — I looked at `WeakReference` here and I think the clearable cell is actually the safer fit for this spot. `_InterceptorParams` holds the request payload (`data`) and handler, and while a task is the *active* in-flight one the only strong reference to that wrapper is `taskRef` itself: `runTask`'s frame is gone, the queue already `removeFirst()`'d it, and the pending callback captured `task.data`/`task.handler`, not the wrapper.

If `taskRef` were a `WeakReference<_InterceptorParams>`, the wrapper (with its large payload) could be GC'd *while the task is still active*; a later `token.cancel()` would then see `target == null`, skip `advance()`, and never release the queue slot — intermittently reintroducing the exact stall this PR fixes, in a GC-timing-dependent way.

The clearable cell gives the lifetime we want: strongly reachable for exactly as long as the task is active (so cancel can always advance), then nulled the instant `advance()` runs — a deterministic release rather than waiting for the next GC. So it stays both leak-free and stall-safe.

That said it's your call — happy to switch to `WeakReference` if you'd still prefer it for consistency; just wanted to flag the trade-off. Rebased onto latest `main` to clear the CHANGELOG conflict.

AlexV525: That's enough detail to clear my concerns! I think we can add a small comment so we won't fall into the same consideration next time. Thanks and please make another rebase before merging

ultramcu:
Done — added a code comment at the `taskRef` declaration explaining why a `WeakReference` isn't used here (it could be collected mid-flight, so cancellation would skip `advance()` and re-introduce the stall), and rebased onto latest `main`. `dart analyze` clean and the QueuedInterceptor tests pass. Should be ready to merge — thanks for the review! 🙏

AlexV525 (approved):

PR #2305, opened 2024-10-03
Fix FormData clone boundary inconsistency issue

helloliuyiming:
Resolves https://github.com/cfug/dio/issues/2303

## Fix boundary inconsistency in FormData.clone method

Description:

This PR addresses an issue where cloning FormData objects resulted in different boundaries each time. The problem was caused by allocating a new boundary for each clone without any boundary reuse logic.

Changes made:
- Added an `initBoundary` parameter to the FormData constructor.
- Implemented boundary reuse logic when cloning FormData objects.
- Updated documentation to explain how `initBoundary` affects `boundaryName`.

Technical details:
- The `FormData` constructor now accepts an optional `initBoundary` parameter.
- When `initBoundary` is provided, it overrides the `boundaryName` configuration.
- The cloning process now reuses the existing boundary when appropriate.

Impact:
This fix ensures consistent behavior when cloning FormData objects and improves efficiency by reusing boundaries where possible. It should resolve issues related to unexpected boundary changes during FormData manipulation.

Testing:
- Added unit tests to verify boundary consistency across cloned FormData objects.
- Manually tested with various use cases to ensure backwards compatibility.

Note to reviewers:
Please pay special attention to the boundary reuse logic and any potential side effects on existing FormData usage.

### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [x] I have added the required tests to prove the fix/feature I'm adding
- [x] I have updated the documentation (if necessary)
- [x] I have run the tests without failures
- [x] I have updated the `CHANGELOG.md` in the corresponding package

PR #2531, opened 2026-05-29
🐛 Fix FormData.clone() dropping boundaryName and camelCaseContentDisposition

ultramcu:
Follow-on to merged #2305 and merged #2008. `FormData.clone()` still drops two of the original constructor options, so a retried multipart request silently re-serialises with the defaults.

### 1. Bug

`clone()` builds the new instance via the no-arg `FormData()` constructor:

```dart
FormData clone() {
  final clone = FormData();           // ← resets options to defaults
  clone._boundary = _boundary;        // (added by #2305)
  ...
}
```

That resets two options that the original may have set:

| Option | Added in | Symptom on a cloned (retried) request |
| --- | --- | --- |
| `boundaryName` | original API | `clone.boundaryName` returns the dio default, not the user's label |
| `camelCaseContentDisposition` | #2008 | wire bytes contain `content-disposition: form-data; …` instead of `Content-Disposition: …` |

The header-casing regression is the user-visible one: any server that matches `Content-Disposition` case-sensitively (some signing gateways, some upload validators) accepts the first attempt and rejects the retry.

### 2. Fix

Forward both options through `clone()`:

```dart
FormData clone() {
  final clone = FormData(
    boundaryName: boundaryName,
    camelCaseContentDisposition: camelCaseContentDisposition,
  );
  clone._boundary = _boundary;
  clone.fields.addAll(fields);
  for (final file in files) {
    clone.files.add(MapEntry(file.key, file.value.clone()));
  }
  return clone;
}
```

One-line behavioural change, no new public API.

### 3. Test

New regression test in `dio/test/formdata_test.dart` (`clone() preserves boundaryName and camelCaseContentDisposition`):

- builds a `FormData` with `boundaryName: '--custom-boundary'` and `camelCaseContentDisposition: true`,
- clones it,
- asserts both getters round-trip,
- decodes `readAsBytes()` on the clone and asserts the body contains `Content-Disposition: form-data; name="name"` and **does not** contain the lower-case `content-disposition:` form.

Fail-before / pass-after verified locally:

- **before the fix**: the new test fails on the wire-body assertions (clone emits `content-disposition:` lower-case).
- **after the fix**: full `dio/test/formdata_test.dart` suite green (8/8).

### 4. CI

- `dart format --output=none --set-exit-if-changed .` on `dio/` — clean (0 of 63 files changed).
- `dart analyze` on `dio/lib/src/form_data.dart` and `dio/test/formdata_test.dart` — no issues.
- Full `dart test` on the `dio/` package — 220 passed / 6 network-gated skipped / 0 failed.

### DIFF
--- a/dio/lib/src/form_data.dart
+++ b/dio/lib/src/form_data.dart
@@ -209,7 +209,15 @@ class FormData {
 
   // Convenience method to clone finalized FormData when retrying requests.
   FormData clone() {
-    final clone = FormData();
+    // Forward the constructor options so that the clone produces a wire
+    // representation equivalent to the original — most notably the
+    // `Content-Disposition` header casing chosen via
+    // [camelCaseContentDisposition]. Without this, a retried multipart
+    // request would silently downgrade to the default lowercase header.
+    final clone = FormData(
+      boundaryName: boundaryName,
+      camelCaseContentDisposition: camelCaseContentDisposition,
+    );
     clone._boundary = _boundary;
     clone.fields.addAll(fields);
     for (final file in files) {
--- a/dio/lib/src/form_data.dart
+++ b/dio/lib/src/form_data.dart
@@ -207,13 +207,19 @@ class FormData {
     );
   }
 
-  // Convenience method to clone finalized FormData when retrying requests.
+  /// Clones this finalized [FormData] so it can be re-sent, for example when
+  /// a request is retried.
+  ///
+  /// The clone forwards the constructor options — [boundaryName] and
+  /// [camelCaseContentDisposition] — so it yields a wire representation
+  /// equivalent to the original, most notably the `Content-Disposition`
+  /// header casing. Without this, a retried multipart request would silently
+  /// downgrade to the default lowercase header.
+  ///
+  /// [files] are deep-cloned, while [fields] are copied as immutable
+  /// key/value entries into a new list. Updating this instance's [fields]
+  /// after calling [clone] does not propagate to the returned clone.
   FormData clone() {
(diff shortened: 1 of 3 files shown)


dio/lib/src/form_data.dart
```diff
@@ -209,7 +209,15 @@ class FormData {
 
   // Convenience method to clone finalized FormData when retrying requests.
   FormData clone() {
-    final clone = FormData();
+    // Forward the constructor options so that the clone produces a wire
```
AlexV525: We can explain the behavior by writing comments on this method instead of this inner comment. Also explain fields are not being updated through `clone()`

AlexV525 (commented):
Generally LGTM with one nit about comment

ultramcu:
Thanks for the review! Addressed the nit: moved the explanation onto the `clone()` doc comment (instead of the inner comment), and documented the copy semantics — `files` are deep-cloned while `fields` are copied as immutable key/value entries, and updating `fields` after `clone()` does not propagate to the returned clone.

Also rebased onto `main` to resolve the CHANGELOG conflict, so the PR is mergeable again. `dart analyze` is clean and the clone test passes.

AlexV525 (approved):

AlexV525:
The branch still needs an update :)

ultramcu:
Rebased onto latest `main` again (kept both the `transformTimeout` and this `FormData.clone()` entry in the CHANGELOG). `dart analyze` clean and the FormData tests pass — should be mergeable now. Thanks for the approve! 🙏

Issue #2405, opened 2025-05-27
Bug：HttpException: Connection closed before full header was received


Clearlove2015:
### Package

dio

### Version

5.7.0

### Operating-System

Windows

### Adapter

Default Dio

### Output of `flutter doctor -v`

```shell
[√] Flutter (Channel stable, 3.29.0, on Microsoft Windows [版本 10.0.26100.4061], locale zh-CN) [622ms]
    • Flutter version 3.29.0 on channel stable at D:\Flutter\Android\flutter-sdk\flutter_windows_3.29.0-stable
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 35c388afb5 (4 months ago), 2025-02-10 12:48:41 -0800
    • Engine revision f73bfc4522
    • Dart version 3.7.0
    • DevTools version 2.42.2

[√] Windows Version (Windows 11 or higher, 24H2, 2009) [2.9s]

[√] Android toolchain - develop for Android devices (Android SDK version 35.0.1) [3.8s]
    • Android SDK at D:\Flutter\Android\android-sdk
    • Platform android-36, build-tools 35.0.1
    • ANDROID_HOME = D:\Flutter\Android\android-sdk
    • Java binary at: D:\Flutter\Android\android-studio-narwhal-canary10\jbr\bin\java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.6+-13391695-b895.109)
    • All Android licenses accepted.

[√] Chrome - develop for the web [171ms]
    • Chrome at C:\Users\Flutter\AppData\Local\Google\Chrome\Application\chrome.exe

[√] Visual Studio - develop Windows apps (Visual Studio Community 2022 17.13.7) [170ms]
    • Visual Studio at D:\Flutter\Microsoft Visual Studio\2022\Community
    • Visual Studio Community 2022 version 17.13.36105.23
    • Windows 10 SDK version 10.0.22621.0

[√] Android Studio (version 2024.3.2) [29ms]
    • Android Studio at D:\Flutter\Android\android-studio-meerkat-feature-drop-canary3
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart
[...]

PR #2533, opened 2026-05-31
fix: classify incomplete response header as connection error

a1573595:
Fixes the Dio-side error classification for #2405.

When `dart:io` fails with `HttpException: Connection closed before full header was received`, Dio currently wraps it as `DioExceptionType.unknown`. This happens while `IOHttpClientAdapter` is awaiting `HttpClientRequest.close()`, before a `HttpClientResponse` is available.

That phase is still part of establishing/receiving the response headers. Since no response headers or status code are available, this is closer to a connection-level failure than an unknown application error. This PR maps that specific `HttpException` to `DioExceptionType.connectionError` while preserving the original `HttpException` in `DioException.error`.

### Why this strategy

This intentionally does not add automatic retry.

The reported user-visible symptom is a failed download, but retrying inside `IOHttpClientAdapter` would be a larger behavioral change:
- requests may be non-idempotent;
- request bodies/streams may not be replayable;
- download progress, cancellation, and partial file handling would need separate policy decisions;
- users may already have retry behavior in interceptors or higher-level download managers.

So this PR keeps the fix narrowly scoped to Dio's error classification. It gives callers a stable `connectionError` signal so application-level retry policies can decide whether and how to retry.

- [x] I have updated this branch with the latest `main` branch to avoid conflicts

Added a regression test that simulates `HttpClientRequest.close()` throwing:

`HttpException: Connection closed before full header was received`

The test goes through `dio.download()` to match the reported scenario and verifies:
- the thrown error is `DioExceptionType.connectionError`;
- the original `HttpException` is preserved;
- stack trace behavior remains covered.

### DIFF
--- a/dio/lib/src/adapters/io_adapter.dart
+++ b/dio/lib/src/adapters/io_adapter.dart
@@ -172,7 +172,22 @@ class IOHttpClientAdapter implements HttpClientAdapter {
         },
       );
     }
-    final responseStream = await future;
+    late final HttpClientResponse responseStream;
+    try {
+      responseStream = await future;
+    } on HttpException catch (e, s) {
+      if (e.message.contains(
+        'Connection closed before full header was received',
+      )) {
+        throw DioException.connectionError(
+          requestOptions: options,
+          reason: e.message,
+          error: e,
+          stackTrace: s,
+        );
+      }
+      rethrow;
+    }
 
     if (validateCertificate != null) {
       final host = options.uri.host;
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -6,6 +6,7 @@ See the [Migration Guide][] for the complete breaking changes list.**
 ## Unreleased
 
 - Fix request hanging indefinitely when async interceptor callbacks throw without calling the handler.
+- Fix `HttpException: Connection closed before full header was received` being reported as `DioExceptionType.unknown`.
 
 ## 5.9.2
 
--- a/dio/lib/src/dio_exception.dart
+++ b/dio/lib/src/dio_exception.dart
@@ -173,6 +173,7 @@ class DioException implements Exception {
     required RequestOptions requestOptions,
     required String reason,
     Object? error,
+    StackTrace? stackTrace,
   }) =>
       DioException(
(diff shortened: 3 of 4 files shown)


dio/lib/src/adapters/io_adapter.dart
```diff
@@ -172,7 +172,21 @@ class IOHttpClientAdapter implements HttpClientAdapter {
         },
       );
     }
-    final responseStream = await future;
+    late final HttpClientResponse responseStream;
+    try {
+      responseStream = await future;
+    } on HttpException catch (e) {
```
AlexV525: Should we also passthrough the original stack trace when catching?

AlexV525:
Could this also happen with other adapters?

a1573595:
> Could this also happen with other adapters?

Good question, I'll look into it.

a1573595:
> Could this also happen with other adapters?

I checked the other adapters.

`dio_web_adapter` uses XHR, so it will not produce this dart:io `HttpException`. Its network-layer `xhr.onError` path is already mapped to `DioExceptionType.connectionError`.

`dio_http2_adapter` does not go through `HttpClientRequest.close()`. Its own connection setup handles `SocketException` as `connectionError`, and response stream errors are propagated from the HTTP/2 stream. If HTTP/2 is not supported and it falls back to `IOHttpClientAdapter`, it will use this PR's fix.

`native_dio_adapter` uses the `package:http` conversion layer for Cronet/Cupertino on Android/iOS/macOS, so it also does not produce this dart:io message. On unsupported platforms it falls back to `HttpClientAdapter()`, which resolves to the IO adapter and therefore uses this fix.

So I don't think the same dart:io-specific fix should be duplicated into the other adapters.

dio/lib/src/adapters/io_adapter.dart
a1573595: Fixed. This path now passes through the stack trace captured by the `HttpException` catch clause.

I also updated the regression test to verify that `DioException.stackTrace` preserves the original stack trace from `request.close()`.

AlexV525 (approved):

PR #2534, opened 2026-06-01
Support web downloads in browser adapter

### LINKED ISSUE #2408: Cannot download the file in the browser
### Package

dio

### Version

5.8.0+1

### Operating-System

Web

### Adapter

Default Dio

```shell

```

### Dart Version

_No response_

### Steps to Reproduce

1. path_provider does not support web.
2. The second parameter is required.
Image

### Expected Result

You can download the file in the browser.

### Actual Result

Cannot download the file in the browser

a1573595:
Fixes #2408

This PR adds Web support for `Dio.download` in `dio_web_adapter`.

Previously, `DioForBrowser.download` always threw `UnsupportedError`, so browser users could not call `Dio.download` at all. This change keeps the existing public API and implements browser downloads by fetching the response bytes through Dio, creating a `Blob` object URL, and triggering a hidden anchor click.

On Web, `savePath` is interpreted as the browser suggested filename, not as a filesystem path.

Implementation details:

- `DioForBrowser.download` now supports browser downloads through `Blob` + object URL + hidden anchor click.
- The regular Dio request path is preserved, including interceptors, headers, status validation, cancel token, timeout, and progress callbacks.
- `String savePath` is treated as the suggested filename on Web.
- `FutureOr<String> Function(Headers)` save path callbacks are supported after response headers are available.
- `FileAccessMode.append` throws `UnsupportedError` on Web.
- Bad responses and cancellations do not trigger a browser download.
- Object URLs are revoked on success and failure paths.

Web limitations are documented:

- browser decides the actual saved location
- the whole response is loaded into memory before triggering the download
- CORS still applies
- `FileAccessMode.append` is unsupported
- `deleteOnError` is a no-op on Web
- custom `lengthHeader` values are not used for Web progress totals

### Test result

Chrome：

https://github.com/user-attachments/assets/143b5809-1bf3-4aaf-8f11-83ea25245310

Safair：

https://github.com/user-attachments/assets/e9a381c9-cb4a-4fe1-ab84-89f63293b406

Firefox：

https://github.com/user-attachments/assets/ae76d524-2542-49fc-810d-c939d33291ce

### DIFF
--- a/plugins/web_adapter/lib/src/dio_impl.dart
+++ b/plugins/web_adapter/lib/src/dio_impl.dart
@@ -1,6 +1,10 @@
+import 'dart:async';
+import 'dart:typed_data';
+
 import 'package:dio/dio.dart';
 
 import 'adapter_impl.dart';
+import 'download_trigger.dart' as download_trigger;
 
 /// Create the [Dio] instance for Web platforms.
 Dio createDio([BaseOptions? options]) => DioForBrowser(options);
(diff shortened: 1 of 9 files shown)


plugins/web_adapter/lib/src/dio_impl.dart
```diff
@@ -26,9 +30,78 @@ class DioForBrowser with DioMixin implements Dio {
...
+    return _suggestedFilenameFromPath(await savePath(response.headers));
   }
+  return _suggestedFilenameFromPath(savePath as String);
+}
+
+String _suggestedFilenameFromPath(String path) {
+  final normalized = path.replaceAll('\\', '/');
+  final slash = normalized.lastIndexOf('/');
+  final filename = slash == -1 ? normalized : normalized.substring(slash + 1);
+  return filename.isEmpty ? 'download' : filename;
```
AlexV525, line 106: We have `package:path` as a dependency; it should be more precise if we use its helper method to determine the filename.

plugins/web_adapter/lib/src/download_trigger.dart
AlexV525, line 1: Looks like the file needs more code comments.

AlexV525:
How will the browser react when triggered? Will a download task be scheduled by a blob request? I can only see the status update on the webpage itself, as the demonstration video doesn't reflect the browser's behavior. Also, I'd like to know if this has any limitations across Web protocol versions and different browsers.

plugins/web_adapter/lib/src/dio_impl.dart
AlexV525: Ah sorry, the dep is for the main library. Please ignore

plugins/web_adapter/lib/src/download_trigger.dart
a1573595: Updated.

- Added comments around the Blob/object URL/anchor click implementation.
- Expanded `dio_web_adapter` README to clarify that the browser schedules a download from a `blob:` URL, and that the returned `Response` only means Dio fetched the bytes and dispatched the download click.
- Documented browser-controlled limitations: actual file write, filename handling, user prompts, page security policy, CORS, HTTP protocol handling through XHR, and required browser APIs.

a1573595:
> video doesn't reflect the browser's behavior. Also, I'd like to know if this has any

When triggered, Dio first finishes the XHR request as bytes, then creates a `Blob`, an object URL, and clicks a hidden `<a download>` element. That schedules a normal browser download task for the generated `blob:` URL.

I validated this behavior with Chrome, Firefox, and Safari demos. The UI differs per browser: Chrome/Firefox show a clearer download entry/panel, while Safari may only show a subtle toolbar/download indicator, but the file is still saved.

Limitations:
- The original request is still XHR, so CORS applies.
- The full response is held in memory before the Blob download starts.
- Dio can suggest the filename, but the browser decides the final save location/name.
- There is no append/delete-on-error filesystem behavior on Web.
- HTTP/1.1, HTTP/2, or HTTP/3 should not change this behavior; transport is handled by the browser before the Blob URL download is scheduled.

AlexV525:
Coverage info doesn't seem to be valid somehow. Paths are missing from it. cc @kuhnroyal

plugins/web_adapter/lib/src/dio_impl.dart
```diff
@@ -26,9 +30,78 @@ class DioForBrowser with DioMixin implements Dio {
...
+    }
+    return response;
+  }
+}
+
+Future<String> _resolveFilename(dynamic savePath, Response response) async {
+  if (savePath is FutureOr<String> Function(Headers)) {
+    response.headers
+      ..set('redirects', response.redirects.length.toString())
+      ..set('uri', response.realUri.toString());
```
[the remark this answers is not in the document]
a1573595: Done

AlexV525 (approved):
Looks good!

PR #2539, opened 2026-06-11
test: scope coverage sources by package/platform

AlexV525:
## Summary
- keep `only_list_changed_files: true` (changed-files filtering behavior unchanged)
- scope web coverage collection to `dio_web_adapter` on Chrome JS run only
- keep `dart2wasm` rerun without coverage

## Coverage policy after this change
- `dio` package: VM coverage only
- `dio_web_adapter`: Chrome (non-wasm) web coverage
- other packages: VM coverage only

## Why
Web coverage from broad package runs and mixed JS/wasm writes made diff coverage noisy and could produce misleading file changes. Scoping coverage sources makes report output align with package intent while keeping test execution coverage-neutral.

## Validation
- On `dio_web_adapter` context with Chrome+wasm sequence, LCOV is generated (`SF` entries present).
- On `dio` context, Chrome web tests run without creating `coverage/chrome`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -118,7 +118,6 @@ jobs:
         with:
           artifact_download_workflow_names: 'Verify packages abilities,coverage_baseline'
           filename: 'coverage/cobertura.xml'
-          only_list_changed_files: true
       - name: '[Coverage] Write PR number to file'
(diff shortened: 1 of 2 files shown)


AlexV525:
context: in https://github.com/cfug/dio/pull/2534, the code coverage reports 0.00% diff, which is absolutely abnormal.

Issue #2279, opened 2024-08-09
Dio Throws Exception on DELETE Request with 204 Response 


HHLiftBridge:
### Package

dio

### Version

5.5.0

### Operating-System

Web

### Adapter

Default Dio

### Output of `flutter doctor -v`

```shell
[√] Flutter (Channel stable, 3.19.6, on Microsoft Windows [Version 10.0.22631.3880], locale en-US)
    • Flutter version 3.19.6 on channel stable at C:\src\flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 54e66469a9 (4 months ago), 2024-04-17 13:08:03 -0700
    • Engine revision c4cd48e186
    • Dart version 3.3.4
    • DevTools version 2.31.1

[√] Windows Version (Installed version of Windows is version 10 or higher)

[√] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
    • Android SDK at C:\Users\andrew\AppData\Local\Android\sdk
    • Platform android-34, build-tools 32.1.0-rc1
    • Java binary at: C:\Program Files\Android\Android Studio\jbr\bin\java
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-10027231)
    • All Android licenses accepted.

[√] Chrome - develop for the web
    • Chrome at C:\Program Files\Google\Chrome\Application\chrome.exe

[X] Visual Studio - develop Windows apps
    X Visual Studio not installed; this is necessary to develop Windows apps.
      Download at https://visualstudio.microsoft.com/downloads/.
      Please install the "Desktop development with C++" workload, including all of its default components

[√] Android Studio (version 2022.3)
    • Android Studio at C:\Program Files\Android\Android Studio
    • Flutter plugin can be installed from:
       https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
       https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.6+0-b2043.56-10027231)

[√] VS Code (version 1.92.0)
    • VS Code at C:\Users\andrew\AppData\Local\Programs\Microsoft VS Code
    • Flutter extension version 3.94.0

[√] Connected device (3 available)
    • Windows (desktop) • windows • windows-x64    • Microsoft Windows [Version 10.0.22631.3880]
    • Chrome (web)      • chrome  • web-javascript • Google Chrome 126.0.6478.127
[...]

PR #2550, opened 2026-07-05
fix(dio): avoid FormatException on empty JSON body with a custom responseDecoder

tonytonycoder11:
### Problem

`FusedTransformer` is the default `Transformer`. When a request sets a custom `responseDecoder` **and** the server returns an empty body with a JSON content-type (a normal `200`/`204`/`304`), the transformer's slow path runs `jsonDecode('')`, throwing `FormatException: Unexpected end of input`.

`SyncTransformer` and `BackgroundTransformer` handle the identical input gracefully — they guard the empty string and return it (which dio normalizes to `null` for typed JSON responses). Only the default transformer crashes.

This is the residual half of #2279: PR #2285 fixed the empty-body → null behavior for the **fast path** (no custom decoder) but never guarded the custom-decoder slow path.

### Fix

Add an `isNotEmpty` check to the slow-path guard so an empty decoded body falls through to the existing `return decodedResponse;` branch, matching `SyncTransformer` / `BackgroundTransformer`:

```dart
if (isJsonContent && decodedResponse != null && decodedResponse.isNotEmpty) {
  return jsonDecode(decodedResponse);
} else if (customResponseDecoder != null) {
  return decodedResponse;
}
```

Non-empty malformed bodies still throw, exactly as before and as the sibling transformers do (the fix is intentionally scoped to match their `isNotEmpty`-only guard).

- [x] I have updated the documentation (if necessary) — n/a, no public API change

Reproduction (no network needed), calling the default transformer directly:

```dart
String decoder(List<int> b, RequestOptions o, ResponseBody rb) =>
    utf8.decode(b, allowMalformed: true);

await FusedTransformer().transformResponse(
  RequestOptions(responseType: ResponseType.json, responseDecoder: decoder),
  ResponseBody.fromBytes([], 200, headers: {
    Headers.contentTypeHeader: [Headers.jsonContentType],
  }),
);
// before: throws FormatException: Unexpected end of input
// after:  returns '' (consistent with SyncTransformer / BackgroundTransformer)
```

The added regression test (`dio/test/transformer_test.dart` → *"empty JSON body with a custom responseDecoder does not throw"*) fails without the fix and passes with it.


AlexV525 (approved):
Thanks for your contribution!

PR #2549, opened 2026-07-04
📝 Add agent contribution guidelines

AlexV525:
Introduces `AGENTS.md` (+ Simplified Chinese mirror) as this repo's rules for AI-agent-assisted contributions, along with pointers from `CONTRIBUTING` and a PR-template checklist item.

`AGENTS.md` is the de-facto standard filename that agent tools (Claude Code, Codex, Cursor, Devin, etc.) pick up automatically, so agents on external forks get these rules without extra setup.

### What the guidelines cover

1. **Motivation first** — every non-trivial change must answer four questions (what can't be done today, who needs it, why in core rather than an interceptor/adapter/transformer/plugin, cost). `Closes #NNNN` alone does not count as prior discussion — the referenced issue must show maintainer buy-in. One PR = one concern.
2. **Mandatory, effective, non-duplicated tests** — tests must fail without the change; coverage minimum is a floor, not a target; no test-theater for coverage numbers.
3. **Compatibility is sacred** — additive-by-default; deprecations with `Use XXX instead. This will be removed in X.0.0` targeting the next major (X is a placeholder — pick the actual next major from the affected package's `pubspec.yaml`); no SDK-lower-bound bumps outside the Compatibility Policy; watch behavioral breaks. Explicit **sensitive-area** list (SSL/TLS + cert pinning, redirects, cookies, headers, timeout/cancellation/connection pooling, interceptor pipeline, request-body encoding) and **dependency-change** rules (no drive-by bumps; state motivation; narrowest constraint; disclose new transitives).
4. **Understand before you change** — no API guessing; cross-platform pub-cache paths + `PUB_CACHE` env var.
5. **Production quality only** — no `TODO`/mock/simplified-and-called-done code.
6. **When to stop and ask** — general rule for ambiguity and out-of-scope work, distinct from §3's breaking-change escalation.
7. **Repository layout** — Melos mono-repo, per-package SDK lower bounds.
8. **Commits, changelog, and PR hygiene**
   - §8.1 **Branch naming** — `category/ticket-id-or-short-description`
   - §8.2 **Commit format** — gitmoji **or** Conventional Commits (either is acceptable) with an emoji reference table matched to the repo's actual usage
   - §8.3 **AI attribution mandatory** — every contributing agent gets a `Co-Authored-By:` trailer; the PR body discloses agent + stage (design / implementation / tests / review)
   - §8.4 CHANGELOG rules
[...]

PR #2556, opened 2026-07-08
✨ feat: warn about CORS preflight requests on Web

### LINKED ISSUE #2201: Prompt for not simple requests on Web
### Request Statement

People usually wonder why their requests are being treated as CORS. Various of configuration could lead to the result before people can realize, such as `connectTimeout` and `sendTimeout` across platforms.

### Solution Brainstorm

We might find relevant cases to add logs about "You're making a complex Web request which might lead to CORS" or add more messages when specific exceptions were thrown on the Web platform.

CaiJingLong:
## Summary

- Detect when a request is not a CORS "simple request" and will trigger a preflight (OPTIONS) request on the Web platform, then emit a `warningLog` explaining why (non-safelisted method, custom headers, non-simple Content-Type, upload progress listener, or `withCredentials`).
- Enrich the `XMLHttpRequest.onError` error message with CORS guidance when the request was preflighted, so users see an actionable hint instead of a generic network error.
- Add an `enableCORSWarning` opt-out flag (defaults to `true`) so apps that handle CORS correctly can silence the per-request warning. The enriched error message is always emitted regardless of this setting.
- Extract the detection logic into pure functions in `lib/src/cors.dart` (`corsPreflightReason`, `collectCorsPreflightReasons`, `corsEnrichedErrorReason`) so they can be unit-tested on the VM without a browser.

Closes #2201.

### AI attribution

*Implementation, tests, and PR workflow by Devin (GLM-5.2); local review pass by the human operator.*

### Verification

Added 23 unit tests covering `corsPreflightReason` (method / content-type / custom-header combinations), `collectCorsPreflightReasons` (upload-listener, withCredentials, combined cases), and `corsEnrichedErrorReason` (empty / non-empty / multi-reason). `dart analyze --fatal-infos` and `dart test -p vm` pass locally on `dio_web_adapter`.

**Unverified:** The browser-side runtime paths (actual `warningLog` emission, `XMLHttpRequest.onError` enrichment in a real browser) are not exercised locally — they require browser CI. The existing `corsTests` in `dio_test` cover CORS preflight classification behavior but do not assert on the new warning or enriched error message.

### DIFF
--- a/plugins/web_adapter/lib/src/adapter_impl.dart
+++ b/plugins/web_adapter/lib/src/adapter_impl.dart
@@ -8,6 +8,10 @@ import 'package:dio/src/utils.dart';
 import 'package:meta/meta.dart';
 import 'package:web/web.dart' as web;
 
+import 'cors.dart';
+
+export 'cors.dart' show corsPreflightReason;
+
 BrowserHttpClientAdapter createAdapter() => BrowserHttpClientAdapter();
 
 /// The default [HttpClientAdapter] for Web platforms.
@@ -62,6 +66,35 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
 
     final xhrTimeout = (connectTimeout + receiveTimeout).inMilliseconds;
     xhr.timeout = xhrTimeout;
+
+    // Detect configurations that will trigger a CORS preflight request so we
+    // can warn the developer and enrich the eventual error message.
+    // See https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests
+    final preflightReasons = <String>[
+      if (corsPreflightReason(options) case final reason?) reason,
+    ];
+    final willRegisterUploadListener = requestStream != null &&
+        (sendTimeout > Duration.zero || onSendProgress != null);
+    if (willRegisterUploadListener) {
+      preflightReasons.add(
+        'an upload progress listener (sendTimeout or onSendProgress) '
+        'forces a preflight request',
+      );
+    }
+    if (xhr.withCredentials) {
+      preflightReasons.add(
+        'withCredentials is enabled, which requires a CORS preflight request',
+      );
+    }
+    if (preflightReasons.isNotEmpty) {
+      warningLog(
+        'This request is not a CORS "simple request" and will trigger a '
+        'preflight (OPTIONS) request: ${preflightReasons.join('; ')}. '
+        'If the server does not handle CORS preflight, the request will fail. '
+        'See https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests',
+        StackTrace.current,
+      );
+    }
 
     final completer = Completer<ResponseBody>();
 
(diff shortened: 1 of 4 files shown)


plugins/web_adapter/lib/src/adapter_impl.dart
```diff
@@ -8,6 +8,10 @@ import 'package:dio/src/utils.dart';
 import 'package:meta/meta.dart';
 import 'package:web/web.dart' as web;
 
+import 'cors.dart';
+
+export 'cors.dart' show corsPreflightReason;
```
AlexV525: This `export` makes `corsPreflightReason` a permanent public API of `dio_web_adapter`, but the test file already imports it via `package:dio_web_adapter/src/cors.dart` and doesn't need the export. Unless we intentionally want to expose this as a downstream API (with a CHANGELOG note and awareness that future signature changes would be breaking), please remove the `export`. Adding public surface for testing only is a compat trap.

plugins/web_adapter/lib/src/adapter_impl.dart
```diff
@@ -63,6 +67,35 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
     final xhrTimeout = (connectTimeout + receiveTimeout).inMilliseconds;
     xhr.timeout = xhrTimeout;
 
+    // Detect configurations that will trigger a CORS preflight request so we
+    // can warn the developer and enrich the eventual error message.
+    // See https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests
+    final preflightReasons = <String>[
```
AlexV525: The three new branches added by this PR — `willRegisterUploadListener` (`sendTimeout` / `onSendProgress`), `withCredentials`, and the enriched `onError` message construction below — are not exercised by the new tests. All 15 tests cover only `corsPreflightReason` (the pure helper). The coverage bump on this file (19.27% → 22.41%) doesn't show whether these specific lines were hit. Could you either add a browser-side integration case in `dio_test` for these paths, or lift them into pure helpers alongside `corsPreflightReason` so they can be unit-tested too?

plugins/web_adapter/lib/src/adapter_impl.dart
```diff
@@ -63,6 +67,35 @@ class BrowserHttpClientAdapter implements HttpClientAdapter {
...
+        'forces a preflight request',
+      );
+    }
+    if (xhr.withCredentials) {
+      preflightReasons.add(
+        'withCredentials is enabled, which requires a CORS preflight request',
+      );
+    }
+    if (preflightReasons.isNotEmpty) {
+      warningLog(
```
AlexV525, line 91: This will fire on essentially every JSON POST in every dio-on-Web app — including the (very common) case where the backend handles CORS preflight correctly. Would you consider an opt-out (e.g., an option on the adapter, or a setter akin to `debugPrint`), or dropping the log to a lower verbosity when the app has no way to know whether preflight succeeded? A permanent warning per request seems louder than the underlying issue warrants.

AlexV525 (commented):
Thanks for driving this — the extraction of `corsPreflightReason` into a pure, testable helper is a nice touch, the CHANGELOG entry is clean, and `feat/web-cors-preflight-warning` already matches the branch-naming convention that just landed in AGENTS.md. Left three inline comments on the substantive points; a few AGENTS.md formalities below.

### AGENTS.md formalities (fixup, or a follow-up)

- **Commit / PR title should be gitmoji + Conventional** per AGENTS.md §8.2 — `✨ feat: warn about CORS preflight requests on Web`.
- **AI attribution disclosure** per §8.3: Devin's `Co-Authored-By` trailer is present ✓, but if GLM-5.2 also produced code / tests / docs (as noted on #2549), it should have its own trailer. Please also add one line to the PR description stating which agent did which stage — design / implementation / tests / review.
- **Test-plan item 4** — “existing `corsTests` still cover runtime behavior” isn't quite right: those tests predate this PR and can't cover the newly-added warning + enriched-error paths. Either run browser CI and check the box, or replace the note with an honest “browser paths not exercised locally”. (Related: AGENTS.md §8.6 now prefers a prose description of what was actually verified over a boilerplate test-plan checklist — no need to restructure this PR, just heads-up for the next one.)

CaiJingLong:
Thanks for the detailed review — all three inline comments and the AGENTS.md formalities have been addressed in e69c194:

1. **Removed the `export` of `corsPreflightReason`** — the test file now imports `src/cors.dart` directly, and no public API surface was added.

2. **Lifted the untested branches into pure functions** — `collectCorsPreflightReasons` (combines `corsPreflightReason` with the upload-listener and `withCredentials` runtime factors) and `corsEnrichedErrorReason` (builds the enriched `onError` message) are now in `cors.dart` and unit-tested with 8 new tests. The adapter calls these pure functions, so every new branch is covered by VM-testable assertions.

3. **Added `enableCORSWarning` opt-out flag** — defaults to `true`, but apps that handle CORS correctly can set `BrowserHttpClientAdapter(enableCORSWarning: false)`. The enriched error message is always emitted regardless of this setting, so the diagnostic value is preserved even when the warning is silenced.

**AGENTS.md formalities:**
- Commit title now uses gitmoji + Conventional format (`✨`).
- Added `Co-Authored-By: GLM-5.2` trailer and AI attribution line in the PR description.
- Replaced the boilerplate test-plan checklist with an honest verification paragraph + an **Unverified** section for the browser-side runtime paths.

AlexV525 (approved):

PR #2557, opened 2026-07-13
✨ Add `query` and `queryUri` for the HTTP QUERY method (RFC 10008)

CaiJingLong:
## Summary

- Add `Dio.query` and `Dio.queryUri` convenience methods for the HTTP QUERY method defined in [RFC 10008](https://www.rfc-editor.org/rfc/rfc10008).
- The QUERY method is safe and idempotent like `GET`, but allows a request body, which is useful for complex queries whose parameters are too large or too structured to fit in the request URI.
- This is a purely additive change: the new methods follow the same pattern as the existing `get`/`post`/`put`/`patch`/`delete`/`head` convenience methods and delegate to `request`/`requestUri` with `checkOptions('QUERY', options)`. No existing public API is changed, and all adapters already pass the method string through (`HttpClient.openUrl` on native, `XMLHttpRequest.open` on Web), so no adapter changes are required.
- Updated `CHANGELOG.md`, `README.md`, and `README-ZH.md`.

Closes #2548

#### Verification

- `melos run format` — pass
- `melos run analyze` — pass
- `dart test test/test_suite_test.dart` (dio, VM) — the two new `QUERY` tests pass, along with the rest of the `HTTP method` group. A few pre-existing failures in `basic_tests` and `timeout_tests` are environment-specific (DNS resolution / TLS handshake / download throughput on the local machine) and unrelated to this change.

#### AI attribution

*Implementation, tests, and docs by Devin (GLM-5.2).*

Generated with [Devin](https://devin.ai)

PR #2558, opened 2026-07-13
docs: require agents to verify RFC references before citing them

CaiJingLong:
## Summary
- Add a rule in §4 of `AGENTS.md` and `AGENTS-ZH.md` requiring agents to verify any RFC / specification / standard reference at rfc-editor.org or datatracker.ietf.org before citing it in commits, changelog, doc comments, README, or PR descriptions.
- If a reference cannot be verified online, drop the citation and describe the behavior in plain words instead of guessing a number.
- Add a corresponding row to the §9 closure table for unverified or wrong RFC / specification references.

## Motivation
Agents routinely hallucinate RFC numbers or attach the wrong title to a number. A wrong citation in a commit message or changelog is worse than no citation, because it misleads downstream readers and reviewers who trust it. This mirrors the existing "never guess an API" rule in §4, extended to external specifications.

## AI attribution
*Implementation and commit by Devin (GLM-5.2).*

Generated with [Devin](https://devin.ai)


AlexV525:
Supportive of this rule, but I'd suggest broadening the scope from "RFC / specification / standard reference" to **any factual claim written into a commit, changelog, doc comment, or PR description**.

Context (see #2557): an agent wrote the HTTP QUERY RFC number as 9239 when it is actually 10008 — but the same failure mode applies just as well to:

- issue / PR numbers ("fixes #NNNN" pointing at an unrelated issue)
- library / API version numbers, deprecation timelines
- attributed quotes, benchmark figures, CVE identifiers
- platform behavior claims ("iOS X.Y and later…")

RFCs are only the most easily-caught instance. A more general phrasing would be along the lines of "for any external fact, verify the source before writing it down; if you cannot verify, describe the observed behavior in your own words instead of citing a number." The §9 closure table row could then be generalized to "unverified external factual reference."

CaiJingLong:
Good point — done in 5317fd5. The rule is now phrased as "verify every external fact before writing it down", with the RFC lookup kept as a concrete example alongside issue/PR numbers, library/API versions, CVE identifiers, deprecation timelines, benchmark figures, attributed quotes, and platform-behavior claims. The §9 closure-table row is generalized to "unverified or wrong external factual reference".

*Revision by Devin (GLM-5.2) per review feedback on #2558.*

AlexV525 (approved):

PR #2567, opened 2026-07-18
fix(dio): preserve shared Future errors across interceptors

### LINKED ISSUE #2565: Concurrent requests can expose unhandled errors in custom deduplication interceptors
### Package

dio

### Version

5.10.0

### Operating-System

iOS

### Adapter

Default Dio

```shell
[✓] Flutter (Channel stable, 3.44.6, on macOS 26.5.2 25F84 darwin-arm64, locale ru-RU)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.1.0)
[✓] Xcode - develop for iOS and macOS (Xcode 26.6)
[✓] Connected device (6 available)
[✓] Network resources

• No issues found!
```

### Dart Version

3.12.2

#### Description

When several identical Dio requests are started synchronously, a custom
deduplication interceptor may receive an error for the original request before
another request has reached its `onRequest` callback.

This is observable when the underlying `HttpClientAdapter` fails immediately.
The error is delivered through `onError`, while a duplicate request may still be
scheduled to enter the interceptor chain.

If the interceptor completes a shared `Completer` with `completeError` before a
duplicate request subscribes to its future, Dart reports an unhandled
asynchronous error. Depending on the interceptor implementation, a duplicate
request can also be dispatched separately or remain unresolved.

```dart
import 'dart:async';
import 'dart:typed_data';

void main() async {
  final dio = Dio()
    ..interceptors.add(DeduplicationInterceptor())
    ..httpClientAdapter = ImmediateErrorAdapter();

  try {
    await Future.wait([
      dio.get<void>('/resource'),
      dio.get<void>('/resource'),
    ]);
  } on DioException catch (error) {
    print('Caught by Future.wait: ${error.message}');
  }
}

class DeduplicationInterceptor extends Interceptor {
  final pending = <String, Completer<Response<Object?>>>{};

  String requestKey(RequestOptions options) => '${options.method}:${options.path}';

  @override
  void onRequest(RequestOptions options, RequestInterceptorHandler handler) {
    final key = requestKey(options);
    final existing = pending[key];

    if (existing != null) {
      existing.future.then(
        handler.resolve,
        onError: (Object error, StackTrace stackTrace) {
          handler.reject(
            error is DioException
                ? error
                : DioException(
                    requestOptions: options,
                    error: error,
                    stackTrace: stackTrace,
                  ),
          );
        },
      );
      return;
    }

    pending[key] = Completer<Response<Object?>>();
    handler.next(options);
  }

  @override
  void onError(DioException error, ErrorInterceptorHandler handler) {
    final key = requestKey(error.requestOptions);
    final completer = pending.remove(key);

    if (completer != null && !completer.isCompleted) {
      completer.completeError(error);
    }

    handler.next(error);
  }
}

class ImmediateErrorAdapter implements HttpClientAdapter {
  @override
  Future<ResponseBody> fetch(
    RequestOptions options,
    Stream<Uint8List>? requestStream, [
    Future<void>? cancelFuture,
  ]) {
    throw DioException(
      requestOptions: options,
      type: DioExceptionType.badResponse,
      error: 'Immediate adapter failure',
    );
  }

  @override
  void close({bool force = false}) {}
}
```

### Expected Result

A custom interceptor should have a predictable way to safely deduplicate
concurrent requests that fail immediately:

- no unhandled asynchronous errors;
- all requests that join an in-flight operation receive the same error;
- no request remains pending indefinitely;
- request scheduling behavior is documented or remains compatible across Dio
  versions.

## Expected output
Unhandled exception:
DioException [bad response]: null
Error: Immediate adapter failure

### Actual Result

With an immediately failing adapter, an error may be emitted before another
identical request reaches `onRequest`.

The following minimal example reports an unhandled `DioException`:

AlexV525:
Closes #2565

## Problem

Dio 5.10.0 introduced a regression in the interceptor execution model. A custom interceptor that deduplicates concurrent requests by sharing a `Completer` can expose an uncaught `DioException`, leave the duplicate request pending, or dispatch the same request more than once when the adapter fails immediately.

The failure is not caused by request deduplication itself. It is caused by the interaction between deduplication and the per-callback error zones introduced by #2499.

## Root Cause

Interceptor callbacks intentionally have public `void` return types:

```dart
void onRequest(RequestOptions options, RequestInterceptorHandler handler)
```

Dart still allows an async implementation or callback to produce a runtime `Future<void>`, but the public static type does not expose that Future. Before #2499, an async exception thrown before `next`, `resolve`, or `reject` could therefore leave the handler's completer pending forever.

#2499 addressed that hang by invoking every callback inside a newly forked `Zone` with a custom `handleUncaughtError`. If the handler was incomplete, the zone converted the uncaught error into a handler error; if the handler was already complete, it forwarded the error to the parent zone. This preserved the public `void` API and did not await callbacks.

The new zone boundary changed Dart's Future error-delivery semantics. Dart Future errors do not cross between distinct error zones in the same way as ordinary values. The deterministic #2565 sequence is:

```text
Request A: callback runs in error zone A
           creates the shared Completer
           calls handler.next()
           proceeds to the adapter

Request B: callback runs in error zone B
           finds A's Completer
           registers an onError listener on its Future

Adapter:   A fails immediately
           A's onError completes the shared Completer with completeError()

Result:    the Future belongs to error zone A, but B's listener belongs to B
           B's listener is not delivered the shared error
           A's handler is already complete, so Dio forwards the error as uncaught
           B's handler never completes
```

The regression is the per-callback error-zone isolation; the immediate adapter failure is the deterministic timing trigger that exposes it in this reproduction. This fix removes the callback-specific zones created by Dio; it does not make a shared Future safe when application code deliberately creates and listens to it across unrelated external error zones, which remain governed by Dart's normal error-zone rules. Before the implementation, a test-only probe reproduced two `onRequest` callbacks, one adapter fetch, zero joined-error deliveries, one uncaught error, and a timed-out pending request on the base implementation. The committed regression test asserts the repaired behavior after the fix.

### 1. Capture runtime callback Futures without changing the API

A private library typedef, `_InterceptorCallback<T, V>`, returns `Object?`. `Interceptor` now has private virtual `_invokeRequest`, `_invokeResponse`, and `_invokeError` entry points that dynamically invoke the public `on*` method:

```dart
final dynamic callback = onRequest;
return callback(options, handler);
```

This is deliberate:

- A normal `Interceptor` subclass can override `Future<void> onRequest/onResponse/onError`; dynamic invocation retains the runtime Future even though the base API is `void`.
- The public method signatures and public callback typedefs remain unchanged.
- The private entry points are used only by the internal request pipeline.

The wrapper compatibility path is kept explicit. `InterceptorsWrapper` and `QueuedInterceptorsWrapper` still expose their existing public `void on*` methods. The wrapper mixin dynamically invokes the constructor-supplied callback and observes its runtime result there. It no longer uses private mixin overrides that bypass a subclass's public `onRequest`, `onResponse`, or `onError` override. This preserves downstream wrapper-subclass behavior while retaining support for async constructor callbacks.

### 2. Observe, do not await, the returned Future

`_observeInterceptorCallback` is registered synchronously immediately after callback invocation, in the callback's existing zone:

```text
invoke callback
  -> obtain runtime result
  -> if result is a Future, attach a success no-op and an error listener
  -> return handler.future to the interceptor pipeline
```

The callback Future is never awaited and is never used as the pipeline completion signal. Handler completion remains the only normal signal that advances the request, response, or error chain.

When the observed Future fails:

| Callback stage | Handler is incomplete | Handler is already complete |
| --- | --- | --- |
| Request | `handler.reject(assureDioException(error, ..., stackTrace), true)` | Report through the callback's existing zone |
| Response | `handler.reject(assureDioException(error, ..., stackTrace), true)` | Report through the callback's existing zone |
| Error | `handler.next(assureDioException(error, ..., stackTrace))` | Report through the callback's existing zone |

The original `StackTrace` is preserved in all conversions. The error listener consumes the original Future error and handles it through the interceptor handler, so the derived listener Future does not create a second uncaught error.

If the callback has already completed its handler and then its returned Future fails, the error is deliberately sent to the zone captured when the listener was registered. This preserves #2499's late-error visibility for fire-and-forget behavior without allowing Dio to complete a handler a second time. A detached Future that the callback neither returns nor handles remains outside Dio's ownership and is not reclassified as a handler failure. If the callback also never completes its handler, Dio cannot use that detached Future to unblock the request, so the request can still remain pending. This is an explicit ownership boundary rather than an attempt to catch every asynchronous task started by user code.

### 3. Preserve normal and queued pipeline timing

For normal interceptors, `DioMixin.fetch` invokes the private `_invoke*` method inside the existing pipeline Future, observes the runtime result, and returns the handler Future. It no longer forks a callback-specific error zone.

For queued interceptors, `QueuedInterceptor._handleQueue` receives the matching private `_invoke*` method. The callback is still called inside the existing synchronous try block, its returned Future is observed immediately, and queue advancement still occurs only through handler completion or cancellation. FIFO behavior and the separate request/response/error queues are unchanged.

The effective timing is therefore:

```text
Before:
  pipeline -> fork a new error zone -> invoke callback -> wait for handler
             callback Future errors are routed through the new zone

After:
  pipeline -> invoke callback in the existing zone
           -> attach an error listener synchronously
           -> wait for handler
             callback Future is observed but never awaited
```

This removes the cross-request error-zone boundary while preserving handler-driven ordering. Awaiting the callback Future was intentionally avoided because the await-based approach in #2139 changed microtask ordering and broke following interceptor behavior reported in #2167; #2169 reverted #2139.

## API and Compatibility Boundaries

- No public method return type changed.
- `InterceptorSendCallback`, `InterceptorSuccessCallback`, and `InterceptorErrorCallback` remain public `void Function(...)` typedefs.
- No public symbol was removed or renamed.
- No Dart SDK lower-bound increase or Dart 3-only language feature was introduced.
- Existing synchronous callbacks and handler methods retain their behavior.
- Async callbacks remain unawaited; only their returned errors are now observed explicitly.
- Subclasses of `Interceptor`, `InterceptorsWrapper`, `QueuedInterceptor`, and `QueuedInterceptorsWrapper` are all covered.
- Wrapper subclasses continue to dispatch through their public `on*` methods instead of being bypassed by an internal closure path.

The wrapper-subclass checks are compatibility guards, not a new user-visible behavior. A temporary separate CHANGELOG bullet for preserving those overrides was intentionally removed in `6b42fa2`; the final changelog contains only the #2565 user-facing fix.

## Tests

The regression and compatibility coverage includes:

- A deterministic #2565 test with an immediate failing adapter and a shared deduplication Future. It asserts both requests terminate, only one adapter fetch occurs, no error escapes the test's surrounding zone in the same error-zone setup, and no pending handler remains.
- Direct `Interceptor` subclasses with async `onRequest`, `onResponse`, and `onError` overrides that throw after an await.
- `InterceptorsWrapper` and `QueuedInterceptorsWrapper` subclasses overriding all three public `on*` methods synchronously, with async wrapper-subclass coverage on the request path.
- Async constructor callbacks covering request, response, and error paths.
- Queued request, response, and error failures, including queue release for a following request.
- Existing behavior for callbacks that call their handler and later fail, callbacks that must not be awaited, duplicate handler calls, and following error interceptors.

The focused VM and Chrome suites each pass all 52 tests.

## Verification

Locally, `melos run format`, `melos run analyze`, and `dart analyze --fatal-infos` passed. CI run 29642622254 passed on min, beta, and stable SDK matrices, including VM, Chrome, Firefox, Flutter, APK build, and coverage.

The final coverage report is:

| File | Base | New | Difference |
| --- | ---: | ---: | ---: |
| `dio/lib/src/dio_mixin.dart` | 94.25% | 94.49% | +0.24% |
| `dio/lib/src/interceptor.dart` | 99.34% | 99.46% | +0.12% |
| Overall | 85.73% | 85.95% | +0.22% |

A full local `melos run test` was attempted, but network-dependent VM and browser tests encountered external HTTP 502, CORS, and timeout failures. The CI workflow starts local httpbun containers on each GitHub Actions runner and completed successfully.

## Unverified

The minimum Dart 2.18 SDK was not installed locally; the remote min SDK workflow passed. Local Firefox and Flutter targets were unavailable or blocked by local configuration, but both passed in the remote CI matrices.

## AI Attribution

Implementation, tests, commit, and local review by Codex.

### DIFF
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -13,6 +13,8 @@ See the [Migration Guide][] for the complete breaking changes list.**
   `BackgroundTransformer`.
 - Fix concurrent requests hanging or reporting uncaught errors when an
   interceptor shares a failing Future, such as in request deduplication.
+- Preserve `onRequest`, `onResponse`, and `onError` overrides in subclasses of
+  `InterceptorsWrapper` and `QueuedInterceptorsWrapper`.
 
 ## 5.10.0
 
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -13,8 +13,6 @@ See the [Migration Guide][] for the complete breaking changes list.**
   `BackgroundTransformer`.
 - Fix concurrent requests hanging or reporting uncaught errors when an
   interceptor shares a failing Future, such as in request deduplication.
-- Preserve `onRequest`, `onResponse`, and `onError` overrides in subclasses of
-  `InterceptorsWrapper` and `QueuedInterceptorsWrapper`.
 
 ## 5.10.0
 
--- a/dio/CHANGELOG.md
+++ b/dio/CHANGELOG.md
@@ -11,6 +11,8 @@ See the [Migration Guide][] for the complete breaking changes list.**
   on an empty response body when a custom `responseDecoder` is set. It now
   returns an empty result, consistent with `SyncTransformer` and
   `BackgroundTransformer`.
+- Fix concurrent requests hanging or reporting uncaught errors when an
+  interceptor shares a failing Future, such as in request deduplication.
 
 ## 5.10.0
 
--- a/dio/lib/src/dio_mixin.dart
+++ b/dio/lib/src/dio_mixin.dart
@@ -426,34 +426,10 @@ abstract class DioMixin implements Dio {
       }
     }
 
-    // Create an error zone that catches uncaught async errors from
-    // interceptor callbacks. This prevents requests from hanging when
-    // an async interceptor throws without calling the handler.
-    // Synchronous exceptions are not caught here and propagate normally.
-    // If the handler was already completed (e.g. a fire-and-forget future
-    // inside the callback failed), forward the error to the parent zone
-    // so it isn't silently swallowed.
-    Zone createInterceptorZone(
-      _BaseHandler handler,
-      void Function(Object error, StackTrace stackTrace) onUncaughtError,
-    ) {
-      return Zone.current.fork(
-        specification: ZoneSpecification(
-          handleUncaughtError: (self, parent, zone, error, stackTrace) {
-            if (!handler.isCompleted) {
-              onUncaughtError(error, stackTrace);
-            } else {
-              parent.handleUncaughtError(zone, error, stackTrace);
-            }
-          },
-        ),
-      );
-    }
-
     // Convert the request interceptor to a functional callback in which
     // we can handle the return value of interceptor callback.
     FutureOr Function(dynamic) requestInterceptorWrapper(
-      InterceptorSendCallback cb,
+      _InterceptorCallback<RequestOptions, RequestInterceptorHandler> cb,
     ) {
       return (dynamic incomingState) {
         final state = incomingState as InterceptorState;
@@ -462,13 +438,15 @@ abstract class DioMixin implements Dio {
             requestOptions.cancelToken,
             Future(() async {
               final handler = RequestInterceptorHandler();
-              createInterceptorZone(
+              final result = cb(state.data as RequestOptions, handler);
+              _observeInterceptorCallback(
+                result,
                 handler,
                 (error, stackTrace) => handler.reject(
                   assureDioException(error, requestOptions, stackTrace),
                   true,
                 ),
-              ).run(() => cb(state.data as RequestOptions, handler));
+              );
               return handler.future;
             }),
           );
@@ -480,7 +458,7 @@ abstract class DioMixin implements Dio {
     // Convert the response interceptor to a functional callback in which
     // we can handle the return value of interceptor callback.
     FutureOr<dynamic> Function(dynamic) responseInterceptorWrapper(
-      InterceptorSuccessCallback cb,
+      _InterceptorCallback<Response<dynamic>, ResponseInterceptorHandler> cb,
     ) {
       return (dynamic incomingState) {
         final state = incomingState as InterceptorState;
@@ -490,13 +468,15 @@ abstract class DioMixin implements Dio {
             requestOptions.cancelToken,
             Future(() async {
               final handler = ResponseInterceptorHandler();
-              createInterceptorZone(
+              final result = cb(state.data as Response, handler);
+              _observeInterceptorCallback(
+                result,
                 handler,
                 (error, stackTrace) => handler.reject(
                   assureDioException(error, requestOptions, stackTrace),
                   true,
                 ),
-              ).run(() => cb(state.data as Response, handler));
+              );
               return handler.future;
             }),
           );
@@ -508,7 +488,7 @@ abstract class DioMixin implements Dio {
     // Convert the error interceptor to a functional callback in which
     // we can handle the return value of interceptor callback.
     FutureOr<dynamic> Function(Object) errorInterceptorWrapper(
-      InterceptorErrorCallback cb,
+      _InterceptorCallback<DioException, ErrorInterceptorHandler> cb,
     ) {
       return (dynamic error) {
         final state = error is InterceptorState
@@ -516,12 +496,14 @@ abstract class DioMixin implements Dio {
             : InterceptorState(assureDioException(error, requestOptions));
         Future<InterceptorState> handleError() async {
           final handler = ErrorInterceptorHandler();
-          createInterceptorZone(
+          final result = cb(state.data, handler);
+          _observeInterceptorCallback(
+            result,
             handler,
             (error, stackTrace) => handler.next(
               assureDioException(error, requestOptions, stackTrace),
             ),
-          ).run(() => cb(state.data, handler));
+          );
           return handler.future;
         }
 
@@ -552,7 +534,7 @@ abstract class DioMixin implements Dio {
     for (final interceptor in interceptors) {
       final fun = interceptor is QueuedInterceptor
           ? interceptor._handleRequest
-          : interceptor.onRequest;
+          : interceptor._invokeRequest;
       future = future.then(requestInterceptorWrapper(fun));
     }
 
@@ -569,6 +551,7 @@ abstract class DioMixin implements Dio {
         } on DioException catch (e) {
           handler.reject(e, true);
         }
+        return null;
       }),
     );
 
@@ -576,7 +559,7 @@ abstract class DioMixin implements Dio {
     for (final interceptor in interceptors) {
       final fun = interceptor is QueuedInterceptor
           ? interceptor._handleResponse
-          : interceptor.onResponse;
+          : interceptor._invokeResponse;
       future = future.then(responseInterceptorWrapper(fun));
     }
 
@@ -584,7 +567,7 @@ abstract class DioMixin implements Dio {
     for (final interceptor in interceptors) {
       final fun = interceptor is QueuedInterceptor
           ? interceptor._handleError
-          : interceptor.onError;
+          : interceptor._invokeError;
       future = future.catchError(errorInterceptorWrapper(fun));
     }
     // Normalize errors, converts errors to [DioException].
--- a/dio/lib/src/interceptor.dart
+++ b/dio/lib/src/interceptor.dart
@@ -318,16 +318,29 @@ mixin _InterceptorWrapperMixin on Interceptor {
   InterceptorSuccessCallback? _onResponse;
   InterceptorErrorCallback? _onError;
 
+  // The public callback typedefs remain void for compatibility. Invoke them
+  // dynamically so an async callback still exposes its runtime Future.
+
   @override
   void onRequest(
     RequestOptions options,
     RequestInterceptorHandler handler,
   ) {
-    if (_onRequest != null) {
-      _onRequest!(options, handler);
-    } else {
+    final callback = _onRequest;
+    if (callback == null) {
       handler.next(options);
+      return;
     }
+    final dynamic dynamicCallback = callback;
+    final result = dynamicCallback(options, handler);
+    _observeInterceptorCallback(
+      result,
+      handler,
+      (error, stackTrace) => handler.reject(
+        DioMixin.assureDioException(error, options, stackTrace),
+        true,
+      ),
+    );
   }
 
   @override
@@ -335,11 +348,25 @@ mixin _InterceptorWrapperMixin on Interceptor {
     Response<dynamic> response,
     ResponseInterceptorHandler handler,
   ) {
-    if (_onResponse != null) {
-      _onResponse!(response, handler);
-    } else {
+    final callback = _onResponse;
+    if (callback == null) {
       handler.next(response);
+      return;
     }
+    final dynamic dynamicCallback = callback;
+    final result = dynamicCallback(response, handler);
+    _observeInterceptorCallback(
+      result,
+      handler,
+      (error, stackTrace) => handler.reject(
+        DioMixin.assureDioException(
+          error,
+          response.requestOptions,
+          stackTrace,
+        ),
+        true,
+      ),
+    );
   }
 
   @override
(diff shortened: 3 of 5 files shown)


CaiJingLong:
## Review conclusion: Approve

### Change summary
Removes the per-callback forked error zones introduced by #2499 (which broke cross-request `Future` error delivery for shared `Completer`s in deduplication interceptors, #2565) and replaces them with a non-awaiting `_observeInterceptorCallback` listener. The listener preserves #2499's hang-prevention for async callbacks (returned `Future` errors reject/advance an incomplete handler) while restoring same-zone `Future` error delivery so shared-failure deduplication works again. Adds private `_invokeRequest/_invokeResponse/_invokeError` dynamic-dispatch entry points to capture the runtime `Future` of async `on*` overrides without changing the public `void` API, and rewires `QueuedInterceptor._handleQueue` and the wrapper mixin through the same observer.

### Review details
| Dimension | Result | Notes |
|------|------|------|
| Correctness | Pass | The observer is registered synchronously in the callback's existing zone, never awaits the `Future`, and keeps handler completion as the only pipeline signal. Incomplete-handler errors route to `reject`/`next`; complete-handler late errors route to `callbackZone.handleUncaughtError`, preserving #2499's late-error visibility. `QueuedInterceptor` returns `null` at the top-level `_handle*` so observation happens exactly once inside `_handleQueue`; FIFO and queue-release-on-cancel are unchanged. The dispatch callback's `return null` + observation also adds a safety net for non-`DioException` escapes from `_dispatchRequest`. |
| Tests | Pass | 52 interceptor/queued tests pass locally (`dart test test/interceptor_test.dart test/queued_interceptor_test.dart`). New coverage includes: the deterministic #2565 reproduction (asserts 2 requests, 1 adapter fetch, 2 joined errors, 0 uncaught, 0 pending), async `Interceptor` subclass errors on all three paths, `InterceptorsWrapper`/`QueuedInterceptorsWrapper` subclass override dispatch + async subclass errors, async constructor callbacks, queued request/response/error failures with queue release for the following request, the #2167 non-awaiting guard, and late-error-after-handler-completion reaching the callback zone. `dio_mixin_test`, `cancel_token_test`, `adapters_test`, `exception_test` also pass. The 7 failures in the full VM suite (`test_suite_test`, `stacktrace_test`, `options_timeout_integration_test`) are environmental network failures (DNS/Socket/real-timeout), not related to this change. |
| Style | Pass | Matches existing patterns; no debug residue; dynamic dispatch is localized and commented. Conventional Commits with scope, lowercase subjects, `Co-Authored-By: Codex` trailer on every commit. Branch `fix/2565-interceptor-error-zones` follows §8.1. |
| Risk | Pass | No public signature/typedef change, no SDK lower-bound increase, no Dart 3-only feature. Behavior delta vs 5.10.0 is narrow: a `Future` started inside a callback that is neither returned nor handled is no longer caught by a Dio-forked zone and will not unblock an incomplete handler — documented as an explicit ownership boundary. This edge case was incidental to #2499 (which targeted the async callback's own returned `Future`, now covered by the observer). Sensitive-area note: this touches the interceptor pipeline ordering/error propagation; the change reverts the #2499 zone boundary while preserving its hang fix, and the #2167 microtask-ordering regression is explicitly guarded by a test. |
| Docs | Pass | `CHANGELOG.md` updated under `## Unreleased` with one user-facing bullet. No public API doc changes needed (no API surface change). |
| Repo constraints | Pass | Motivation grounded in #2565; one PR/one concern; regression test included; AI attribution present in PR description and commit trailers; no drive-by dependency bumps; no public-API break. |

### Confirmed key points
- Correctness: cross-request shared-`Future` error delivery restored; #2499 async-throw hang prevention preserved via the observer; `QueuedInterceptor` queue ordering and cancel-release unchanged; wrapper subclass `on*` overrides dispatch correctly (private `_invoke*` overrides removed from the mixin).
- Tests: #2565 regression test asserts the exact issue invariants (single fetch, both requests terminate, no uncaught error, no pending handler); #2167 non-awaiting guard prevents reintroducing the #2139 microtask regression; all 52 interceptor/queued tests pass locally.
- Risk: no public API or SDK-constraint change; the only behavior delta (unowned detached `Future`s no longer caught) is a documented ownership boundary, not a regression of #2499's targeted case.

- The ownership-boundary rationale for unowned detached `Future`s is described in the PR body but not in a code comment near `_observeInterceptorCallback`; a one-line pointer comment there would help future maintainers understand why only returned `Future`s are observed.
- Consider noting in the PR description that the second CHANGELOG bullet from an earlier commit was intentionally removed (commit `6b42fa2`), so reviewers don't expect a "preserve wrapper overrides" changelog entry — the net behavior vs 5.10.0 for wrapper subclass overrides is unchanged.

AlexV525:
> The ownership-boundary rationale for unowned detached `Future`s is described in the PR body but not in a code comment near `_observeInterceptorCallback`; a one-line pointer comment there would help future maintainers understand why only returned `Future`s are observed.

Addressed in `64189f6`. I added a focused comment next to `_observeInterceptorCallback` explaining that only the returned Future can be associated with the handler, while detached asynchronous work remains owned by the callback's zone.

> Consider noting in the PR description that the second CHANGELOG bullet from an earlier commit was intentionally removed (commit `6b42fa2`), so reviewers don't expect a "preserve wrapper overrides" changelog entry — the net behavior vs 5.10.0 for wrapper subclass overrides is unchanged.

Addressed in the PR description under **API and Compatibility Boundaries**. It now states that the wrapper-subclass checks are compatibility guards rather than a new user-visible behavior, and that `6b42fa2` intentionally removed the temporary separate CHANGELOG entry.

Thank you for the detailed review.

CaiJingLong (approved):
## Review conclusion: Approve

### Change summary
Removes the per-callback forked error zones introduced by #2499 (which broke cross-request `Future` error delivery for shared `Completer`s in deduplication interceptors, #2565) and replaces them with a non-awaiting `_observeInterceptorCallback` listener. The listener preserves #2499's hang-prevention for async callbacks (returned `Future` errors reject/advance an incomplete handler) while restoring same-zone `Future` error delivery so shared-failure deduplication works again. Adds private `_invokeRequest/_invokeResponse/_invokeError` dynamic-dispatch entry points to capture the runtime `Future` of async `on*` overrides without changing the public `void` API, and rewires `QueuedInterceptor._handleQueue` and the wrapper mixin through the same observer.

### Review details
| Dimension | Result | Notes |
|------|------|------|
| Correctness | Pass | The observer is registered synchronously in the callback's existing zone, never awaits the `Future`, and keeps handler completion as the only pipeline signal. Incomplete-handler errors route to `reject`/`next`; complete-handler late errors route to `callbackZone.handleUncaughtError`, preserving #2499's late-error visibility. `QueuedInterceptor` returns `null` at the top-level `_handle*` so observation happens exactly once inside `_handleQueue`; FIFO and queue-release-on-cancel are unchanged. The dispatch callback's `return null` + observation also adds a safety net for non-`DioException` escapes from `_dispatchRequest`. |
| Tests | Pass | 52 interceptor/queued tests pass locally (`dart test test/interceptor_test.dart test/queued_interceptor_test.dart`). New coverage includes: the deterministic #2565 reproduction (asserts 2 requests, 1 adapter fetch, 2 joined errors, 0 uncaught, 0 pending), async `Interceptor` subclass errors on all three paths, `InterceptorsWrapper`/`QueuedInterceptorsWrapper` subclass override dispatch + async subclass errors, async constructor callbacks, queued request/response/error failures with queue release for the following request, the #2167 non-awaiting guard, and late-error-after-handler-completion reaching the callback zone. `dio_mixin_test`, `cancel_token_test`, `adapters_test`, `exception_test` also pass. The 7 failures in the full VM suite (`test_suite_test`, `stacktrace_test`, `options_timeout_integration_test`) are environmental network failures (DNS/Socket/real-timeout), not related to this change. |
| Style | Pass | Matches existing patterns; no debug residue; dynamic dispatch is localized and commented. Conventional Commits with scope, lowercase subjects, `Co-Authored-By: Codex` trailer on every commit. Branch `fix/2565-interceptor-error-zones` follows §8.1. |
| Risk | Pass | No public signature/typedef change, no SDK lower-bound increase, no Dart 3-only feature. Behavior delta vs 5.10.0 is narrow: a `Future` started inside a callback that is neither returned nor handled is no longer caught by a Dio-forked zone and will not unblock an incomplete handler — documented as an explicit ownership boundary. This edge case was incidental to #2499 (which targeted the async callback's own returned `Future`, now covered by the observer). Sensitive-area note: this touches the interceptor pipeline ordering/error propagation; the change reverts the #2499 zone boundary while preserving its hang fix, and the #2167 microtask-ordering regression is explicitly guarded by a test. |
| Docs | Pass | `CHANGELOG.md` updated under `## Unreleased` with one user-facing bullet. No public API doc changes needed (no API surface change). |
| Repo constraints | Pass | Motivation grounded in #2565; one PR/one concern; regression test included; AI attribution present in PR description and commit trailers; no drive-by dependency bumps; no public-API break. |

### Confirmed key points
- Correctness: cross-request shared-`Future` error delivery restored; #2499 async-throw hang prevention preserved via the observer; `QueuedInterceptor` queue ordering and cancel-release unchanged; wrapper subclass `on*` overrides dispatch correctly (private `_invoke*` overrides removed from the mixin).
- Tests: #2565 regression test asserts the exact issue invariants (single fetch, both requests terminate, no uncaught error, no pending handler); #2167 non-awaiting guard prevents reintroducing the #2139 microtask regression; all 52 interceptor/queued tests pass locally.
- Risk: no public API or SDK-constraint change; the only behavior delta (unowned detached `Future`s no longer caught) is a documented ownership boundary, not a regression of #2499's targeted case.

- The ownership-boundary rationale for unowned detached `Future`s is described in the PR body but not in a code comment near `_observeInterceptorCallback`; a one-line pointer comment there would help future maintainers understand why only returned `Future`s are observed.
- Consider noting in the PR description that the second CHANGELOG bullet from an earlier commit was intentionally removed (commit `6b42fa2`), so reviewers don't expect a "preserve wrapper overrides" changelog entry — the net behavior vs 5.10.0 for wrapper subclass overrides is unchanged.

### Verification performed
- `dart analyze --fatal-infos lib/src/interceptor.dart lib/src/dio_mixin.dart`: no issues.
- `dart test test/interceptor_test.dart test/queued_interceptor_test.dart`: all 52 tests passed.
- `dart test test/dio_mixin_test.dart test/cancel_token_test.dart test/adapters_test.dart test/exception_test.dart`: all passed (one skipped TLS test).
- Full `dart test` suite: 241 passed, 6 skipped, 7 failed — all 7 failures are network-dependent environmental tests (DNS resolution, real socket connections, live timeouts) unrelated to this change.

PR #2469, opened 2025-12-12
fix: Http2Adapter does not fail with StateError when connection is closed

vanelizarov:
Periodically I receive unhandled `StateError:  Bad state: Cannot add event after closing` in Crashlytics:

<details>
<summary>Stack Trace</summary>

```
StateError: Bad state: Cannot add event after closing
  *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** (***)
  pid: 19891, tid: 483148275864, name 1.ui (unparsed)
  android arch: arm64 comp: yes sim: no (os:)
  build_id: '4f1bdaeddbb49260128a5ff3c1e2e0c4' (unparsed)
  isolate_dso_base: 6ca549c000, vm_dso_base: 6ca549c000 (unparsed)
  isolate_instructions: 6ca5622940, vm_instructions: 6ca560c000 (unparsed)
#0      _StreamController.add (third_party/dart/sdk/lib/async/stream_controller.dart:617:24)
#1      _StreamSinkWrapper.add (third_party/dart/sdk/lib/async/stream_controller.dart:898:13)
#2      Http2Adapter._fetch.<anonymous closure> (/Users/vanelizarov/.pub-cache/git/dio-5259d4ae4fb665b8212c343c015e53fcbfe7d0b7/plugins/http2_adapter/lib/src/http2_adapter.dart:144:33)
#3      _RootZone.runUnaryGuarded (third_party/dart/sdk/lib/async/zone.dart:1778:10)
#4      _BufferingStreamSubscription._sendData (third_party/dart/sdk/lib/async/stream_impl.dart:381:11)
#5      _BufferingStreamSubscription._add (third_party/dart/sdk/lib/async/stream_impl.dart:312:7)
#6      _SinkTransformerStreamSubscription._add (third_party/dart/sdk/lib/async/stream_transformers.dart:67:11)
#7      _EventSinkWrapper.add (third_party/dart/sdk/lib/async/stream_transformers.dart:13:11)
#8      _transform.<anonymous closure> (/Users/vanelizarov/.pub-cache/hosted/pub.dev/dio-5.9.0/lib/src/progress_stream/io_progress_stream.dart:33:16)
#9      _HandlerEventSink.add (third_party/dart/sdk/lib/async/stream_transformers.dart:230:17)
#10     _SinkTransformerStreamSubscription._handleData (third_party/dart/sdk/lib/async/stream_transformers.dart:115:24)
[...]

PR #2571, opened 2026-07-20
fix(http2_adapter): handle closed streams during uploads

AlexV525:
Supersedes #2469.

The original report and reproduction are credited to @vanelizarov. This replacement is rebuilt on the current `main` branch and addresses the unresolved review feedback and the cancellation hang found while auditing the earlier implementation.

### Motivation

When an HTTP/2 peer closes a connection before a streamed request body finishes, `http2` closes its outgoing stream controller. The request body subscription can then deliver another chunk to the closed sink, producing an unhandled `StateError: Bad state: Cannot add event after closing` instead of the transport error that caused the closure.

### Changes

- Stop the request body subscription when the outgoing HTTP/2 sink has already closed.
- Explicitly settle request-body waiting when cancellation stops the subscription, rather than relying on `onDone`, which is not called after cancellation.
- Keep cancellation callbacks weakly referenced so shared, never-cancelled tokens do not retain completed request resources.
- Assert the exact `TransportConnectionException` for a peer connection closure and cover cancellation while the source stream remains open.
- Update the `dio_http2_adapter` changelog.

### Verification

With only the new tests applied to `main`, the connection-closure test fails with the reported `StateError`, and the cancellation test times out after two seconds. Both pass with this change, and the cancellation test also confirms that the source stream listener is removed.

`dart test` for `plugins/http2_adapter`, `melos run format`, and `melos run analyze` complete successfully. An independent adversarial pass also exercised 20 concurrent cancellations, send timeout, a synchronous stream controller, and request-source error propagation.

### Hosted verification

The min, stable, and beta workflows all pass with the configured httpbun and proxy services. The stable workflow also passes formatting, analysis, publish dry-run, VM/Chrome/Firefox/Flutter tests, the example APK build, and coverage reporting. Changed-file coverage for `http2_adapter.dart` increases from 75.69% to 82.42%.

### AI assistance

Implementation, tests, and local review were performed with Codex. A separate Codex sub-agent performed the adversarial review. The original diagnosis and reproduction came from @vanelizarov in #2469.

- [ ] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none *(not applicable - this supersedes #2469)*
- [ ] I have updated the documentation (if necessary) *(not applicable - no public API or documented behavior changes)*

The replacement preserves the final outgoing-sink close as a normal `await`; the observed `StateError` originates from adding the next body chunk after the transport has closed the sink, not from closing the sink again.

### DIFF
--- a/plugins/http2_adapter/CHANGELOG.md
+++ b/plugins/http2_adapter/CHANGELOG.md
@@ -5,6 +5,8 @@ See the [Migration Guide][] for the complete breaking changes list.**
 
 ## Unreleased
 
+- Prevent uploads from throwing a `StateError` when the HTTP/2 connection closes
+  before the request body finishes streaming.
 - Run HTTP integration and certificate pinning tests against the configured
   test server.
 
--- a/plugins/http2_adapter/lib/src/http2_adapter.dart
+++ b/plugins/http2_adapter/lib/src/http2_adapter.dart
@@ -127,12 +127,6 @@ class Http2Adapter implements HttpClientAdapter {
     final streamWR = WeakReference<ClientTransportStream>(stream);
 
     final hasRequestData = requestStream != null;
-    if (hasRequestData && cancelFuture != null) {
-      cancelFuture.whenComplete(() {
-        streamWR.target?.outgoingMessages.close();
-      });
-    }
-
     List<Uint8List>? list;
     if (!excludeMethods.contains(options.method) && hasRequestData) {
       list = await requestStream.toList();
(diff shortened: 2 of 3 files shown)


CaiJingLong (approved):
## Review conclusion: Approve

### Summary of changes

Stops the request-body subscription when the HTTP/2 outgoing sink has already closed (peer connection closure during upload), and explicitly settles the adapter future on cancellation instead of relying on `onDone` (which is not called after `cancel()`). Cancellation callbacks now hold only `WeakReference`s to the subscription/completer/stream so a shared, never-completed cancel token cannot retain completed request resources.

### Review details

|-----------|------------|-------|
| Correctness | Pass | The `StateError` from `stream.outgoingMessages.add(...)` on a closed sink is now caught and the subscription is stopped via `stopRequestStream()`, letting the response listener's `onError` surface the real `TransportConnectionException`. Cancellation now completes `requestCompleter` explicitly, fixing the hang caused by `onDone` not firing after `cancel()`. Idempotency holds: `stopRequestStream`, the `cancelFuture.whenComplete`, and the sendTimeout path all guard on `requestCompleter.isCompleted` and use `.ignore()` on repeated close/cancel. |
| Tests | Pass | Two new tests under `group('request stream')`. Verified locally: with only the tests applied to `main`, the closure test fails with `StateError: Bad state: Cannot add event after closing` and the cancellation test times out at 2s; both pass with this change. The cancellation test also asserts `requestController.hasListener` is `false`, confirming the subscription is removed. |
| Style | Pass | Matches existing patterns (`WeakReference` already used for `streamWR`, `.ignore()` idiom consistent with the rest of the file). No debug residue, no commented-out code. `dart analyze` clean, `dart format` clean. |
| Risk | Pass | Internal implementation change only; no public API or behavior change for the non-error path. Sensitive area (transport/cancellation) is touched, but the change narrows failure modes rather than broadening them. SDK lower bound is `>=3.0.0`; `WeakReference` has been available since Dart 2.17. Regression risk is low: the non-cancellation, non-closure path is behaviorally identical (listen → add → onDone completes → close). |
| Documentation | Pass | `plugins/http2_adapter/CHANGELOG.md` updated under `## Unreleased` with a user-facing bullet. No public API or documented behavior changes, so no README/doc-comment updates needed (correctly marked N/A in the PR checklist). |
| Repo constraints | Pass | Branch `fix/2469-http2-closed-stream` follows `category/ticket-id-description`. Commit uses Conventional Commits (`fix(http2_adapter): ...`, lowercase imperative subject) with `Co-Authored-By` trailers for both human and AI contributors. PR description discloses AI assistance and states verification honestly in prose. Supersedes #2469 per the description. |

### Confirmed key points

- Correctness: the reported `StateError` is eliminated and the underlying transport error propagates; the cancellation hang is fixed by explicitly completing `requestCompleter` in the `cancelFuture.whenComplete` callback.
- Tests: regression tests reproduce both bugs on `main` and pass with the fix; `dart test test/http2_test.dart` passes (the only failures in the broader suite are environmental — proxy/httpbun/SSL-pinning services not available locally).
- Risk: no public API change, no SDK bump, no dependency change; sensitive transport path is made more robust.

- The three pre-existing environmental test failures (`request with payload via proxy`, `test_suite_test.dart`, `SSL pinning untrusted certificate tested and allowed #2`) are unrelated to this PR and require external services; no action needed here.

> This comment was generated by AI agent omp (model: glm-5-2).

PR #2498, opened 2026-03-11
fix(cookie_manager): deduplicate cookies when request is retried

ersanKolay:
## Summary

Fixes #2442.

When a request is retried (e.g. after token refresh), `loadCookies` merges `previousCookies` from the existing header with `savedCookies` from the cookie jar without checking for duplicates. This causes cookies to accumulate on each retry attempt:

```
1st attempt: a=1; b=2
2nd attempt: a=1; b=2; a=1; b=2
3rd attempt: a=1; b=2; a=1; b=2; a=1; b=2
```

The fix filters out `previousCookies` entries whose names already exist in `savedCookies`, so jar cookies take precedence and no duplicates are produced.

The deduplication is done in `loadCookies` rather than `getCookies` to preserve the existing RFC 6265 behavior where same-name cookies with different paths are valid.

## Test plan

- Added `no duplicate cookies on retry` test that simulates the retry scenario
- All existing tests pass (16/16)
- `dart format` and `dart analyze` clean

PR #2572, opened 2026-07-20
fix(cookie_manager): prevent duplicate cookies on reused request options

### LINKED ISSUE #2442: [Question][CookieManager] About override the old cookies by new cookies when they have the same keys?
### Package

cookie_manager

### Version

3.3.0

### Operating-System

iOS

### Adapter

Default Dio

```shell
[!] Flutter (Channel stable, 3.32.7, on macOS 15.3.1 24D70 darwin-arm64, locale en-VN) [2.9s]
    • Flutter version 3.32.7 on channel stable at /Users/myname/Desktop/Development/SDKs/flutter
    ! Upstream repository unknown source is not a standard remote.
      Set environment variable "FLUTTER_GIT_URL" to unknown source to dismiss this error.
    • Framework revision d7b523b356 (9 days ago), 2025-07-15 17:03:46 -0700
    • Engine revision 39d6d6e699
    • Dart version 3.8.1
    • DevTools version 2.45.1
    • If those were intentional, you can disregard the above warnings; however it is recommended to use "git"
      directly to perform update checks and upgrades.

[✓] Android toolchain - develop for Android devices (Android SDK version 36.0.0) [7.8s]
    • Android SDK at /Users/myname/Library/Android/sdk
    • Platform android-36, build-tools 36.0.0
    • ANDROID_HOME = /Users/myname/Library/Android/sdk
    • ANDROID_SDK_ROOT = /Users/myname/Library/Android/sdk
    • Java binary at: /Users/myname/.CICD/JDK/jdk-17.jdk/Contents/Home/bin/java
      This JDK is specified in your Flutter configuration.
      To change the current JDK, run: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 17.0.7+0-17.0.7b1000.6-10550314)

[✓] Xcode - develop for iOS and macOS (Xcode 16.2) [7.9s]
    • Build 16C5032a
    • CocoaPods version 1.16.2

[✓] Chrome - develop for the web [31ms]

[✓] Android Studio (version 2025.1.2) [31ms]
    • Android Studio at /Applications/Android Studio Preview.app/Contents
    • Java version OpenJDK Runtime Environment (build 21.0.6+-13391695-b895.109)

[✓] Android Studio (version 2024.2) [30ms]
    • Android Studio at /Applications/Android Studio 3.app/Contents
    • Java version OpenJDK Runtime Environment (build 21.0.5+-12932927-b750.29)

[✓] IntelliJ IDEA Community Edition (version 2025.1.2) [29ms]
    • IntelliJ at /Applications/IntelliJ IDEA CE.app

[✓] VS Code (version 1.102.2) [11ms]
    • Flutter extension version 3.114.0

[✓] Connected device (3 available) [12.6s]
    • iPhone 15 (mobile) • 9142A079-C987-4702-9215-9E21E12C28DC • ios            •
      com.apple.CoreSimulator.SimRuntime.iOS-17-4 (simulator)
    • macOS (desktop)    • macos                                • darwin-arm64   • macOS 15.3.1 24D70
      darwin-arm64
    • Chrome (web)       • chrome                               • web-javascript • Google Chrome
      138.0.7204.169

[✓] Network resources [502ms]

! Doctor found issues in 1 category.
```

### Dart Version

3.8.1

### Steps to Reproduce

In [loadCookies](https://github.com/cfug/dio/blob/main/plugins/cookie_manager/lib/src/cookie_mgr.dart#L133) function, why don’t we override the old cookie with the new cookie value when the keys are the same?
```dart
Future<String> loadCookies(RequestOptions options) async {
  final savedCookies = await cookieJar.loadForRequest(options.uri);
  final previousCookies =
      options.headers[HttpHeaders.cookieHeader] as String?; // <== (1)
  final cookies = getCookies([
    ...?previousCookies
        ?.split(';')
        .where((e) => e.isNotEmpty)
        .map((c) => Cookie.fromSetCookieValue(c)),
    ...savedCookies,
  ]);
  return cookies;
}
```
When does case (1) happen? In most case, I debugged, its values are null, but my in production, rarely case it has values.

Am I correct in thinking that if savedCookies contain the same keys as previousCookies, the old values will not be overridden, and the returned cookies will contain duplicate keys?
For example:
```dart
previousCookies = 'a=1; b=2';
savedCookies   = 'a=3; b=4';

// Result:
cookies = 'a=1; a=3; b=2; b=4';
```

### Expected Result

N/A

### Actual Result

N/A

AlexV525:
Closes #2442.
Supersedes #2498.

## Summary

Reusing the same `RequestOptions` instance sends it through `CookieManager` again with the Cookie header produced by the previous pass. Merging that generated header with the cookie jar again makes jar cookies accumulate on every pass.

This change keeps weak per-`RequestOptions` state containing the original source header and the last header generated by the manager. When the current header still exactly matches that generated value, `loadCookies` rebuilds it from the original caller input and the latest jar state.

Unlike the name-only filtering proposed in #2498, this preserves the established behavior where caller-provided and jar cookies with the same name can coexist. It also removes stale jar values after updates, deletion, or an origin change.

The state is intentionally scoped to the same `CookieManager` and `RequestOptions` instance reported in #2442. A copied request or a different manager instance is a new identity and is outside this fix.

### Why an `Expando`

[`Expando<T>`](https://api.dart.dev/dart-core/Expando-class.html) associates private data with an object identity without adding a field to that object's class. The association does not keep its property value alive after the key object becomes inaccessible. It has been part of `dart:core` since before Dart 1.0, so this use does not raise the package's Dart 2.18 lower bound.

`CookieManager` owns one instance:

```dart
final Expando<_CookieHeaderState> _cookieHeaderStates = Expando();
```

Each key is a `RequestOptions` object, and each value contains two immutable strings: the header supplied as the source of the merge and the header last generated by this manager. Keeping this bookkeeping outside `RequestOptions.extra` avoids reserving a user-visible key, exposing internal state to loggers, or copying an internal marker through `RequestOptions.copyWith`.

### State transitions

| Interceptor pass | Current Cookie header | Source used for the merge | Result |
| --- | --- | --- | --- |
| First pass | Caller header or `null` | Current header | Source + cookies currently loaded from the jar |
| Reused options, unchanged header | Exactly equals this manager's last output | Stored original source | Original source + freshly loaded jar cookies |
| Header replaced between passes | Does not equal the last output | New current header | New source + freshly loaded jar cookies |

For example:

```text
first pass:  source=provided=value, jar=session=old
             -> provided=value; session=old

second pass after the jar changes to session=new:
             source=provided=value, jar=session=new
             -> provided=value; session=new
```

The old jar contribution is not treated as caller input, so it is neither duplicated nor retained after an update, deletion, or origin change.

### Interceptor ordering

Request interceptors run in registration order. `CookieManager` reads both `options.uri` and the current Cookie header at its position in that chain, so interceptors that mutate either value must run before it:

```dart
dio.interceptors
  ..add(uriOrCookieMutatingInterceptor)
  ..add(CookieManager(cookieJar))
  ..add(interceptorThatDoesNotChangeUriOrCookies);
```

Changes made by a later interceptor have these consequences:

| Later mutation | Effect when the same options are reused |
| --- | --- |
| Authorization, body, method, `extra`, or unrelated headers | Does not affect Cookie state matching |
| Replace the complete Cookie header | The replacement becomes the source for the next merge |
| Append, reorder, or reformat the generated Cookie header | It no longer exactly matches the stored output; the embedded old jar cookies can be treated as source cookies and merged again |
| Change `baseUrl` or path | The current request may carry cookies selected for the URI observed before that change |

The URI case is an ordering requirement for the current request, not only for retries. Moving a request from one origin or path to another after `CookieManager` has selected cookies can send cookies outside the scope for which they were loaded.

### Why not deduplicate cookie names or values

A serialized Cookie request header contains `name=value` pairs but not the domain and path metadata that identified the cookies in the jar. The same request can legitimately contain multiple cookies with the same name, and dio also deliberately preserves a caller-provided cookie alongside a same-name jar cookie.

Filtering by name, or by a reconstructed cookie identity, therefore cannot distinguish a cookie generated by the previous interceptor pass from one explicitly supplied by the caller. Tracking the exact previous output preserves that provenance without changing first-pass merge behavior.

### API and identity boundaries

- No public API, request header precedence, or cookie ordering is changed.
- No state is added to `RequestOptions.extra`.
- Entries are scoped to one `CookieManager` and one `RequestOptions` object identity.
- A `copyWith` result or a different manager instance is a new identity and intentionally starts with its current header as a new source.

## Verification

Added six regression tests covering three sequential `Dio.fetch` calls with the same options, caller-provided same-name cookies, jar update and deletion, origin changes, caller header replacement, and state isolation between options. All six fail against `origin/main` and pass with this change.

The `dio_cookie_manager` package passes 21 tests. Scoped Melos format and fatal-info analysis checks are clean. Local coverage records 74 of 75 executable lines in `cookie_mgr.dart` (98.67%), including all new state logic.

Implementation and tests by Codex; independent authenticity and adversarial review passes by two Codex subagents.

- [ ] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none *(not applicable: this replaces #2498)*
- [ ] I have updated the documentation *(not applicable: no public API or usage changes)*
- [x] I have run the affected package tests without failures

### Additional context and info

The previous implementation in #2498 used RFC 6265 cookie-store identity rules to justify filtering request-header cookies. Cookie-store replacement and request-header merging are different operations; this PR avoids changing first-pass merge semantics.

### DIFF
--- a/plugins/cookie_manager/CHANGELOG.md
+++ b/plugins/cookie_manager/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-*None.*
+- Prevent duplicate cookies when reusing request options while preserving
+  caller-provided cookies.
 
 ## 3.4.0
 
--- a/plugins/cookie_manager/lib/src/cookie_mgr.dart
+++ b/plugins/cookie_manager/lib/src/cookie_mgr.dart
@@ -29,6 +29,11 @@ class _CookieHeaderState {
 }
 
 /// Cookie manager for HTTP requests based on [CookieJar].
+///
+/// Register this after interceptors that may change [RequestOptions.uri] or
+/// the Cookie request header. Cookies are selected from the URI and reused
+/// request options are recognized only while the header exactly matches this
+/// manager's previous output.
 class CookieManager extends Interceptor {
   CookieManager(
     this.cookieJar, {
@@ -154,6 +159,12 @@ class CookieManager extends Interceptor {
   }
 
   /// Load cookies in cookie string for the request.
+  ///
+  /// State is scoped to the identity of [options]. When the same instance is
+  /// reused, its original Cookie header is restored only if the current header
+  /// exactly matches the previous result of this method. Any other header is
+  /// treated as a new source. Incrementally modifying a generated header after
+  /// this manager can therefore cause saved cookies to be merged again.
   Future<String> loadCookies(RequestOptions options) async {
     final savedCookies = await cookieJar.loadForRequest(options.uri);
     final previousCookies =
--- a/plugins/cookie_manager/lib/src/cookie_mgr.dart
+++ b/plugins/cookie_manager/lib/src/cookie_mgr.dart
@@ -21,6 +21,13 @@ const _kIsWeb = _kIsWebInterop || _kIsWebUtil || identical(0, 0.0);
 /// attribute like "expires=Sun, 19 Feb 3000 01:43:15 GMT", which could also contain commas.
 final _setCookieReg = RegExp('(?<=)(,)(?=[^;]+?=)');
 
+class _CookieHeaderState {
+  const _CookieHeaderState(this.source, this.merged);
+
+  final String? source;
+  final String merged;
+}
+
 /// Cookie manager for HTTP requests based on [CookieJar].
 class CookieManager extends Interceptor {
   CookieManager(
@@ -37,6 +44,8 @@ class CookieManager extends Interceptor {
 
   /// Whether to ignore invalid cookies during parsing or saving.
   bool ignoreInvalidCookies;
+
+  final Expando<_CookieHeaderState> _cookieHeaderStates = Expando();
 
   /// Merge cookies into a Cookie string.
   /// Cookies with longer paths are listed before cookies with shorter paths.
@@ -149,14 +158,22 @@ class CookieManager extends Interceptor {
     final savedCookies = await cookieJar.loadForRequest(options.uri);
     final previousCookies =
         options.headers[HttpHeaders.cookieHeader] as String?;
+    final previousState = _cookieHeaderStates[options];
+    // Rebuild a header that still exactly matches this manager's last output.
+    // Any other header becomes the source for the next merge.
+    final sourceCookies =
+        previousState != null && previousState.merged == previousCookies
+            ? previousState.source
+            : previousCookies;
     final cookies = getCookies([
-      ...?previousCookies
+      ...?sourceCookies
           ?.split(';')
           .where((e) => e.isNotEmpty)
           .map((c) => _fromSetCookieValue(c))
           .whereType<Cookie>(), // Use .nonNulls when the minimum SDK is 3.0.
       ...savedCookies,
     ]);
+    _cookieHeaderStates[options] = _CookieHeaderState(sourceCookies, cookies);
     return cookies;
   }
 
--- a/plugins/cookie_manager/test/cookies_test.dart
+++ b/plugins/cookie_manager/test/cookies_test.dart
@@ -1,4 +1,5 @@
 @TestOn('vm')
+import 'dart:async';
 import 'dart:io';
 import 'dart:typed_data';
 
@@ -8,20 +9,38 @@ import 'package:dio/io.dart';
 import 'package:dio_cookie_manager/dio_cookie_manager.dart';
 import 'package:test/test.dart';
 
-class MockRequestInterceptorHandler extends RequestInterceptorHandler {
-  MockRequestInterceptorHandler(this.expectResult);
+class _TestRequestInterceptorHandler extends RequestInterceptorHandler {
+  final Completer<RequestOptions> _result = Completer();
 
-  final String expectResult;
+  Future<RequestOptions> get result => _result.future;
 
   @override
   void next(RequestOptions requestOptions) {
-    final c = requestOptions.headers[HttpHeaders.cookieHeader];
-    expect(c == expectResult, true);
-    super.next(requestOptions);
+    _result.complete(requestOptions);
+  }
+
+  @override
+  void reject(
+    DioException error, [
+    bool callFollowingErrorInterceptor = false,
+  ]) {
+    _result.completeError(error, error.stackTrace);
   }
 }
 
 class MockResponseInterceptorHandler extends ResponseInterceptorHandler {}
+
+Future<void> expectRequestCookies(
+  CookieManager cookieManager,
+  RequestOptions options,
+  String? expected,
+) async {
+  final handler = _TestRequestInterceptorHandler();
+  await cookieManager.onRequest(options, handler);
+  final result = await handler.result;
+  expect(result, same(options));
+  expect(options.headers[HttpHeaders.cookieHeader], expected);
+}
 
 class _MockRejectRequestInterceptorHandler extends RequestInterceptorHandler {
   _MockRejectRequestInterceptorHandler(this.matcher);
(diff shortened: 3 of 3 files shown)


CaiJingLong (approved):
## Review Conclusion: Approve

### Summary of Changes

This PR fixes duplicate cookies accumulating when the same `RequestOptions` instance is reused across repeated interceptor passes (the retry scenario from #2442). `CookieManager` now keeps weak per-`RequestOptions` state (via `Expando`) recording the original caller-supplied Cookie header (`source`) and the last header generated by the manager (`merged`). On a reused pass whose current header exactly matches the stored `merged` value, `loadCookies` rebuilds from the stored original `source` plus freshly loaded jar cookies, instead of treating the previous output as new caller input.

### Review Details

|-----------|------------|-------|
| Correctness | Pass | State transition logic is sound. Traced first-pass, reused-pass, jar-update, caller-header-replacement, and origin-change paths; all produce correct non-duplicating output. The `merged == previousCookies` guard correctly distinguishes the manager's own previous output from a caller-modified header. `Expando` does not retain the key after GC, so no memory leak. Empty-jar / empty-header edge cases are handled (empty `merged` vs `null` header falls through to the new-source branch, which is correct). |
| Tests | Pass | 6 new regression tests under `reusing request options` cover: repeated `Dio.fetch` with the same options, same-name caller + jar cookies, jar update + `deleteAll`, caller header replacement, origin change, and per-options state isolation. Verified locally: all 6 fail against `origin/main` source (e.g. `provided=first; saved=value; saved=value`) and pass with this PR. The end-to-end test via `_CookieRecordingAdapter` asserts the header observed at the adapter level, not just after `onRequest`, which is the right place to catch duplication. Full suite: 21/21 pass. |
| Style | Pass | Follows existing file conventions. `_CookieHeaderState` is a private immutable value class with a `const` constructor. `Expando` is a standard `dart:core` API available well below the package's 2.18 lower bound — no SDK bump. No debug residue, no commented-out code. |
| Risk | Pass | No public API, signature, default behavior, or header precedence change. State is scoped to `CookieManager` + `RequestOptions` identity; `copyWith` and different manager instances correctly start fresh (verified by the isolation test). The documented limitation — a later interceptor that appends/reformats the generated header will cause jar cookies to be re-merged on reuse — is pre-existing on `main` (same duplicate occurs) and is honestly disclosed in both the PR body and the `loadCookies`/class doc comments. Concurrent reuse of one `RequestOptions` across in-flight requests is outside the fix scope and is not a regression. |
| Docs | Pass | `CHANGELOG.md` updated under `## Unreleased`. Class and `loadCookies` doc comments document the new ordering requirement (register after URI/cookie-mutating interceptors) and the exact-match reuse semantics. No public API or usage changes, so README/API docs need no update. |
| Repo constraints | Pass | Branch `fix/2442-cookie-retry-dedup` follows `category/ticket-id-description`. Commits use Conventional Commits (`fix(...)`, `docs(...)`) with `Co-Authored-By: Codex` attribution. PR description discloses AI usage. PR template items are checked honestly; unchecked items carry `(not applicable — reason)` notes. Motivation traces to #2442; supersedes #2498 with a narrower, semantics-preserving approach. |

### Confirmed Key Points

- **Correctness:** The `source`/`merged` state machine correctly reconstructs the original caller input on reuse and reloads jar cookies fresh, eliminating the accumulation reported in #2442 while preserving the existing same-name caller + jar coexistence behavior. No off-by-one, null-deref, or unhandled branch found.
- **Tests:** 6 regression tests reproduce the issue against `main` and pass with the fix; full `dio_cookie_manager` suite (21 tests) green; `dart analyze` clean; `dart format` clean.
- **Risk:** Additive, identity-scoped, no API/behavioral break. The only behavioral subtlety (later-interceptor header mutation defeating the exact-match guard) is pre-existing and explicitly documented.

- The PR checklist item "I have read the Documentation" is unchecked without a note. Consider checking it or adding a brief `(not applicable — no public API change)` note for consistency with the other unchecked items, though this is purely cosmetic.

> 本评论由 AI agent omp（模型: glm-5-2）生成

PR #2479, opened 2026-01-08
fix(native_dio_adapter): gracefully fall back when Cronet unavailable

DevMohammadSalameh:
## Summary

Fixes #2444

When Google Play Services is unavailable (e.g., on Android emulators without Google APIs or devices without Play Services), `CronetEngine` initialization throws a `JniException`:

> "All available Cronet providers are disabled. A provider should be enabled before it can be used."

This change:
- Wraps `CronetAdapter` creation in a try-catch block
- Falls back to the standard `HttpClientAdapter` (dart:io) when Cronet is unavailable
- Adds an optional `onCronetUnavailable` callback for developers who want to be notified when the fallback occurs

## Changes

- `native_adapter.dart`: Added graceful fallback with try-catch
- `CHANGELOG.md`: Documented the new behavior

## Usage

```dart
// Basic usage - fallback happens silently
final dio = Dio();
dio.httpClientAdapter = NativeAdapter();

// With callback to log or handle the fallback
dio.httpClientAdapter = NativeAdapter(
  onCronetUnavailable: (error) {
    print('Cronet unavailable, using dart:io fallback: $error');
  },
);
```

## Test Plan

- [x] `melos run format` - passes
- [x] `melos run analyze` - passes  
- [x] `flutter test` in native_dio_adapter - all 10 tests pass
- [ ] Manual test on Android emulator without Google Play Services (requires device testing)

PR #2445, opened 2025-09-17
[native_dio_adapter] Fall back on `IOClient` when `CronetEngine` initialization fails

LukasMirbt:
Draft for how to solve https://github.com/cfug/dio/issues/2444.

### New Pull Request Checklist

- [x] I have read the [Documentation](https://pub.dev/documentation/dio/latest/)
- [x] I have searched for a similar pull request in the [project](https://github.com/cfug/dio/pulls) and found none
- [x] I have updated this branch with the latest `main` branch to avoid conflicts (via merge from master or rebase)
- [ ] I have added the required tests to prove the fix/feature I'm adding
- [ ] I have updated the documentation (if necessary)
- [ ] I have run the tests without failures
- [ ] I have updated the `CHANGELOG.md` in the corresponding package

PR #2573, opened 2026-07-21
feat(native_dio_adapter): opt-in Cronet provider fallback

### LINKED ISSUE #2444: [native_dio_adapter] All network requests fail when Google Play services are unavailable
### Package

native_dio_adapter

### Version

1.5.0

### Operating-System

Android

### Adapter

NativeAdapter

### Steps to Reproduce

1. Create an Android emulator that does not have access to Google Play Services (select Services -> Android Open Source)
2. Make a network request on a `Dio` instance with a `NativeAdapter` 
3. Note that the request fails and an exception is thrown when initializing the `CronetEngine`.

### Expected Result

`NativeAdapter` should gracefully handle devices without Google Play services (for example by falling back on `IOClient`).

Cronet can be embedded as a workaround but in my opinion it's quite a big footgun that all network requests fail on certain devices if the developer is not aware of how `cronet_http` interacts with Google Play services.

I created a [draft PR](https://github.com/cfug/dio/pull/2445) with a possible approach for handling the issue.

### Actual Result

`NativeAdapter` throws an exception and network requests fail.

```
JniException (Exception in Java code called through JNI: java.lang.RuntimeException: All available Cronet providers are disabled. A provider should be enabled before it can be used.
      
      java.lang.RuntimeException: All available Cronet providers are disabled. A provider should be enabled before it can be used.
      	at org.chromium.net.CronetEngine$Builder.getEnabledCronetProviders(CronetEngine.java:365)
      	at org.chromium.net.CronetEngine$Builder.createBuilderDelegate(CronetEngine.java:327)
      	at org.chromium.net.CronetEngine$Builder.<init>(CronetEngine.java:75)
      	at android.os.MessageQueue.nativePollOnce(Native Method)
      	at android.os.MessageQueue.next(MessageQueue.java:346)
      	at android.os.Looper.loopOnce(Looper.java:189)
      	at android.os.Looper.loop(Looper.java:317)
      	at android.app.ActivityThread.main(ActivityThread.java:8705)
      	at java.lang.reflect.Method.invoke(Native Method)
      	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
      	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:886)
      )
```

AlexV525:
## Summary

Fixes #2444.

Adds an opt-in `createFallbackAdapter` factory to `NativeAdapter`. When Cronet reports that every installed provider on the device is disabled (for example, AOSP emulators or devices without Google Play services), the caller-supplied factory returns any `HttpClientAdapter` and requests continue via that adapter — instead of every request failing on `CronetEngine.build()`.

```dart
NativeAdapter(
  createFallbackAdapter: (error, stackTrace) => IOHttpClientAdapter(),
)
```

### Key properties

- **Opt-in.** The parameter defaults to `null`. Without it, `NativeAdapter` continues to use Cronet and propagates initialization errors exactly as it does today. No behavior change for existing applications.
- **Precise error classification.** Matches the exact Chromium provider-disabled `RuntimeException` on a typed `JniException` (adds `jni` as a direct dependency). Does **not** fall back on any other `JniException`, non-`JniException` errors carrying similar text, or engine initialization errors of other kinds.
- **Post-init errors are untouched.** Connection, TLS, timeout, redirect, cancellation, and response-stream errors from a successfully-initialized Cronet path remain Cronet errors and are propagated unchanged.
- **Sticky selection.** The choice is made lazily on the first `fetch` and cached; later requests do not probe Cronet again. `close()` closes the selected adapter once. Closing before the first request does not initialize Cronet or create the fallback.
- **Adapter, not `Client`.** The factory returns any `HttpClientAdapter` — callers pick the one whose TLS, proxy, cookie, transport, and observability behavior matches their needs.

### Why this shape

This PR replaces the two prior attempts against #2444, which were rejected on their approach:

- **#2445** wrapped `CronetEngine.build()` in `CronetAdapter` and switched the internal `http.Client` mid-request. That is not safe once `ConversionLayerAdapter` has consumed the Dio request stream.
- **#2479** made the fallback silent and default, fixed the fallback to `HttpClientAdapter()` (dart:io), and matched on `.toString()`. That silently changes networking behavior for existing apps and is fragile against exception-message changes.

The rationale for the current design is captured in `docs/plans/cronet-provider-fallback.md`.

## Changes

- `plugins/native_dio_adapter/lib/src/cronet_fallback_adapter.dart` — new Android-only lazy selection wrapper `CronetWithFallbackAdapter`, the classifier `isCronetProviderUnavailable`, and the `CreateFallbackAdapter` typedef. Includes a `@visibleForTesting` constructor that lets unit tests inject a controllable "build cronet path" seam without native code.
- `plugins/native_dio_adapter/lib/src/native_adapter.dart` — new `createFallbackAdapter` parameter; wired on Android only, only when the factory is supplied.
- `plugins/native_dio_adapter/lib/native_dio_adapter.dart` — export the `CreateFallbackAdapter` typedef.
- `plugins/native_dio_adapter/pubspec.yaml` — add direct `jni: ^0.14.0` dependency (compatible with `cronet_http: ^1.5.0`'s transitive `^0.14.2`).
- `plugins/native_dio_adapter/README.md` — opt-in example section with the observable-behavior caveat.
- `plugins/native_dio_adapter/CHANGELOG.md` — Unreleased entry.
- `plugins/native_dio_adapter/test/cronet_fallback_adapter_test.dart` — 12 focused tests using a controllable "build cronet adapter" seam and mock `HttpClientAdapter` implementations.

## Test plan

- [x] `melos run format` — passes.
- [x] `melos run analyze` — passes.
- [x] `melos run test` — passes (26 tests in `native_dio_adapter`, including 12 new).
- [x] Adversarial second-opinion review against the plan — no blocking issues.
- [ ] Android integration on an AOSP emulator without Google Play services. This complements unit coverage; not blocking on it here. If the CI matrix cannot provide the required emulator, this should be recorded separately per the plan.

### Unit coverage (per plan)

- Classified error triggers the fallback factory exactly once and forwards the unchanged request, body, and cancel future by identity.
- Non-matching `JniException` (wrong exception class or wrong message) is rethrown and no fallback is created.
- Non-`JniException` error carrying the same text is rejected by the classifier.
- `ArgumentError` during Cronet build is rethrown, not fallen back.
- Errors thrown by the successfully-built Cronet adapter (post-init connection/TLS/timeout errors) do NOT trigger a fallback.
- Selected fallback and selected Cronet adapter are each reused across subsequent requests.
- `close` after selection closes the selected adapter exactly once (and later `close` calls are no-ops).
- `close` before any request does not invoke the build seam or create a fallback.

The "no fallback factory" case is satisfied by construction — `NativeAdapter` does not instantiate the wrapper when the factory is `null`, and calling with a `null` factory is indistinguishable from prior versions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/plugins/native_dio_adapter/CHANGELOG.md
+++ b/plugins/native_dio_adapter/CHANGELOG.md
@@ -2,7 +2,15 @@
 
 ## Unreleased
 
-*None.*
+- Add opt-in `createFallbackAdapter` to `NativeAdapter`. On Android, when the
+  device has installed Cronet providers but every provider is disabled (for
+  example AOSP emulators or devices without Google Play services), the
+  supplied factory returns any `HttpClientAdapter` and requests continue via
+  that adapter. Detection is strictly limited to Cronet's
+  provider-disabled `RuntimeException`; every other Cronet error is
+  propagated unchanged. Adapter selection is sticky for the lifetime of the
+  `NativeAdapter`. Fixes
+  [#2444](https://github.com/cfug/dio/issues/2444).
 
 ## 1.6.0
 
--- a/plugins/native_dio_adapter/README.md
+++ b/plugins/native_dio_adapter/README.md
@@ -41,6 +41,40 @@ final dioClient = Dio();
 dioClient.httpClientAdapter = NativeAdapter();
 ```
 
+### Opt-in Cronet provider fallback (Android)
+
+On Android, `NativeAdapter` uses Cronet. Some devices — for example, AOSP
+emulators or devices without Google Play services — install Cronet providers
+but leave every provider disabled. On those devices `CronetEngine.build()`
+throws and every request fails
+([issue #2444](https://github.com/cfug/dio/issues/2444)).
+
+If your application needs to support that environment, pass an opt-in
+`createFallbackAdapter`. The factory is invoked **only** when the provider is
+known to be disabled and lets you choose any `HttpClientAdapter`:
+
+```dart
+import 'package:dio/io.dart';
+import 'package:native_dio_adapter/native_dio_adapter.dart';
+
+dioClient.httpClientAdapter = NativeAdapter(
+  createFallbackAdapter: (error, stackTrace) => IOHttpClientAdapter(),
+);
+```
+
+Notes:
+
+- Detection is limited to Cronet's provider-disabled `RuntimeException`.
+  Connection, TLS, timeout, redirect, cancellation, and response-stream
+  errors remain Cronet errors and are propagated unchanged.
+- The selection is sticky for the lifetime of the `NativeAdapter`. Once
+  Cronet is picked, later requests do not probe again; once the fallback is
+  picked, later requests reuse it.
+- Changing adapters can change observable networking behavior: TLS
+  configuration, proxy handling, cookie storage, supported protocols
+  (HTTP/2, HTTP/3), and connection pooling. Callers opting in own that
+  tradeoff — pick the adapter that best matches your requirements.
+
 ### Use embedded Cronet
 
 Starting from `cronet_http` v1.2.0,
--- a/plugins/native_dio_adapter/lib/native_dio_adapter.dart
+++ b/plugins/native_dio_adapter/lib/native_dio_adapter.dart
@@ -5,5 +5,6 @@ export 'package:cupertino_http/cupertino_http.dart';
 
 export 'src/conversion_layer_adapter.dart';
 export 'src/cronet_adapter.dart';
+export 'src/cronet_fallback_adapter.dart' show CreateFallbackAdapter;
 export 'src/cupertino_adapter.dart';
 export 'src/native_adapter.dart';
--- a/plugins/native_dio_adapter/lib/src/cronet_fallback_adapter.dart
+++ b/plugins/native_dio_adapter/lib/src/cronet_fallback_adapter.dart
@@ -1,0 +1,131 @@
+import 'dart:typed_data' show Uint8List;
+
+import 'package:cronet_http/cronet_http.dart';
+import 'package:dio/dio.dart';
+import 'package:flutter/foundation.dart' show visibleForTesting;
+import 'package:jni/jni.dart' show JniException;
+
+import 'cronet_adapter.dart';
+
+/// Signature for building the fallback [HttpClientAdapter] used when the
+/// default Cronet provider is unavailable on the current device.
+///
+/// See [NativeAdapter.new] and the package README for the intended contract.
+typedef CreateFallbackAdapter = HttpClientAdapter Function(
+  Object error,
+  StackTrace stackTrace,
+);
+
+/// Exact `RuntimeException` message thrown by Chromium's Cronet API when
+/// every registered `CronetProvider` on the device is disabled.
+///
+/// Chromium's provider-selection branch throws this exact `RuntimeException`:
+/// https://chromium.googlesource.com/chromium/src/+/lkgr/components/cronet/android/api/src/org/chromium/net/CronetEngine.java
+///
+/// The referenced `CronetEngine.Builder.getPreferredCronetProvider` branch
+/// throws this message when providers exist but all are disabled. It is
+/// distinct from the separate "Unable to find any Cronet provider" error,
+/// which reports that no provider was discovered at all.
+const cronetProvidersDisabledMessage =
+    'java.lang.RuntimeException: All available Cronet providers are disabled. '
+    'A provider should be enabled before it can be used.';
+
+/// Classifies the failure that indicates all installed Cronet providers on
+/// the device are disabled.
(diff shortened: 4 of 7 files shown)


CaiJingLong (approved):
## Review Conclusion: Approve

### Change Summary

Adds an opt-in `createFallbackAdapter` factory parameter to `NativeAdapter`. On Android, when `CronetEngine.build()` throws the specific provider-disabled `RuntimeException` (all Cronet providers disabled, e.g. AOSP emulators / devices without Google Play services), the caller-supplied factory returns any `HttpClientAdapter` and requests continue via that adapter. The parameter defaults to `null`; without it, behavior is unchanged. Fixes #2444.

### Review Details

|-----------|-----------|-------|
| Correctness | Pass | Lazy selection in `_selectAdapter()` is synchronous with no await points, so no race in a single Dart isolate. The try/catch wraps only `_buildCronetAdapter()`, so post-init errors (connection/TLS/timeout/redirect/cancellation) from a successfully-built Cronet adapter propagate unchanged — verified by test. The classifier `isCronetProviderUnavailable` requires `error is JniException` **and** `message.contains(cronetProvidersDisabledMessage)` with the exact `java.lang.RuntimeException:` prefix, so a wrong exception class or a non-`JniException` carrying the same text is rejected. Sticky selection via `_selected` is correct. `close()` is idempotent (`_closed` flag) and does not build Cronet when no request was made. |
| Tests | Pass | 12 new focused tests in `cronet_fallback_adapter_test.dart` using a `@visibleForTesting` build seam and recording mock adapters. Covers: classifier positive/negative (wrong class, wrong message, non-JniException), fallback invoked exactly once with identity-preserved request/body/cancel, non-matching rethrow, `ArgumentError` rethrow, post-init errors do not trigger fallback, sticky reuse of both fallback and Cronet paths, `close` after selection (exactly once, `force` forwarded), `close` before any request (no build, no fallback). Full package suite: 26 tests pass. `dart analyze` clean. |
| Style | Pass | Matches existing module boundaries and naming. `CreateFallbackAdapter` typedef exported via `show` only — `CronetWithFallbackAdapter`, the classifier, and internals are not part of the public API. No debug remnants. `dart format` clean. |
| Risk | Pass | Additive, opt-in, `null` default — no behavior change for existing apps, no public-API break. Direct `jni: ^0.14.0` dependency is justified (needed for the typed `JniException` match) and compatible with `cronet_http: ^1.5.0`'s transitive `jni: ^0.14.2` (resolves to 0.14.2). The eager `CronetEngine.build()` in the fallback path vs. the lazy build in the non-fallback path is intentional — it is the mechanism that surfaces the provider-disabled error before any request is delegated. Sensitive area (network adapter), but the change is well-isolated behind the opt-in parameter. |
| Docs | Pass | `README.md` adds an opt-in section with the observable-behavior caveat. `CHANGELOG.md` Unreleased entry is user-facing and references #2444. API doc on the new parameter is thorough (contract, stickiness, close semantics, platform scope, example). |
| Repo constraints | Pass | Branch `feat/cronet-provider-fallback` follows `category/short-description`. Commit uses Conventional Commits (`feat(native_dio_adapter): ...`), lowercase subject, no gitmoji+prefix mix. `Co-Authored-By` AI attribution present. CHANGELOG updated. No public-API break. No drive-by dependency bumps — the `jni` addition is the point and is documented. |

### Confirmed Key Points

- **Correctness**: The classifier is precise (typed `JniException` + exact `RuntimeException` message via `contains` to tolerate the appended Java stack trace). Post-init errors are untouched because the try/catch scope is only the build seam. Selection is sticky; `close()` is idempotent and never initializes Cronet when no request was made.
- **Tests**: 12 new tests + 14 existing = 26 pass; `dart analyze` and `dart format` clean. Coverage on the new file is 78.26%.
- **Risk**: Opt-in with `null` default — zero behavior change for non-opting apps. The `jni` direct dependency is compatible with the existing transitive resolution. No public symbols removed/renamed; only an additive optional parameter and a new exported typedef.

- The PR body references `docs/plans/cronet-provider-fallback.md`, but that file is not part of the diff. Consider either including it or dropping the reference to avoid a dangling citation.
- Android integration on an AOSP emulator without Google Play services remains unverified (acknowledged in the PR body). The unit tests use a controllable build seam and mock the `JniException`; real-device validation would close the loop on the exact runtime `JniException.message` shape.
- Pre-existing (not introduced by this PR): the package declares `sdk: '>=3.1.0'` while `cronet_http: ^1.5.0` transitively pulls `jni: ^0.14.2` which requires `sdk: '>=3.3.0'`. This constraint mismatch predates this PR and is not worsened by the added direct `jni: ^0.14.0` dependency.

This comment was generated by AI agent omp (model: devin/glm-5-2).

PR #2595, opened 2026-08-20
✨ Deprecate misspelled `handshakeTimout` in favor of `handshakeTimeout`

AlexV525:
## Motivation

The `ConnectionManager` factory introduced in #2435 (dio_http2_adapter 2.x) exposed the handshake timeout under a misspelled parameter name — `handshakeTimout` — while everything downstream of it already used the correct spelling: the private `_ConnectionManager` constructor parameter, the `handshakeTimeout` CHANGELOG entry, and the field it initializes. The typo was noticed during review of #2583.

Per the compatibility policy this is fixed via a deprecation cycle rather than a silent rename.

## Changes

- Add the correctly spelled `handshakeTimeout` parameter to the `ConnectionManager` factory.
- Deprecate `handshakeTimout` with `@Deprecated('Use handshakeTimeout instead. This will be removed in 3.0.0')` (current version is 2.8.0, so the removal targets the next major).
- Both parameters are nullable with no default at the factory; resolution is `handshakeTimeout ?? handshakeTimout`, and `_ConnectionManager` keeps applying its own 15 s default when neither is passed. Passing both makes the new spelling win.
- Existing callers that pass `handshakeTimout` compile and behave identically until the alias is removed.

## Verification

- The existing handshake-timeout test now uses the new spelling, and a companion test exercises the deprecated alias (asserting the same `TimeoutException` path), so the forwarding stays guarded until removal — both pass.
- Full package suite green: `dart test` in `plugins/http2_adapter` → 105 passed, 4 skipped; `dart analyze` and `dart format` clean.

*Implementation, tests, and CHANGELOG by GLM-5.2 via omp.*

### DIFF
--- a/plugins/http2_adapter/CHANGELOG.md
+++ b/plugins/http2_adapter/CHANGELOG.md
@@ -14,6 +14,9 @@ See the [Migration Guide][] for the complete breaking changes list.**
   successful TLS handshake (server selected `http/1.1` instead of `h2`),
   the `SecureSocket` is now explicitly destroyed before routing to
   `fallbackAdapter`, rather than being abandoned.
+- Deprecate the misspelled `ConnectionManager.handshakeTimout` parameter in
+  favor of the correctly spelled `handshakeTimeout`; the deprecated alias
+  keeps working and is scheduled for removal in 3.0.0.
 
 ## 2.8.0
 
--- a/plugins/http2_adapter/lib/src/connection_manager.dart
+++ b/plugins/http2_adapter/lib/src/connection_manager.dart
@@ -15,7 +15,11 @@ abstract class ConnectionManager {
   /// request through [Http2Adapter.fallbackAdapter].
   factory ConnectionManager({
     Duration idleTimeout = const Duration(seconds: 15),
-    Duration handshakeTimout = const Duration(seconds: 15),
+    @Deprecated(
+      'Use handshakeTimeout instead. This will be removed in 3.0.0',
+    )
+    Duration? handshakeTimout,
+    Duration? handshakeTimeout,
     void Function(Uri uri, ClientSetting)? onClientCreate,
     ProxyConnectedPredicate proxyConnectedPredicate =
         defaultProxyConnectedPredicate,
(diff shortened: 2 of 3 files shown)


CaiJingLong (approved):
Thanks for the clean deprecation cycle — this is the right way to remediate the
`handshakeTimout` typo introduced in #2435. I verified the key claims against
the source on this branch:

- `@Deprecated('Use handshakeTimeout instead. This will be removed in 3.0.0')`
  satisfies AGENTS.md §3: replacement + removal version are stated, and 3.0.0
  is the next major after the current 2.8.0.
- No behavioral change for the default path. Previously
  `handshakeTimout` defaulted to `15s` at the factory; now both params are
  nullable, `handshakeTimeout ?? handshakeTimout` resolves to `null` when
  neither is passed, and `_ConnectionManager` still applies its own
  `?? const Duration(seconds: 15)`. The effective default is unchanged.
- Existing explicit callers of `handshakeTimout` compile and behave
  identically until removal.
- Tests cover both spellings and assert the same `TimeoutException` path;
  the deprecated-alias test stays green because the root
  `analysis_options.yaml` sets `deprecated_member_use_from_same_package: ignore`.
- CHANGELOG entry is under `## Unreleased` as one concise bullet; README and
  the package migration guide don't reference the misspelled name, so no doc
  updates are needed there.
- CI is green (analyze / workflows beta / min / stable) and coverage is flat
  at 86.05%.

Minor, non-blocking suggestions:

1. **Test the "both passed" case.** The PR body states "passing both makes the
   new spelling win", but there is no test asserting that
   `ConnectionManager(handshakeTimeout: X, handshakeTimout: Y)` forwards `X`.
   A small test would guard that resolution until the alias is removed.

2. **Self-document the deprecated call site.** The new alias test relies on the
   global `deprecated_member_use_from_same_package: ignore`. Adding an explicit
   `// ignore: deprecated_member_use_from_same_package` on that
   `ConnectionManager(handshakeTimout: ...)` call would make the intent
   self-documenting and protect the test if the global ignore is ever narrowed.

3. **Commit type.** `✨` (feat) is defensible since a new parameter is added,
   though `🐛` (fix) also fits since this remediates a typo. Not worth
   re-spinning the commit either way.

Nice work — ship it after considering the above.

PR #2591, opened 2026-08-14
fix(dio): support classes that implement Interceptor

AlexV525:
Closes #2590

## Problem

`NoSuchMethodError: Class 'X' has no instance method '_invokeError'` at runtime when a class uses `implements Interceptor` instead of `extends Interceptor`.

PR #2567 added private instance methods (`_invokeRequest`, `_invokeResponse`, `_invokeError`) on `Interceptor` and dispatched the interceptor pipeline through them. These methods exist only via inheritance — a class that `implements Interceptor` provides the public `onRequest`/`onResponse`/`onError` methods but does not inherit the private ones, so the pipeline crashes at runtime.

This is a behavioral regression introduced in 5.11.0 (the release that shipped #2567).

## Fix

Move the dynamic-dispatch mechanism from private instance methods on `Interceptor` to a top-level helper function `_invokeCallbackDynamically`, and call it via closures at each dispatch site (`DioMixin.fetch` and `QueuedInterceptor._handle*`). The pipeline now references exclusively the **public** `onRequest`/`onResponse`/`onError` methods. Both `extends Interceptor` and `implements Interceptor` work.

The dynamic invocation is preserved: a `Future<void>` returned by an `async` override of a `void` method is still captured at runtime so `_observeInterceptorCallback` can observe asynchronous errors — the original motivation from #2567.

### Why a top-level function instead of private instance methods?

Private instance methods are not part of the interface contract. Any class using `implements Interceptor` would not inherit them. A top-level function that takes the public method as a parameter keeps dispatch through the public API surface only. The `Interceptor` class doc comment now documents this dispatch contract.

## Compatibility

- No public API change: no signature change, no symbol added/removed, no typedef change.
- No SDK lower-bound increase.
- No behavior change for `extends Interceptor`, `InterceptorsWrapper`, `QueuedInterceptor`, or `QueuedInterceptorsWrapper`.
- The fix restores 5.10 behavior for `implements Interceptor` users.

## Verification

Added 3 regression tests covering `implements Interceptor`:
- Synchronous callbacks through request → response stages.
- Error stage reached on a failing response (404).
- Async `onRequest` override that throws — dynamic dispatch captures the `Future`, observer rejects the handler.

[...]

PR #2596, opened 2026-08-20
👷 Require a changelog entry when a package's `lib` changes

### LINKED ISSUE #1662: Add CI step to ensure a changelog entry is added
### Request Statement

We should add a CI job, which ensures that a PR always adds a changelog entry.
We can avoid things like [this](https://github.com/cfug/dio/pull/1659) follow-up by doing that.

dylanpulver:
Closes #1662

The only guard today is the manual checkbox from #1679, which that PR called a temporary solution. It misses things: of the last 200 commits on `main`, 14 changed a package's `lib` without adding a line to that package's `CHANGELOG.md`, including #2464 and #2534.

This adds a `Check changelog entry` workflow and `scripts/check_changelog_entry.sh`. The script diffs the PR against its base, takes every changed path under `<package>/lib/`, and requires `<package>/CHANGELOG.md` to have gained at least one line; a changelog edit that only removes lines does not count. The package root comes from the path rather than a hardcoded list, so a new package is covered the day it is created, and a package with no `CHANGELOG.md` is exempt, which keeps `dio_test` and the example apps out of the way. Only paths under `lib/` trigger it, so tests, docs, workflows and `pubspec.yaml` never do.

The escape hatch is a `skip-changelog` label rather than a magic string in the commit message, because only someone with write access can apply one, so the waiver stays a maintainer decision. The workflow listens to `labeled`/`unlabeled`, so applying it re-runs the check without a new push. The label does not exist in this repo yet; its name lives only in `SKIP_CHANGELOG_LABEL`.

Trigger is `pull_request`, not `pull_request_target`, so a fork PR runs in the fork's context with a read-only token: no secret is read, the job requests `contents: read` and nothing else, and nothing is posted back. No `paths` filter, on purpose, so it always reports a result and can be made a required check without blocking docs-only PRs.

I ran the script over the last 200 commits of `main`, treating each as a PR, against an independently written Python implementation of the same rule. The two agreed on all 200: 186 pass, 14 fail. #2591 (changed `dio/lib`, added a changelog line) passes; #2464 (changed `dio/lib/src/options.dart`, added only tests) fails; #2534, which changed both `dio/lib` and `plugins/web_adapter/lib` but updated only the web adapter changelog, fails naming `dio` alone. I reproduced a GitHub merge ref locally with `git merge --no-ff` and confirmed it fails on a missing entry, passes once added, passes with the label in the payload, and still fails when the changelog is touched but only loses a line. A label payload with shell metacharacters is parsed as JSON and stays inert. `shellcheck` clean; `zizmor` 1.29.0 with the repo's config and `--persona=pedantic` reports nothing for the new workflow or for `.github` as a whole.

Unverified: the workflow has never executed on GitHub Actions. Only a real run exercises the event payload, the merge commit `actions/checkout` leaves at `HEAD`, the sparse checkout of `scripts`, and the re-run on a label change; I emulated each locally. The first real evidence is this PR, which changes no package `lib` and should pass by finding nothing to require. I did not run `melos run test` or `melos run analyze`, since no Dart source is touched.

No `CHANGELOG.md` here, since AGENTS.md scopes that to packages the change touches and this touches none, matching workflow-only PRs such as #2587.

Claude (Opus) was used to design and implement this change and to run the local verification described above. I reviewed the result and own it.

### DIFF
--- a/.github/workflows/check_changelog_entry.yml
+++ b/.github/workflows/check_changelog_entry.yml
@@ -1,13 +1,19 @@
 # Require a changelog entry whenever a pull request changes a package's `lib/`.
 name: Check changelog entry
 
-# The workflow deliberately runs for every pull request instead of filtering on
-# paths, so that the check always reports a result and can be made required.
+# Only pull requests that touch a package's `lib/` can need an entry, so the
+# rest never start a job. `paths` is matched against the whole pull request
+# (base...head), not just the newest push, so a pull request stays in scope
+# while its changelog entry is added and the check re-runs to pass.
 # `labeled` and `unlabeled` are included so that adding or removing the
 # `skip-changelog` label re-evaluates the pull request.
+# Note: a filtered workflow reports no result on the pull requests it skips, so
+# this check cannot be made a required status check as written.
 on:
   pull_request:
     types: [ opened, reopened, synchronize, labeled, unlabeled ]
+    paths:
+      - '**/lib/**'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
--- a/.github/workflows/check_changelog_entry.yml
+++ b/.github/workflows/check_changelog_entry.yml
@@ -1,0 +1,39 @@
+# Require a changelog entry whenever a pull request changes a package's `lib/`.
+name: Check changelog entry
+
+# The workflow deliberately runs for every pull request instead of filtering on
+# paths, so that the check always reports a result and can be made required.
+# `labeled` and `unlabeled` are included so that adding or removing the
+# `skip-changelog` label re-evaluates the pull request.
+on:
+  pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/check_changelog_entry.yml
```diff
@@ -0,0 +1,39 @@
+# Require a changelog entry whenever a pull request changes a package's `lib/`.
+name: Check changelog entry
+
+# The workflow deliberately runs for every pull request instead of filtering on
+# paths, so that the check always reports a result and can be made required.
+# `labeled` and `unlabeled` are included so that adding or removing the
+# `skip-changelog` label re-evaluates the pull request.
+on:
+  pull_request:
+    types: [ opened, reopened, synchronize, labeled, unlabeled ]
```
AlexV525, line 14: Would a path glob filter be more effective here?

PR #2601, opened 2026-09-04
🐛 Ignore 1xx interim responses in http2_adapter

### LINKED ISSUE #2600: http2_adapter: 1xx informational responses (e.g. 103 Early Hints) crash the request
### Package

http2_adapter

### Version

2.9.0

### Operating-System

Linux

### Adapter

Http2Adapter

```shell
[✓] Flutter (Channel stable, 3.47.1, on Arch Linux 7.1.9-zen1-2-zen, locale en_US.UTF-8)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.0.0)
    ! Multiple adb binaries found. This can cause conflicts and device detection issues:
        - /home/grish/Android/Sdk/platform-tools/adb
        - /usr/bin/adb
[✗] Chrome - develop for the web (Cannot find Chrome executable at google-chrome)
    ! Cannot find Chrome. Try setting CHROME_EXECUTABLE to a Chrome executable.
[✓] Linux toolchain - develop for Linux desktop
[✓] Connected device (1 available)
[✓] Network resources
```

### Dart Version

Dart SDK version: 3.13.1 (stable) (Tue Aug 18 01:00:59 2026 -0700) on "linux_x64"

### Steps to Reproduce

Self-contained, no external server needed (h2c works over plain `http://`):

```yaml
dependencies:
  dio: ^5.2.0
  dio_http2_adapter: ^2.9.0
  http2: ^2.3.0
```

```dart
import 'dart:async';
import 'dart:io';
import 'package:dio_http2_adapter/dio_http2_adapter.dart';
import 'package:http2/http2.dart';

Future<void> main() async {
  final server = await _EarlyHintsServer.bind();
  final dio = Dio();
  dio.httpClientAdapter = Http2Adapter(ConnectionManager());
  try {
    final res = await dio.get<String>(
      'http://127.0.0.1:${server.port}/',
      options: Options(responseType: ResponseType.plain),
    );
    print('STATUS: ${res.statusCode}');
  } on DioException catch (e) {
    print('DioException type=${e.type} status=${e.response?.statusCode}');
  }
  await Future<void>.delayed(const Duration(seconds: 1));
  await server.close();
}

class _EarlyHintsServer {
  _EarlyHintsServer._(this._socket, this._connections);
  final ServerSocket _socket;
  final List<ServerTransportConnection> _connections;
  int get port => _socket.port;

  static Future<_EarlyHintsServer> bind() async {
    final connections = <ServerTransportConnection>[];
    final socket = await ServerSocket.bind('127.0.0.1', 0);
    socket.listen((client) {
      final connection = ServerTransportConnection.viaSocket(client);
      connections.add(connection);
      connection.incomingStreams.listen((stream) {
        stream.sendHeaders([
          Header.ascii(':status', '103'),
          Header.ascii('link', '</s.css>; rel=preload; as=style'),
        ], endStream: false);
        stream.sendHeaders([
          Header.ascii(':status', '200'),
          Header.ascii('content-type', 'text/plain'),
        ], endStream: false);
        stream.sendData('hello'.codeUnits, endStream: true);
      });
    });
    return _EarlyHintsServer._(socket, connections);
  }

  Future<void> close() async {
    for (final c in _connections) {
      await c.finish();
    }
    await _socket.close();
  }
}
```

Crashes every run.

### Expected Result

dio should ignore the `103` and resolve with the final `200` response and its body (`hello`). RFC 9110 §15.2 requires clients to tolerate one or more interim 1xx responses before the final response on the same stream the `103` here isn't the response to act on.

### Actual Result

dio treats the `103` as the final response. It resolves with a `DioException` of `type: badResponse` and `statusCode: 103` (103 fails the default `validateStatus` check). When the real final HEADERS frame (`:status: 200`) then arrives on the same stream, the adapter tries to complete the same `Completer` a second time and throws an unhandled exception that crashes the isolate:

```
Unhandled exception:
Bad state: Future already completed
#0      _AsyncCompleter.complete (dart:async/future_impl.dart:97:31)
#1      Http2Adapter._fetch.<anonymous closure> (package:dio_http2_adapter/src/http2_adapter.dart:236:31)
```

CDNs don't send Early Hints on every response, so this fails intermittently the same endpoint succeeds most of the time and crashes whenever a `103` happens to precede the `200`.

Cause, `lib/src/http2_adapter.dart` lines 228–236 (same in 2.8.0, 2.9.0, and main):

```dart
final status = responseHeaders.value(':status');
if (status != null) {
  statusCode = int.parse(status);
  responseHeaders.removeAll(':status');
  needRedirect = _needRedirect(options, statusCode);
  needResponse =
      !needRedirect && options.validateStatus(statusCode) ||
          options.receiveDataWhenStatusError;
  responseCompleter.complete();   // fires for the 103, then again for the 200
}
```

`validateStatus` and the `ConnectionManager`/`ClientSetting` options don't touch this path the double-complete is unconditional. A secondary effect of the same code: nothing resets `responseHeaders` between HEADERS frames, only `:status` gets removed, so headers from the `103` (like `link`) also leak into whatever ends up in the final response.

I'd propose skipping 1xx HEADERS frames (other than 101, which HTTP/2 has no use for) instead of completing on them, and clearing `responseHeaders` when a new HEADERS frame starts so interim headers don't carry over.

chiliec:

Closes #2600.

**Problem.** When an HTTP/2 server sends a `103 Early Hints` (or any 1xx)
HEADERS frame before the final response, `Http2Adapter._fetch` treated the
interim frame as final: it parsed the `103` `:status`, completed the response
`Completer`, and then completed the *same* `Completer` again when the real
`200` HEADERS frame arrived on the same stream, crashing the isolate with
`Bad state: Future already completed`. As a secondary effect, `responseHeaders`
was never reset between HEADERS frames, so interim headers (e.g. `link`) leaked
into the final response.

**Fix.** In the HEADERS-frame handler, 1xx responses are now skipped (the
handler returns without completing), and the accumulated response headers are
reset per final HEADERS frame so interim headers don't carry over. Single-frame
responses (the common case) are unaffected.

**Note (sensitive area — header handling, §3).** This touches header handling,
so I kept the change minimal and additive: no public API change, no signature
change, default behavior for normal single-HEADERS-frame responses is
identical.

**Verification.** Added a regression test (`plugins/http2_adapter/test/early_hints_test.dart`)
that drives a local h2c server emitting `103` then `200` on the same stream and
asserts the request resolves with `200`, body `hello`, and that the interim
`link` header does not leak. Confirmed genuine RED→GREEN: reverting only the
source change makes the test fail with the exact `Bad state: Future already
completed` crash; with the fix it passes. `redirect_test.dart` (14 tests, same
HEADERS-frame path) and `headers_test.dart` still pass; `dart analyze` and
`dart format` are clean on both changed files.

*AI disclosure (§8.3): implementation and tests were produced with Claude; a human owns and has reviewed the change.*

### DIFF
--- a/plugins/http2_adapter/CHANGELOG.md
+++ b/plugins/http2_adapter/CHANGELOG.md
@@ -5,7 +5,11 @@ See the [Migration Guide][] for the complete breaking changes list.**
 
 ## Unreleased
 
-*None.*
+- Fix crash on 1xx informational responses (e.g. `103 Early Hints`): interim
+  responses are now ignored so the request resolves with the final response
+  instead of throwing `Bad state: Future already completed`. Response headers
+  are also reset per HEADERS frame so interim headers no longer leak into the
+  final response.
 
 ## 2.9.0
 
--- a/plugins/http2_adapter/CHANGELOG.md
+++ b/plugins/http2_adapter/CHANGELOG.md
@@ -9,7 +9,12 @@ See the [Migration Guide][] for the complete breaking changes list.**
   responses are now ignored so the request resolves with the final response
   instead of throwing `Bad state: Future already completed`. Response headers
   are also reset per HEADERS frame so interim headers no longer leak into the
-  final response.
+  final response. Trailer HEADERS frames (a HEADERS frame with no `:status`
+  after the response body, e.g. gRPC trailing metadata) are consequently
+  discarded rather than merged into the response headers; proper trailer
+  support is tracked in https://github.com/cfug/dio/issues/2602. A malformed
+  1xx response carrying END_STREAM now fails fast with a `connectionError`
+  instead of hanging until `receiveTimeout`.
 
 ## 2.9.0
 
--- a/plugins/http2_adapter/lib/src/http2_adapter.dart
+++ b/plugins/http2_adapter/lib/src/http2_adapter.dart
@@ -234,6 +234,21 @@ class Http2Adapter implements HttpClientAdapter {
           if (status != null) {
             final code = int.parse(status);
             if (code >= 100 && code < 200) {
+              // Interim (1xx) responses precede a final response and never
+              // terminate the stream (RFC 9110 §15.2, RFC 9113 §8.4). A 1xx
+              // HEADERS frame carrying END_STREAM is therefore malformed: the
+              // stream ends without a final response. Fail fast instead of
+              // waiting forever on a pooled connection (matching the behavior
+              // of Go's `x/net/http2`).
+              if (message.endStream && !responseCompleter.isCompleted) {
+                responseCompleter.completeError(
+                  DioException.connectionError(
+                    requestOptions: options,
+                    reason: 'Received an interim 1xx response with END_STREAM; '
+                        'the stream ended without a final response.',
+                  ),
+                );
+              }
               return;
             }
             statusCode = code;
--- a/plugins/http2_adapter/lib/src/http2_adapter.dart
+++ b/plugins/http2_adapter/lib/src/http2_adapter.dart
@@ -219,16 +219,28 @@ class Http2Adapter implements HttpClientAdapter {
     responseSubscription = stream.incomingMessages.listen(
       (StreamMessage message) async {
         if (message is HeadersStreamMessage) {
+          final frameHeaders = <MapEntry<String, String>>[];
+          String? status;
           for (final header in message.headers) {
             final name = utf8.decode(header.name);
             final value = utf8.decode(header.value);
-            responseHeaders.add(name, value);
+            if (name == ':status') {
+              status = value;
+            } else {
+              frameHeaders.add(MapEntry(name, value));
+            }
           }
 
-          final status = responseHeaders.value(':status');
           if (status != null) {
-            statusCode = int.parse(status);
-            responseHeaders.removeAll(':status');
+            final code = int.parse(status);
+            if (code >= 100 && code < 200) {
+              return;
+            }
+            statusCode = code;
+            responseHeaders.clear();
+            for (final entry in frameHeaders) {
+              responseHeaders.add(entry.key, entry.value);
+            }
             needRedirect = _needRedirect(options, statusCode);
             needResponse =
                 !needRedirect && options.validateStatus(statusCode) ||
--- a/plugins/http2_adapter/test/early_hints_test.dart
+++ b/plugins/http2_adapter/test/early_hints_test.dart
@@ -1,0 +1,142 @@
+import 'dart:async';
+import 'dart:io';
+
+import 'package:dio/dio.dart';
+import 'package:dio_http2_adapter/dio_http2_adapter.dart';
+import 'package:http2/transport.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('ignores 1xx interim responses and resolves with the final response',
+      () async {
+    final fixture = await _H2Fixture.serve((stream) {
+      stream.sendHeaders(
+        [
+          Header.ascii(':status', '103'),
+          Header.ascii('link', '</s.css>; rel=preload; as=style'),
+        ],
+        endStream: false,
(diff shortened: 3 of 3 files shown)


AlexV525:
Review notes — both non-blocking, verified locally (the regression test reproduces the exact `Bad state: Future already completed` crash against current `main` and passes with the fix; `dart test`, `dart analyze`, `dart format` all clean):

## 1. Trailers are now deterministically dropped — please disclose it and add a hardening test

A trailer section arrives as a HEADERS frame *without* `:status` and with END_STREAM (RFC 9113 §8.4, §8.8.5). The new code collects `frameHeaders` but discards them whenever `status == null`. Before this PR, trailers were merged into `responseHeaders` — but only racily (observable only when they landed before the response object was built, since `Headers.map` exposes the live internal map), so a deterministic drop is defensible. Still, it silently discards server data (e.g. gRPC trailing metadata — the only channel for `grpc-status`, required by the gRPC/HTTP2 spec). Please:

- add one sentence to the CHANGELOG entry / PR description noting that trailers are now discarded;
- add a test where the server sends a trailer HEADERS frame after DATA and asserts the request still completes normally (pins the behavior, guards regressions).

Proper trailer support is an API addition — opened as #2602.

## 2. Malformed `1xx` + END_STREAM should fail fast; and `badResponse` would be the wrong type

`HeadersStreamMessage` exposes `endStream` (package:http2, `transport.dart`). A 1xx HEADERS frame with END_STREAM terminates the stream without a final response — malformed per RFC 9110 §15.2 ("1xx … prior to a final response"; "terminated by the end of the header section") together with RFC 9113 §8.4 (interim responses carry no trailers and precede a final response). Go's `x/net/http2` errors out on this ("1xx informational response with END_STREAM flag"). With `receiveTimeout` unset (the default), the current code waits forever while the stream holds a pooled connection. Suggested shape:

```dart
if (code >= 100 && code < 200) {
  if (message.endStream && !responseCompleter.isCompleted) {
    responseCompleter.completeError(
      DioException.connectionError(
        requestOptions: options,
        reason: 'Received an interim 1xx response with END_STREAM; '
            'the stream ended without a final response.',
      ),
    );
  }
  return;
}
```

`connectionError` matches the existing precedent in this same method ("Received redirect without location header." already uses `DioException.connectionError`); `badResponse` would mean "a final response failed `validateStatus`" — which is exactly the pre-fix symptom this PR removes.

Both points are the same concern (HEADERS-frame handling robustness), so folding them into this PR seems reasonable if you agree.

*Comment drafted with AI assistance (GLM), reviewed and submitted by a human.*

GrishMahat:
@chiliec Thank you for fixing this issue! Are you planning to update this PR according to the review comments?

AlexV525:
Both points addressed in 51bc3a0 — verified locally:

- **1xx + END_STREAM fail-fast**: implemented as suggested (`connectionError`, guarded by `!responseCompleter.isCompleted`). Confirmed genuine RED→GREEN: with the previous adapter version the new test times out (`TimeoutException` — the exact hang being guarded); with the fix it passes and asserts the right `DioExceptionType` and message.
- **Trailer disclosure + hardening test**: CHANGELOG now documents the drop and links #2602; the new trailer-after-DATA test passes and pins the behavior (`grpc-status` stays absent).
- Full package suite: **108 passed / 4 skipped** (was 106), `dart analyze` and `dart format` clean on the branch.

One housekeeping item (§8.3): commit 51bc3a0 is missing the `Co-Authored-By: Claude <noreply@anthropic.com>` trailer that 50c0cb2 carries — please add it if this commit was likewise AI-produced. Note that amending/rebasing to add it is fine here since the PR has no attached review comments on those positions yet (or just include it in the next commit).

With that, this looks ready to me from the implementation side. Trailer support itself: discussion continues in #2602.

*Comment drafted with AI assistance (GLM), reviewed and submitted by a human.*

